PR #8, opened 2016-02-29
Make stickers list immutable

### LINKED ISSUE #5: Refactor constants/stickers.js to be immutable.

mxstbr:
Closes #5

### DIFF
--- a/src/components/Editor/index.js
+++ b/src/components/Editor/index.js
@@ -69,6 +69,16 @@ export default class UnicornEditor extends Component {
   };
 
   render() {
+    const stickerElements = stickers.map((sticker) => (
+      <button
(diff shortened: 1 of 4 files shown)


src/components/Editor/index.js
```diff
@@ -69,6 +69,16 @@ export default class UnicornEditor extends Component {
   };
 
   render() {
+    const stickerElements = stickers.map((sticker) => (
```
mxstbr: Will this ever rerender, should we move this and the `.toArray` in the `return` on line 94 to the constructor?

nikgraf: Yep, I was actually thinking to have the stickers not hardcoded, but as a prop that you can pass in.

PR #24, opened 2016-03-02
Fix link & improve popover behaviour

nikgraf:


src/linkPlugin/Link/index.js
```diff
@@ -7,8 +7,8 @@ export default class Link extends Component {
 
   render() {
     return (
-      <a {...this.props} href={this.props.children} style={styles.root}>
-        {this.props.children}
+      <a { ...this.props } href={ this.props.children[0].props.text } style={ styles.root }>
```
nikgraf: There might be a better/safer solution. any ideas?

PR #54, opened 2016-03-06
feat(Mentions): adding hover state and filtering for mentions

jpuri:
Hey @nikgraf,

PR has changes for:
1. Adding hovered state to dropdown options.
2. Adding filtering for mentions, the filtering logic is that text entered by the user can be anywhere in option and it returns first 5 options filtered.

I could not add keyboard handling as keystrokes are handled at level of editor and are currently not communicated over to plugins. Since dropdown is not focused it can not get keyDown event. We should create hook for these events in plugins which can be called by editor.

### DIFF
--- a/src/mentionPlugin/MentionSearch/index.js
+++ b/src/mentionPlugin/MentionSearch/index.js
@@ -14,12 +14,16 @@ export default (mentions) => {
     };
 
     // Get the first 5 mentions that match
-    getMentionsForFilter = () => (
-      mentions
+    getMentionsForFilter = () => {
 
-      // TODO better search algorithm than startsWith
-      // mentions.filter((mention) => mention.get('handle').startsWith(this.mentionSearch)).slice(0, 10)
-    );
+      const userEntered = this.props.editor.props.editorState.getCurrentContent().getLastBlock().getText();
+      const lastMentionStarts = userEntered.lastIndexOf("@");
+      const mentionValue = userEntered.substring(lastMentionStarts + 1, userEntered.length);
+      let filteredValues = mentions && mentions.filter((mention) =>
+        !mentionValue || mention.get('handle').toLowerCase().indexOf(mentionValue.toLowerCase()) > -1);
+      const size = filteredValues.size < 5 ? filteredValues.size : 5;
+      return filteredValues.setSize(size);
+    };
 
     renderItemForMention = (mention) => (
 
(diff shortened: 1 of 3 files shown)


mxstbr:
Awesome, looks good!

nikgraf:
sweet @jpuri - ok, I will look into how we can take over key events

src/mentionPlugin/MentionSearch/index.js
```diff
@@ -48,6 +54,7 @@ export default (mentions) => {
               ))
             }
           </Dropdown>
+          : void 0}
```
nikgraf: Is there are reason to use `void 0` instead of `undefined`? I think UglifyJs does the transform anyway and `void 0` is a bit confusing.

src/mentionPlugin/MentionSearch/index.js
```diff
@@ -14,12 +14,16 @@ export default (mentions) => {
...
 
     // Get the first 5 mentions that match
-    getMentionsForFilter = () => (
-      mentions
+    getMentionsForFilter = () => {
 
-      // TODO better search algorithm than startsWith
-      // mentions.filter((mention) => mention.get('handle').startsWith(this.mentionSearch)).slice(0, 10)
-    );
+      const userEntered = this.props.editor.props.editorState.getCurrentContent().getLastBlock().getText();
```
nikgraf: While the filtering works well it' pretty much broken since the mention doesn't need to be the last `@`.

nikgraf: Filtering works well!

It breaks when the `@` mention is not happening at the end. See: 
filter

jpuri:
Yeh, a block can have multiple mentions, I need to check how can I get the current word where cursor is from draftjs editor, also then we would need to close the mentions drop-down when user clicks somewhere else on the page right now you can change mouse position and still select some mention option.

src/mentionPlugin/MentionSearch/index.js
jpuri: That is actually more of habit, I will change that to undefined. Also we actually do not need a separate dropdown component, here is not doing anything more then applying styles to a single div.

jpuri:
May be a good way to propagate the events will be to create hooks in plugins for events like keydown, editorFocus, editorBlur - if plugins register their listeners we can call them as event occurs.

nikgraf:
working on it!

Issue #66, opened 2016-03-17
Plugin Configuration


juliankrispel:
Mention Plugin

```js
{
  mentions: List([Map({
    link: String,
    avatarUrl: String  
  })]),
  theme: {
    search: Map(),
    mention: Map(),
   },
}
```

Hashtag Plugin
```js
{
   theme: {
      hashtag: Map(),
   }
}
```

Undo Plugin
```js
{
  theme: {
    undoButton: Map(),
    redoButton: Map(),
  }
}
```

Link Plugin
```js
{
   theme: {
      link: Map(),
   }
}
```

stickerPlugin
```js
{
   theme: {
     sticker: Map(),
     stickerSelect: Map(),
   }
}
```

emojiPlugin

```js
{
   emojis: List([Map(
     url: String,
     name: String,
     key: String,
     unicodeCharacter: String, // (replaced with image, should only show image if it is in the list of emojis)
   )]),
   theme: {
     emoji: Map(),
     emojiSelect: Map(),
   }
}
```

nikgraf:
@juliankrispel I just checked out thoughts here again. How would the config for the mention plugin look like with a callback function that returns a promise to fetch data from a server?

Should `mentions` be either a List or function or rather have two props? Maybe something else?

juliankrispel:
@nikgraf I guess it could be optional: `list` or a `function` returning a list, or a `promise` which returns a list.

nikgraf:
sounds good to me :+1:

nikgraf:
done :smile:

PR #67, opened 2016-03-17
Themable Plugins

juliankrispel:
Hey there.

Right so I did the work to enable themable plugins. I also converted most of the plugins to be themable, although that's not really complete yet.

The gist of it is that you pass an immutable map that contains all the styles.

These styles can be either strings or objects. If they are strings they'll be attached as a `className prop`, if they are objects they'll be attached as a `styles prop`.

Let me know if you think that's sane.


juliankrispel:
looks like I have a bunch of eslint errors whoops

juliankrispel:
will fix this #66

PR #70, opened 2016-03-17
new strategy for theming hashtags

nikgraf:


src/hashtagPlugin/index.js
```diff
@@ -1,23 +1,33 @@
...
+  hashtag: styles.hashtag,
 });
 
-const hashtagPlugin = (config = {}) => ({
-  pluginProps: {
-    decorators: [
-      {
-        strategy: hashtagStrategy,
-        component: Hashtag,
-        theme: defaultHashTagTheme.merge(config.theme),
```
nikgraf: I reverted this back in order to keep the API as similar as the original draftjs API: https://facebook.github.io/draft-js/docs/advanced-topics-decorators.html#compositedecorator

In addition this makes it a bit more flexible, because then people who build a plugin just for themselves do not need to know anything about theme.

PR #96, opened 2016-03-23
Emoji plugin

nikgraf:


draft-js-emoji-plugin/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "draft-js-emoji-plugin",
+  "version": "0.0.0",
```
psbrandt: Did you want to bump the version?

psbrandt:
You need to add the plugin to the linter ignore list I think :).

nikgraf:
good point @ linting :+1:

draft-js-emoji-plugin/package.json
nikgraf: i noticed we actually need to compile it with webpack and export the css, otherwise the css import is still in there and it won't work without css modules

PR #97, opened 2016-03-23
Add missing ImmutableJS Map import

razh:
Otherwise, ES6!


nikgraf:
oh yeah, thx a lot! we need some tests here :smile:

PR #114, opened 2016-03-24
Added usage description and removed custom emoji editor

### LINKED ISSUE #113: 'Custom' Emoji editor on website broken
Basically the custom editor on the emoji page looks like this right now if you try autocomplete:

image

juliankrispel:
fixes #113 cc @nikgraf


juliankrispel:
Fyi the reason why the 'Custom' emoji editor was broken because we're not merging themes with the defaults and the usage on the emoji homepage implied merging, this is what it looked like:
```
const theme = Map({
  emoji: emojiStyles.emoji,
})
```

obviously that's going to break because the theme isn't merged and then you have a ton of undefined theme attributes.

I think merging has value, more about this in #116

nikgraf:
@juliankrispel logo is in there, but not used anywhere?

juliankrispel:
@nikgraf that's a mistake will remove!

juliankrispel:
@nikgraf removed

PR #117, opened 2016-03-24
added tab key completion 

### LINKED ISSUE #108: When you press the tab key on mention suggestions the current suggestion is not selected
At the moment you can only select a mention suggestion by pressing enter or using your mouse but it's also a very common convention to use the tab key to select the current suggestion.

juliankrispel:
to editor as well as emoji and mention plugin fixes #108

cc @nikgraf 

I guess this pr would be a ton more useful with specs, sry about that.


nikgraf:
looks good, thx :+1:

PR #131, opened 2016-03-26
Create detailed Webpack instruction to import CSS

rajaraodv:
Just adding more instructions to help people know exactly how to import CSS via webpack.


mxstbr:
This is much better than before, thanks so much for taking the time to add this! ✨

PR #141, opened 2016-03-28
Remove unnecessary parameter from addEmoji and addMention modifiers

jaaberg:
I suggest we remove the parameter selection from the addEmoji and the addMention modifiers as we can get it from the editorState. Whenever the modifiers are used, we retrieve the selection from the editorState just to pass it on to the function, and it seems better to me to just retrieve it inside the function. 

You could argue that it is possible to pass in another selection than what is in the current editorState (as is done in getSearchText), but for all use cases of the addEmoji and the addMention modifiers we just retrieve the selection from the current editorState. 

(I'm new to doing pull requests, so any feedback to improve is appreciated.)


nikgraf:
Looks good to me. Thanks for your contribution :smile:

PR #145, opened 2016-03-29
Add draft-js-dnd-plugin for block moving and file uploading + demo

bkniffler:
The example isn't yet documented, its basically only the demo. Also there is no navigation point, though you can find it under the /plugins/dnd route. I've removed everything server related, the example is github pages friendly, and no additional dependencies are necessary.

The changes to `draft-js-plugins-editor/src/Editor/index.js` aren't exactly DRY, but we will need to figure out a better solution to call plugin methods anyways.


nikgraf:
Todos we can do after merging this:

- [ ] Update Readme
- [ ] Fill the simple editor already with image content so people see something right away
- [ ] Improve the styling of buttons to move the image (Left/Center/Right)
- [ ] Release the plugin on NPM & make sure Ben & Nik can release it

draft-js-dnd-plugin/src/Image/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+    const className = `${theme.get('imageWrapper')} ${theme.get(alignment || 'center')}`;
+    return (
+        <figure className={ className }
+          contentEditable={false}
+          data-offset-key={ `${block.get('key')}-0-0` }
+          onDragStart={onDragStart} draggable={draggable}
+        >
+          <img src={blockProps.src || (`/images ${blockProps.url}`)} width="100%" height="auto" className={ theme.get('image') } />
+          {blockProps.progress >= 0 ? <div style={{ position: 'absolute', width: `${100 - blockProps.progress}%`, right: 0, top: 0, height: '100%', backgroundColor: 'white' }} /> : null}
```
nikgraf: While calculating the width with inline-styles the rest should probably go into a class.

draft-js-dnd-plugin/src/index.js
```diff
@@ -0,0 +1,55 @@
...
+
+      // This is for block-alignment-wrapper, only temporarily living here
+      injectBlockProps: (block, getEditorState, onChange) => {
+        const editorState = getEditorState();
+        return {
+          refreshEditorState: () => onChange(EditorState.forceSelection(editorState, editorState.getCurrentContent().getSelectionAfter())),
+        };
+      },
+    },
+    addListener: emitter.addListener,
```
nikgraf: How does addListener work? What's the use-case?

draft-js-dnd-plugin/src/index.js
```diff
@@ -0,0 +1,55 @@
...
+  // This is for block-alignment-wrapper, only temporarily living here
+  left: imageStyles.left,
+  center: imageStyles.center,
+  right: imageStyles.right,
+});
+
+const uploadPlugin = (config = {}) => {
+  const theme = config.theme ? config.theme : defaultTheme;
+
+  const attachButtons = config.attachButtons !== false;
```
nikgraf: nice!

draft-js-dnd-plugin/src/Image/index.js
```diff
@@ -0,0 +1,66 @@
...
+    this.props.align('right');
+  };
+
+  render() {
+    const { blockProps, block, attachButtons, theme, alignment, onDragStart, draggable } = this.props;
+
+    const buttons = [
+      <span className={ theme.get('imageButton') }
+        onClick={this.alignLeft}
+        style={{ marginLeft: '-2.4em' }}
```
nikgraf: What about making a theme property: leftImageButton, centerImageButton, rightImageButton. This case it's more flexible. The hard-coded marginLeft -2.4em might not work for other people.

draft-js-dnd-plugin/src/Image/index.js
```diff
@@ -0,0 +1,66 @@
...
+    ];
+
+    const className = `${theme.get('imageWrapper')} ${theme.get(alignment || 'center')}`;
+    return (
+        <figure className={ className }
+          contentEditable={false}
+          data-offset-key={ `${block.get('key')}-0-0` }
+          onDragStart={onDragStart} draggable={draggable}
+        >
+          <img src={blockProps.src || (`/images ${blockProps.url}`)} width="100%" height="auto" className={ theme.get('image') } />
```
nikgraf: The url here probably should be the full-url.

draft-js-dnd-plugin/src/components/block-alignment-wrapper.js
```diff
@@ -0,0 +1,36 @@
...
+        this.props.blockProps.refreshEditorState();
+      }
+    }
+
+    render() {
+      const { blockProps } = this.props;
+      return (
+          <C {...this.props}
+            alignment={blockProps.alignment}
+            align={::this.align}
```
nikgraf: Binding inside the render function would cause a re-binding on ever render. That's why we opted in for the syntax with arrow functions:

```
align = (alignment) => {
  …
}
```

draft-js-dnd-plugin/src/components/dnd-wrapper.js
```diff
@@ -0,0 +1,54 @@
...
+
+    componentWillUnmount() {
+      this.DOMNode.removeEventListener('dragover', this.onDragOver);
+      this.DOMNode.removeEventListener('dragleave', this.onDragLeave);
+      this.DOMNode.removeEventListener('drop', this.onDragDrop);
+
+      plugin.removeListener('progress', this.onProgress);
+    }
+
+    onDragOver = (e) => {
```
nikgraf: what about `event` instead of `e`?

draft-js-dnd-plugin/src/components/dnd-wrapper.js
```diff
@@ -0,0 +1,54 @@
...
+      this.setState({ isDragging: false });
+    };
+
+    onProgress = (e) => {
+      this.setState({ progress: e, isDragging: false });
+    };
+
+    render() {
+      const { isDragging, progress } = this.state;
+      return <C ref="com" isDragging={isDragging} progress={progress} />;
```
nikgraf: What about `WrappedComponent ` instead of `C`

nikgraf:
@bkniffler 

1. Nope, I don't see preview image.
2/3. Didn't notice you can actually drop images. Then the name is probably right. Really cool. We need to think about how to make it obvious  that drag&drop is possible via styling/hover. This also goes hand in hand with the styling of the Left/Center/Right button. I like your style on wysiwyg you wrote. Let's use that and we can think about a consistent default style later :smile:

draft-js-dnd-plugin/src/modifiers/onDropBlock.js
```diff
@@ -0,0 +1,40 @@
+import AddBlock from './addBlock';
+import RemoveBlock from './removeBlock';
+import { Entity } from 'draft-js';
+
+export default function onDropBlock() {
+  return function onDropBlockInner(e) {
```
nikgraf: event instead of e would be great

draft-js-dnd-plugin/src/modifiers/onDropFile.js
```diff
@@ -0,0 +1,115 @@
...
+        previews.forEach(x => {
+          state = AddBlock(state, selection, 'image', { ...x, progress: 1 });
+        });
+        onChange(state);
+
+        // Perform upload
+        upload(data, uploadedFiles => {
+          // Success, remove 'progress' and 'src'
+          let newEditorState = editorState();
+          uploadedFiles.forEach(x => {
```
nikgraf: can you replace `x` with something like `file`?

draft-js-dnd-plugin/src/modifiers/onDropBlock.js
```diff
@@ -0,0 +1,40 @@
...
+    const { selection, dataTransfer, editorState, onChange } = e;
+
+    const state = editorState();
```
nikgraf: since `editorState` is actually not the editorState, but rather getEditorState I recommend to rename it.

same for `onChange`, better to use `updateEditorState`

draft-js-dnd-plugin/src/modifiers/onDropBlock.js
```diff
@@ -0,0 +1,40 @@
...
+
+    // Get data 'text' (anything else won't move the cursor) and expecting kind of data (text/key)
+    const raw = dataTransfer.data.getData('text');
+    const data = raw ? raw.split(':') : [];
```
nikgraf: Why split with a colon?

draft-js-dnd-plugin/src/modifiers/onDropBlock.js
```diff
@@ -0,0 +1,40 @@
+import AddBlock from './addBlock';
+import RemoveBlock from './removeBlock';
```
nikgraf: RemoveBlock is a function and should be `removeBlock` to avoid confusion

draft-js-dnd-plugin/src/modifiers/onDropFile.js
```diff
@@ -0,0 +1,115 @@
...
+  });
+}
+
+function readFiles(files) {
+  return Promise.all(files.map(readFile));
+}
+
+function getBlockWhereEntityData(state, query) {
+  const map = state.getCurrentContent().get('blockMap').toArray();
+  for (let i = 0; i < map.length; i++) {
```
nikgraf: we probably can use `reduce` from immutable.js instead of converting it to an array & iterating over it

draft-js-dnd-plugin/src/modifiers/onDropFile.js
```diff
@@ -0,0 +1,115 @@
...
+
+  return null;
+}
+
+export default function onDropFile(config) {
+  return function onDropFileInner(e) {
+    const { props, selection, files, editorState, onChange } = e;
+
+    // Get upload function from config or editor props
+    const upload = config.upload || props.upload;
```
nikgraf: does this mean you could pass an upload function to the editor? Why not only have it inside config?

draft-js-dnd-plugin/src/modifiers/onDropFile.js
```diff
@@ -0,0 +1,115 @@
...
+    const progress = config.progress || props.progress || ((x) => config.emitter.emit('progress', x));
```
nikgraf: I would suggest to simply use callback functions instead of emitting events or is there any benefit using an event emitter?

draft-js-dnd-plugin/src/modifiers/onDropFile.js
```diff
@@ -0,0 +1,115 @@
...
+      }
+
+      data.formData = data;
+
+      // Read files on client side
+      readFiles(data.files).then(previews => {
+        // Add blocks for each image before uploading
+        let state = editorState();
+        previews.forEach(x => {
+          state = AddBlock(state, selection, 'image', { ...x, progress: 1 });
```
nikgraf: AddBlock should be addBlock

bkniffler:
Ah, shame, it should do that:
marz 31 2016 02 28

I'm going to give the styling a thought, its tricky to accomplish something unobstrusive and still obvious :)

draft-js-dnd-plugin/src/Image/index.js
bkniffler: You're right, but I've planned to get rid of these buttons anyways and replace them by a proper toolbar, so I didn't put a lot of effort in that. I'll just leave the styling until I'm done with the toolbar, which I will be able to work on and finish by the weekend.

draft-js-dnd-plugin/src/index.js
bkniffler: Thats part of the event emitter to propagate plugin events (like upload progress).

Checkout these parts:
https://github.com/draft-js-plugins/draft-js-plugins/blob/dnd/docs/client/components/pages/Dnd/SimpleDndEditor/index.js#L60
https://github.com/draft-js-plugins/draft-js-plugins/blob/dnd/draft-js-dnd-plugin/src/components/dnd-wrapper.js#L14

I'm creating the plugin, and hand it to a decorator function as an argument, so it can hook into the upload progress event.

Of course, we could think about other ways, but only an event emitter would allow multiple targets to easily bind to the same event. Also it will allow to hook/unhook into events at any time. It will be particularly useful as soon as plugins start to depend on each other.

I'm really interested to hear what you @nikgraf and @juliankrispel think about that.

draft-js-dnd-plugin/src/modifiers/onDropBlock.js
bkniffler: Yeah, this is something I wanted to give another thought. If you start dragging a draggable block, it will add a text key into dataTransfer that looks like this: `key:${block.key}`. The problem with dataTransfer is, you cannot add json (except if its stringified, which would be a pain to use, as you'd have to try parse the stringified json, it could still only be some text dropped on the editor).

Maybe I will change the `key:${block.key}` to something more unique like `DRAFTJS_DND_KEY:${block.key}` to really avoid conflicts. We could also add the blocks key to `window.CURRENT_DRAFTJS_DND_KEY` and read it from there, but I think everyone will agree this isn't a wise solution :)

The editor will also take care of `type:${some_block_type}` and create a block accordingly. This could be renamed to `DRAFTJS_BLOCK_TYPE:${some_block_type}`.

Maybe a good opportunity for string constants, gonna work on that.

draft-js-dnd-plugin/src/modifiers/onDropFile.js
bkniffler: See my comment on your `How does addListener work? What's the use-case?` question ;)

bkniffler:
Okay, everything that I gave a :+1: is fixed!

mxstbr:
Holy moly this is awesome, well done!

PR #151, opened 2016-03-29
Add editorState to handleKeyCommand callback

### LINKED ISSUE #126: getEditorState from Plugin
Hey @nikgraf 

First of all thanks for this cool project!
I was playing around with it and asking myself how to get the editor state from a plugin without passing it in via a component.

E.g: When I want to add a plugin, that handles keyCommands, I would use the `handleKeyCommand` callback, but I have no access to the editorState.

Of course I could add a `<KeyCommandComponent editorState={this.state.editorState}>` explicitly, but then I need to register the Plugin in two places.

What am I missing?

juliankrispel:
should fix #126

plz review @nikgraf @HaNdTriX


HaNdTriX:
@juliankrispel looks good to me. Thanks!
I had the same PR in the pipe. Sorry for the [duplicate](https://github.com/draft-js-plugins/draft-js-plugins/pull/152)!

I also added the callbacks to the `keyBindingFn` to be more consistent.
@nikgraf feel free to pick the one that makes more sense to you :+1:

PR #152, opened 2016-03-29
Pass editorState getter & setter as param to keyBinding func fix #126

### LINKED ISSUE #126: getEditorState from Plugin

HaNdTriX:
This PR allows to update the editor state from keyBinding functions such as  `handleKeyCommand` or `keyBindingFn`.

## Example

    import {RichUtils} from 'draft-js';

    const createDefaultKeyCommandPlugin = (config = {}) => {
      return {
        pluginProps: {
          handleKeyCommand: (command, getEditorState, updateEditorState) => {
            return false;
            const editorState = getEditorState();
            const newState = RichUtils.handleKeyCommand(editorState, command);
            if (newState) {
              updateEditorState(newState);
              return true;
            }
            return false;
          }
        },
      };
    };

    export default createDefaultKeyCommandPlugin;

### DIFF
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -123,7 +123,7 @@ export default class PluginEditor extends Component {
     preventDefaultBehaviour = this.plugins
       .map((plugin) => {
         if (plugin.handleKeyCommand) {
-          const handled = plugin.handleKeyCommand(command);
+          const handled = plugin.handleKeyCommand(command, this.getEditorState, this.onChange);
           if (handled === true) {
             return handled;
           }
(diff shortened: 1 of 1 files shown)


nikgraf:
@HaNdTriX in the end I picked the version from @juliankrispel 

nevertheless I merged yours before so you are listed in the contributors. thx for your contribution

Issue #156, opened 2016-03-31
Image/Drag&Drop Plugin Improvements


nikgraf:
- [ ] Update Readme for dnd
- [ ] Fill the simple editor already with image content so people see something right away
- [ ] Improve the styling of buttons to move the image (Left/Center/Right)
- [ ] Release the plugin on NPM & make sure Ben & Nik can release it
- [ ] Is dnd the best name? I expected to drag&drop images.
- [ ] see if we need the event emitter and if callbacks might be enough …
- [x] change (e) to (event)
- [ ] make it optional to have image file upload via dropping, make image drag&drop optional
- [x] replace (C) with (WrapperComponent)
- [ ] deprecate injectBlockProps 
- [ ] investigate if handleDroppedFiles & handleDrop can be simplified together with the listeners

nikgraf:
@bkniffler do you see a way to avoid using injectBlockProps? I would like to keep it simple and this concept is a bit harder to understand …

bkniffler:
These are the todos I currently see need to be done, based on your list

- [ ] Check eventEmitter, maybe replace by simple callbacks
- [ ] Update Readme for dnd
- [ ] Update example with API documentation
- [ ] Make example image component more simple, remove alignment feature
- [ ] Create a more complex example with a custom image that uses the toolbar-plugin and alignment
- [ ] Allow for custom upload placeholder (no placeholder, gray box with spinning loader, or a preview of the file)
- [ ] Release the plugin on NPM & make sure Ben & Nik can release it
- [ ] Enable/disable drag&drop of images/blocks (manually and on readOnly=true)

Is the name still an issue for you?

nikgraf:
@bkniffler my thougths on name: dnd doesn't necessarily have to mean image drag & drop. I could video or audio or even text that is dragged around. To me it seems like this almost is a image plugin with 3 parts:

- upload
- image management
- image drag & drop support

What do you think?

bkniffler:
I agree, DnD doesn't mean image drag&drop. The plugin should provide all the handlers to make drag&drop possible in general, wether it is video, audio or images, wether they are dropped from the desktop, they are blocks within the editor itself or blocks in a toolbar dropped on the editor. The only thing that really is image specific is the image component, which is rather simple (if you strip out the alignment stuff that should not really be part of this plugin anyways).

[...]

PR #159, opened 2016-03-31
Remove inject block props

bkniffler:
As requested in https://github.com/draft-js-plugins/draft-js-plugins/issues/156

For now, it might not be necessary. We can reevaluate something like that as soon as I see no other reasonable way around it.


nikgraf:
yay! thx - that makes it easier to merge all the other editorImprovements

PR #160, opened 2016-03-31
refactor addEmoji modifier

jaaberg:
Refactored the variable names and comments as it probably was forgotten when copying code from addMention.

(I'm new to making PR's, so any feedback to improve is really appreciated.)


nikgraf:
@jaaberg great! thx for your contribution

P.S. ping In case you want to do more, but don't where to start. There is a ton of work to be done :smile:

jaaberg:
@nikgraf Since I'm kind of deepdiving into draft now anyways, I could probably use some time contributing. If you have some simpler tasks I'll be happy to help out, but I can't guarantee how much time I have due to fulltime job and currently moving as well. 

But please point me to where I can start and I will give it a try.

PR #165, opened 2016-04-01
Editor refactoring

nikgraf:
I added a couple tests and refactored the editor based on a mixture what we had + Ian's suggestions. @ianstormtaylor would you mind reviewing these changes?
@juliankrispel can you also have a look? :smile: 

@bkniffler the `handleDrop` & `handleDroppedFiles` had as last argument `this.props`. I didn't find any point in the code where it was used nor could I think of why it would be useful and therefor removed it. Can you confirm that we don't need it?


juliankrispel:
@nikgraf looking <3

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -24,14 +20,23 @@ export default class PluginEditor extends Component {
...
 
+  static defaultProps = { plugins: [] };
+
   constructor(props) {
     super(props);
-    this.plugins = List(props.plugins).toArray();
-    const compositeDecorator = createCompositeDecorator(this.plugins, this.getEditorState, this.onChange);
+
+    // attach proxy methods like `focus` or `blur`
+    for (const method of proxies) {
```
juliankrispel: nice, I wanted to add this too

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -24,14 +20,23 @@ export default class PluginEditor extends Component {
...
+      this[method] = (...args) => (
+        this.refs.editor[method].apply(this.refs.editor, args)
```
juliankrispel: Minor nitpick, write `this.refs.editor[method](...args)` instead

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -64,195 +69,61 @@ export default class PluginEditor extends Component {
...
-    };
-
-    const keepHandlers = ['onChange', 'handleKeyCommand'];
-
-    // bind random onListeners and handleListeners
-    this.plugins.forEach((plug) => {
-      Object.keys(plug).forEach((attrName) => {
-        if (attrName.indexOf('on') === 0 && keepHandlers.indexOf(attrName !== -1)) {
-          listeners[attrName] = this.createOnListener(attrName);
+        if (attrName.indexOf('handle') === 0) {
```
juliankrispel: would be nice if draft-js would let us import the hook names rather than having to do stuff like this

juliankrispel:
@nikgraf lgtm!

juliankrispel:
:+1:

draft-js-plugins-editor/src/Editor/index.js
nikgraf: yeah, that's better

bkniffler:
Nice! 
The props indeed weren't necessary anymore, so its cool.

ianstormtaylor:
From the slack too, here's my current version of the editor, in case it's helpful: https://gist.github.com/ianstormtaylor/f28bdb94b0bf6593e45c

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -24,14 +20,23 @@ export default class PluginEditor extends Component {
...
+
   constructor(props) {
     super(props);
-    this.plugins = List(props.plugins).toArray();
-    const compositeDecorator = createCompositeDecorator(this.plugins, this.getEditorState, this.onChange);
+
+    // attach proxy methods like `focus` or `blur`
+    for (const method of proxies) {
+      this[method] = (...args) => (
+        this.refs.editor[method](...args)
```
ianstormtaylor: I think this should have a `return` so that the proxied return values are preserved.

nikgraf: good point, thx

nikgraf: hehe, just added a return and figured: we are already returning the proxied return value because of the arrow function without curly braces

ianstormtaylor: ahhh haha good call, i assumed they were curly without looking, my bad

nikgraf: yeah, to be honest I find this rather confusing. I'm more in favour of an explicit return, but this is a default airbnb rule now and I rather stick to their rules. By far the best well maintained eslint & jscs preset

PR #172, opened 2016-04-02
Hook arguments

nikgraf:
This will make plugin system more flexible for future changes.

In the future we can easily extend out the arguments depending on new features that will be available in DraftJS. We even could pass in the whole pluginEditor context (as we had it in the first implementation & @ianstormtaylor suggested) and it would not result in a braking the API.

Yet I don't feel comfortable to expand the API and pass down the whole pluginEditor context until there is a valid use-case.`focus()` came up as a suggestion, but this can easily be done with `forceSelection` on the editorState as well.

I'm curious how you guys feel about it? @juliankrispel @ianstormtaylor @bkniffler 

P.S. I merged it already so I can move on with some refactoring. If there are good arguments against this change I'm happy to revert back.

### DIFF
--- a/draft-js-dnd-plugin/src/blockRendererFn.js
+++ b/draft-js-dnd-plugin/src/blockRendererFn.js
@@ -2,7 +2,7 @@ import { Entity } from 'draft-js';
 import removeBlock from './modifiers/removeBlock';
 import refreshEditorState from './modifiers/refreshEditorState';
 
-export default (config) => (contentBlock, getEditorState, setEditorState) => {
+export default (config) => (contentBlock, { getEditorState, setEditorState }) => {
   const type = contentBlock.getType();
   if (type === 'image') {
     const entityKey = contentBlock.getEntityAt(0);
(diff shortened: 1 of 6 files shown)


bkniffler:
Seems like a good move and very much inline with what I'd expect from a hook; we won't always need every argument and we can always decide to add more arguments without breaking anything. No objection :+1:

ianstormtaylor:
+1 for this change, but I still don't see how this is better than passing the `editor` object itself.

nikgraf:
@ianstormtaylor I talked to Isaac and none of the public methods are supposed to be used by people except `focus`. Even focus you don't need at you can force the focus by manipulating the Selection.

I wonder if there is a use-case that would require to expose the `editorContext` to the plugins. I worried that people start to do hacky things like overwriting the behaviour of `blockRendererFn` because they couldn't figure out how to write a proper `blockRendererFn` for their plugin.

juliankrispel:
I'm all for restriction by default. All good things are better with a smaller API

Julian Krispel-Samsel
rainforestqa.com
goodafternoon.co

PR #173, opened 2016-04-03
createPluginHooks improvements

jl-:
only needs to call `createEventHooks` or `createFnHooks` once for the given `attrName` ?


nikgraf:
Totally see the value here, although it took me a while to understand why this is useful. May I ask you to refactor it so it generates an array containing all the hook keys (attrName) & then iterate over this list?

I believe this would make the code way more understandable and prevents unnecessary calls to `createEventHooks `

nikgraf:
nice one! thx

Issue #174, opened 2016-04-03
Extending Linkify with prefix and target config settings


adrianmc:
I wanted to get a better idea of how to contribute to a plugin so I decided to try to implement prefixes and target attributes to the Linkify plugin. Users will call the plugin like this:

```js
const linkifyPlugin = createLinkifyPlugin({
  prefix: 'http://',
  target: '_blank',
});
```

And the plugin will check to see if the href already begins with the prefix. If it does not, it will be added onto the front. Also, the target is set to whatever the user has chosen, otherwise it is null.

This is my rough first attempt. I know I'll need to clean it all up before submitting it for a pull request. I guess I'm just posting here to get some feedback on how I should refactor this since there doesn't yet seem to be a "dev conventions" guide written out.

Have a look at my commit here: https://github.com/adrianmc/draft-js-plugins/commit/2ea2413958c0d414ef34059b3c74d9173b145c54

Thanks!

nikgraf:
You changes look really good! I didn't even notice that http was missing before …

Two suggestions:

- by default the config for target should be '_self' (I thought about _blank, but matching the default behaviour of links is probably better to avoid confusion!
- by default the config for prefix should be 'http://'

both defaults can be set here: https://github.com/adrianmc/draft-js-plugins/commit/2ea2413958c0d414ef34059b3c74d9173b145c54#diff-2cd4a092a2a5a7cc5222f02ac1d3612fR19

This will get a good version out of the box, but will be able to configure it by themselves.

Definitely open a PR …

P.S: How should a dev conventions guide look like? What would you expect in there?

adrianmc:
@nikgraf

Great!

Regarding dev conventions, I just wasn't sure if some of the stuff I was doing was "kosher". For example, I wasn't sure that I could piggyback onto the config object since it originally only carried the theme information. I also re-wrote the eslint disabling comments as well as reformatted the props destructuring vertically. All of this stuff, I wasn't sure if I'd upset the original intent of the author.

I guess I'm pretty new to making major contributions (i.e. not simple bug fixes/typos/version bumps) to open sourced projects.

adrianmc:
Uh oh, it seems that the Travis build has failed in my PR #175 because I used the ES2015 `startsWith()` function.

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

I thought that Babel will compile down automatically?

nikgraf:
[...]

PR #175, opened 2016-04-03
Adding ability to set prefix and target for Linkify plugin

adrianmc:
Closes this issue: #174

### DIFF
--- a/docs/client/components/pages/Linkify/CustomLinkifyEditor/index.js
+++ b/docs/client/components/pages/Linkify/CustomLinkifyEditor/index.js
@@ -4,7 +4,7 @@ import Editor from 'draft-js-plugins-editor';
 import createLinkifyPlugin from 'draft-js-linkify-plugin';
 import editorStyles from './editorStyles.css';
 
-const linkifyPlugin = createLinkifyPlugin();
+const linkifyPlugin = createLinkifyPlugin({ target: '_blank' });
 const plugins = [linkifyPlugin];
 
 export default class CustomMentionEditor extends Component {
--- a/docs/client/components/pages/Linkify/SimpleLinkifyEditor/index.js
+++ b/docs/client/components/pages/Linkify/SimpleLinkifyEditor/index.js
@@ -4,7 +4,10 @@ import Editor from 'draft-js-plugins-editor';
 import createLinkifyPlugin from 'draft-js-linkify-plugin';
 import editorStyles from './editorStyles.css';
 
-const linkifyPlugin = createLinkifyPlugin();
+const linkifyPlugin = createLinkifyPlugin({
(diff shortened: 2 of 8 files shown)


adrianmc:
Uh oh, it seems that the Travis build has failed because I used the ES2015 startsWith() function.

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

I thought that Babel will compile down automatically?

nikgraf:
While we use babel to compile the code it is shipped to npm as ES5 code. People might not use babel-polyfills and as a library creator you don't want to prescribe them to it. Can you change it to use `indexOf`? Sorry, didn't see it before.

One more thing that is not necessary, but would make my life a lot easier is to add the new params to the docs here and even extend the CustomLinkifyPlugin to use target blank (it's a great example): https://github.com/draft-js-plugins/draft-js-plugins/tree/master/docs/client/components/pages/Linkify

Updating the Changelog in the plugin would be nice as well :smile: 

Also not necessary, but would be really cool is adding tests for it. That said the test coverage is pretty low except for the plugin editor. I'm aiming to improve it. In any way, I will merge it once the indexOf is change. Let me know if you want to do more.

adrianmc:
No worries, great suggestions. I'll get on those changes. I'll probably have it done in a day or so.

As for tests, I'm still pretty green to testing. Do you have any recommendations for how I should approach this? Should I just follow the same pattern in the existing tests file for the Link component?

adrianmc:
@nikgraf Several things:

1. I've replaced the prefix functionality with Linkify's own intelligent url mode. Previously, if we had `google.com` and `ftp://google.com` we would get hrefs like `http://google.com` and `http://ftp://google.com` respectively.

  Use of `IndexOf()` or `startsWith()` is no longer necessary.

  That's obviously bad, but by using Linkify's own methods, the result is now what the user expects (`http://google.com` and `ftp://google.com`).

  I've written tests for this, you can take a look. I also don't believe this requires specific documentation since it is the natural behaviour of the Linkify It library (https://github.com/markdown-it/linkify-it).

2. I have added the `_blank` example to CustomLinkifyEditor as you have requested. Please take a look.

3. I've updated the CHANGELOG as requested.

4. I added a bunch of tests that I think show the benefits of using the Linkify module's prefix functionality. Take a look to see if you think they are sufficient.

nikgraf:
This is soooo good! Thanks @adrianmc

nikgraf:
@adrianmc you changes are live on https://www.draft-js-plugins.com

I'm planning on cutting 1.0.0 of the linkify plugin mid/end this week.

adrianmc:
Awesome! Let me see what else I can help out with.

PR #182, opened 2016-04-05
fix mention regex if content block starts with @

svrcekmichal:
**Bug:**
if you type `@` at the start of content block it will not show mention search. 
**Fix:**
regex now checks if user type `@` as first content block char or after whitespace char


nikgraf:
that's better! thx :+1:

Issue #48, opened 2016-03-05
History Plugin


nikgraf:
The editor state contains a list of the operations. Would be nice to render this a history. Probably something similar to what Google Docs History, Photoshop History or Redux Devtools.

nikgraf:
I don't have clear picture what it should look like as I don't know if we even can get the description of the change like this one: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-sticker-plugin/src/modifiers/addSticker.js#L91, but something like this or an alternative like the Google Docs History might be interesting.

The first one here is photoshop with command information. The second one is from Google docs. By clicking on the stop you can directly go back to that state without even loosing the future history. Would you be interested to take up on this and become the main creator/maintainer of a history plugin? (could all be done in the same PR and you let me know when it's done)

<img width="208" alt="screen shot 2016-04-05 at 07 41 10" src="https://cloud.githubusercontent.com/assets/223045/14286012/73d783b0-fb02-11e5-81ac-7f7d3fbdba4a.png">
<img width="246" alt="screen shot 2016-04-05 at 07 43 00" src="https://cloud.githubusercontent.com/assets/223045/14286011/73d67132-fb02-11e5-9fb6-959f109844d9.png">

PR #183, opened 2016-04-05
Exposed History component for the Undo/Redo plugin

adrianmc:
I saw that the History component was not being used in any of the examples. Not only that, it wasn't being properly exposed so I couldn't even try it. Not sure how much value is in this, but at least it's no longer dead code with this PR.

Reference: #48 and #53 

Here's what the new example looks like:

History example

### DIFF
--- a/docs/client/components/pages/Undo/CustomUndoEditor/historyStyles.css
+++ b/docs/client/components/pages/Undo/CustomUndoEditor/historyStyles.css
@@ -1,0 +1,7 @@
+.historyItem {
+  color: #aaa;
+}
+
+.historyItemActive {
+  color: blue;
+}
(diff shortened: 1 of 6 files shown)


nikgraf:
I didn't want to expose it as I believe the current version has very little value. In addition the history should probably be a separate plugin. I don't have clear picture what it should look like as I don't know if we even can get the description of the change like this one: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-sticker-plugin/src/modifiers/addSticker.js#L91, but something like this or an alternative like the Google Docs History might be interesting.

The first one here is photoshop with command information. The second one is from Google docs. By clicking on the stop you can directly go back to that state without even loosing the future history.

screen shot 2016-04-05 at 07 41 10
screen shot 2016-04-05 at 07 43 00

adrianmc:
Hmmm, that's interesting. Thanks for putting up the pictures, it gives me a better idea of what would be of value. I also agree that it should probably be part of a separate plugin instead.

This is something that I do want to work on in the future, but perhaps I'll hold off on it. I'd like to spend my time contributing to parts of the plugin ecosystem that can further other parts of my understanding of Draft.js.

I think I'll leave this for now.

nikgraf:
sure, what about if we do a skype/hangout call & I explain what's on my mind?

adrianmc:
@nikgraf sure, I'll reach out to you on Twitter DM.

PR #187, opened 2016-04-08
Editor patch

bkniffler:
- Allow https://github.com/draft-js-plugins/draft-js-plugins/issues/156#issuecomment-207060175
- Allow for props to behave like the plugins instead of them overwriting plugin behaviour

I know `createFnHooks` isn't as beautiful as it used to be, but this is an important step. If you come up with a more clever way, please help refactoring it!

The changes will allow something like this:
```
export default function DndPlugin = (options) => {
    return {
        blockRendererFn: (contentBlock) => {
            return {
                decorators: [Draggable],
                props: {
                    allowDrag: options.allowDag !== false
                }
            }
        }
    }
}
```

The plugin will apply a decorator to all blocks, and also append the allowDrag prop.

### DIFF
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -88,14 +88,18 @@ class PluginEditor extends Component {
         if (typeof plugin[methodName] !== 'function') continue;
         const result = plugin[methodName](...newArgs);
         if (result !== undefined) {
-          const { decorators: pluginDecorators, props, ...rest } = result;
+          const { decorators: pluginDecorators, props, ...rest } = result; // eslint-disable-line no-use-before-define
           if (pluginDecorators) decorators = [...decorators, ...pluginDecorators];
           block = { ...block, ...rest, props: { ...block.props, ...props } };
         }
-      } if (block.component) {
+      }
+
+      if (block.component) {
         decorators.forEach(decorator => { block.component = decorator(block.component); });
         return block;
-      } return false;
+      }
+
+      return false;
     } else if (methodName === 'blockStyleFn') {
       let styles;
       for (const plugin of plugins) {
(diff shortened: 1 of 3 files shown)


draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -94,7 +126,7 @@ class PluginEditor extends Component {
     const pluginHooks = {};
     const eventHookKeys = [];
     const fnHookKeys = [];
-    const plugins = this.resolvePlugins();
+    const plugins = [this.props, ...this.resolvePlugins()];
```
nikgraf: Nice one!

nikgraf:
Super cool change! I had some of these changes on my mind for days and you did it even better than thought of.

In the last week I added a ton of test for the editor itself. Luckily the change in props triggered a test to fail :smile: Do you have time to fix this one test & write others for the new features? If now I'm happy to do so before merging.

bkniffler:
Okay, I fixed the test, but I'm currently finishing other work as I'm gonna go on vacation tomorrow (florida, yay!). Got my laptop and internet though, I will continue to work on the plugins. It would help if you wrote the test.

I've added a rough placeholder of how I imagine the test could work to the test file

```jsx
    it('blockRendererFnWithDecorator', () => {
      const decorator = (Component) => (props) => <Component {...props} />;
      const component = () => null;
      const plugin = {
        blockRendererFn: () => ({
          decorators: [decorator],
          props: { pluginProp: true },
        }),
      };
      customHook = () => ({
        component,
        props: { editorProp: true },
      });
      const result = mount(
        <PluginEditor
          editorState={ editorState }
          onChange={ onChange }
          plugins={ [plugin] }
          blockRendererFn={ customHook }
        />
      );
      /*
      should result in blockRendererFn returning {
        component: <decorator><component/></decorator>,
        props: { pluginProp: true, editorProps: true }
      }
       */
      expect(!!result);
    });
```

juliankrispel:
merged in master and added spec as suggested cc @nikgraf @bkniffler.

PR #188, opened 2016-04-08
fixed block src issue if upload the same picture by dnd.

madeinfree:
Hi, Thanks for awesome plugins first, I'm very interesting for this dnd plugin.

I found the little issue when i use the Drag&Drop, if I drop the same picture on Editor, it would be broken and always reload the first picture which are the same base64 pictures.

I used the `time` to make the different key for temporary.

if it has the best solution to solve it, feel free to discussion with me :D

### DIFF
--- a/draft-js-dnd-plugin/src/Image/index.js
+++ b/draft-js-dnd-plugin/src/Image/index.js
@@ -49,13 +49,14 @@ class Image extends Component {
     ];
 
     const className = `${theme.get('imageWrapper')} ${theme.get(alignment || 'center')}`;
+    const fixedBlockUrl = blockProps.src.split('<blockFixed>')[0];
     return (
         <figure className={ className }
           contentEditable={false}
           data-offset-key={ `${block.get('key')}-0-0` }
           onDragStart={onDragStart} draggable={draggable}
         >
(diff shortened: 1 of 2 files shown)


nikgraf:
Hi @madeinfree, sorry for the late response. I was busy the last couple days. This could be an interesting solution. Any specific reason why you closed the PR? @bkniffler is the guy who knows most about the dnd plugin.

@bkniffler what do you think about this PR?

madeinfree:
Thanks @nikgraf, I would reopen the PR.
Because the dnd demo example look like not stable from #156, I confusion my PR is appropriate or not now is my reason why I close my PR..

I have some question about data save to storage from convertToRaw, I don't know how can I do.

nikgraf:
@madeinfree Since @bkniffler is currently on vacation we need to wait a bit until he can review this change. Sorry for the inconvenience …

bkniffler:
Hey, sorry for the long absence @madeinfree. I'm working on a bigger update where this won't be an issue any more, thank you for the PR and for making us aware of this!

TheoRyley:
This issue still exists.

bkniffler:
Hey, thanks for reporting this @TheoRyley. It should be fixed now by https://github.com/draft-js-plugins/draft-js-plugins/commit/01eb28d2466a797e4153a794065bc1e08ba6e5fc

Issue #102, opened 2016-03-23
Character Count Plugin


neuling:
Shows a simple character counter (optional: conditional styling if counter exceeds configurable char limits?)

nikgraf:
Love it!

adrianmc:
I'll take this up just to get familiar with creating plugins. Can I get assigned?

mxstbr:
Can't assign people that aren't collaborators, so your comment will have to suffice :+1: Looking forward to it!

nikgraf:
@adrianmc I added you as a collaborator. you now can work from a branch instead of a fork :)

adrianmc:
@nikgraf Awesome! thanks a lot :+1:

adrianmc:
As promised, here is my first attempt in making something useful. Check out the `counter-plugin` branch here: https://github.com/draft-js-plugins/draft-js-plugins/tree/counter-plugin

You can see it in action by:

1. `git clone https://github.com/draft-js-plugins/draft-js-plugins.git`
2. `cd draft-js-plugins/`
3. `git pull origin counter-plugin`
4. `npm install`
5. `cd docs`
6. `npm install`
7. `npm run start`

Go to [`http://localhost:3000/plugin/counter`](http://localhost:3000/plugin/counter) in your browser and check it out!

😄 

Or... you can watch this gif I made:

![counter plugin being shown off](http://i.imgur.com/txYAnHA.gif)

nikgraf:
@adrianmc really appreciate your effort & the good execution here!

PR #190, opened 2016-04-10
Counter plugin

adrianmc:
#102 

This is my first attempt at the above issue.


draft-js-counter-plugin/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "draft-js-counter-plugin",
+  "version": "0.0.5",
```
nikgraf: Can you start from version 0.0.1? I'm not entirely sure, but I think keeping all the components & versions aligned will

draft-js-counter-plugin/src/CharCounter/index.js
```diff
@@ -0,0 +1,34 @@
...
+class CharCounter extends Component {
+
+  static propTypes = {
+    editorState: PropTypes.any.isRequired,
+    theme: PropTypes.any,
+  };
+
+  getCharCount(editorState) {
+    const plainText = editorState.getCurrentContent().getPlainText('');
+    const cleanString = plainText.replace(/(?:\r\n|\r|\n)/g, '').trim();
```
nikgraf: Can you comment on every part of the regex to make it clear which parts are there & why?

draft-js-counter-plugin/src/__test__/index.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
```
nikgraf: I would recommend to split up this test file & have a separate test in every component folder. The idea is to find the related test in the folder where the actual code is.

nikgraf:
@adrianmc I'm going to merge this! Really good work. Can you improve the things I mentioned and submit a separate PR?

In addition it would be great to get to at least the same quality level of the Emoji documentation or better. (Emoji is currently the best, but I hope to get all up to this level soon)

1. Description
2. Getting Started + Hint about CSS
3. Configuration Parameters
4. Simple Example
5. Custom Example (with a custom theme)

PR #205, opened 2016-04-17
[Counter Plugin] Added ability to incorporate custom counting functions

adrianmc:
I was looking at the counter plugin and thinking about how rigid it is, so I decided to try to implement a BYOF (Bring Your Own Function) feature so that the user can have more flexibility.

With this PR, the user can simply drop in their own function that takes a plaintext string as input and outputs a numerical count.

I've added tests to illustrate this.

What do you guys think? Yay or nay?


nikgraf:
Good one!

PR #213, opened 2016-04-19
[Counter Plugin] Complete Docs

adrianmc:
I've typed up some docs for the counter plugin. In this PR, I also did the following specific things:

1. Used flex box to better layout the configuration parameters section
2. Used normal JS Object instead of Map for the themed editor example

I do want to say that the whole process was kind of tedious and confusing. There are so many import statements required, and I had to do a lot of copy and paste from the other plugins. I think we should improve this in the future. I assume we'll overhaul the entire process after we cut the next release, so I'm excited for that 😄


nikgraf:
Well done!

Making the docs easier to edit sound good. What do you have in mind to simplify the process?

PR #245, opened 2016-04-29
Fix hashtag references

djkirby:
There are a bunch of references to the hashtag plugin from non-hashtag plugins.


nikgraf:
Thx for all the fixes 👍

Issue #233, opened 2016-04-24
TypeError: Cannot read property 'decorators' of null


sugarshin:
In [this line](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-plugins-editor/src/Editor/index.js#L109) you will receive an error such as the following.

```
TypeError: Cannot read property 'decorators' of null
```

What about a null equivalent judgment?

```js
if (result) {}
```

nikgraf:
@sugarshin sounds good. Wanna do a PR?

What about being more explicit & check for not undefined & not null?

nikgraf:
Just publish 1.0.1 including this fix.

PR #250, opened 2016-05-02
patch https://github.com/draft-js-plugins/draft-js-plugins/issues/233

sugarshin:
plz review.

ref: https://github.com/draft-js-plugins/draft-js-plugins/issues/233


nikgraf:
thx @sugarshin 👍

Issue #78, opened 2016-03-20
Autocomplete Component (to be used by other plugins)


juliankrispel:
The mention plugin has fairly fleshed out predictive text functionality, instead of redoing this for every plugin with functionality like that it'd be great to keep improving/working on one and making it available for building other plugins.

nikgraf:
Agreed, what I usually do in auch a case is copy the whole thing and build it out. Once it's fairly fleshed out I check where it makes sense to refactor code out into a common module. I would say: whatever works best for you is fine.

There is definitely value in such tooling

mjrussell:
Im actually pretty interested in creating this capability. I'll try to take this on. 

My first goal is to split the Suggestions (autocomplete) from the decorator which displays the Mention/Emoji/etc and then make this part a "plugin creator" i.e. taking parameters to render the suggestion entries.

I think the one downside of this approach I could see is that the mention plugin would need to be 2 separate plugins, i.e. mention suggestions, and mentions and I don't see an easy way to "merge" the result of the plugin-creator and the regular decorator.

danielberndt:
I really would like to see this feature as well. Here's one requirement for my current use case that might be interesting to consider: I would like to be able to offer an autocomplete for hashtags that also allows me to create Tag-Entities if I type a new tag not offered by the autocomplete.

nikgraf:
@danielberndt this actually is done by @mjrussell https://github.com/mjrussell/draft-js-autocomplete-plugin-creator, but could use some documentation … interested to helm him out?

PR #260, opened 2016-05-12
AutoComplete Plugin Creator (Feedback welcome)

mjrussell:
**Not ready to merge, looking for some feedback**

Took a stab at #78.

* Created an abstraction for autocomplete suggestions while typing in a draft-js-plugin (aka a plugin-creator)
* Refactored the mention plugin to use the plugin-creator

The plugin-creator API:
```js
const createCompletionPlugin = (
  completionSuggestionsStrategy,
  addModifier,
  SuggestionEntry,
  suggestionsThemeKey = 'completionSuggestions',
  additionalDecorators = [],
) => (config = {}) => {
```

The basic idea is to pull out the dropdown and suggestion scaffold from the rendering of Suggestion Entry as well as the modifier used in the onClick of an Entry. Another requirement I had is that the decorator for displaying the entity creator by the `addModifier` should be external to remain as flexible as possible.

I refactored the mention plugin as a proof that this would be usable (with only 1 external breaking change that could be fixed), but I haven't attempted the emoji plugin yet. I'd like to make sure this is on the right track before tackling emoji since I see some pain points (mainly with the custom onChange for the emoji plugin)

I think the current questions I have are:
* Is this a good place to split the abstraction/do we like the new API for the complete-plugin-creator?
* Should the mention & emoji plugins use this internally or should they remain independent?


draft-js-mention-plugin/src/index.js
```diff
@@ -82,54 +31,22 @@ const createMentionPlugin = (config = {}) => {
...
-    onDownArrow: (keyboardEvent) => callbacks.onDownArrow && callbacks.onDownArrow(keyboardEvent),
-    onTab: (keyboardEvent) => callbacks.onTab && callbacks.onTab(keyboardEvent),
-    onUpArrow: (keyboardEvent) => callbacks.onUpArrow && callbacks.onUpArrow(keyboardEvent),
-    onEscape: (keyboardEvent) => callbacks.onEscape && callbacks.onEscape(keyboardEvent),
-    handleReturn: (keyboardEvent) => callbacks.handleReturn && callbacks.handleReturn(keyboardEvent),
-    onChange: (editorState) => {
-      if (callbacks.onChange) return callbacks.onChange(editorState);
-      return editorState;
-    },
+  const completionPlugin = createCompletionPlugin(
```
nikgraf: The mentions plugin as well as Emoji currently take some parameters. Maybe I missed it, but it seems they are not passed down.

Instead of passing in multiple parameters I recommend to design the API in a way to pass in one object containing all sorts of configurations. This makes it easier to extend the config in the future and especially to eventually remove things.

nikgraf:
What about `draft-js-autocomplete-plugin-creator` as a name? Complete can mean a lot of things 😃

nikgraf:
This will probably need a guide/tutorial otherwise it's pretty tough to use. The autocomplete has quite some complexity.

I'm bit cautious to use it for core Emoji & Mention plugin. It creates another level of complexity and makes especially npm releases hard. Nevertheless I would love to see it come to happen as a tool, because then people could do a lot more cool stuff 🎉

nikgraf:
@mjrussell do you have some time to create some docs and a PR to add it to the Readme? I'm now a bit more certain that this is a fine tool, but best to have in a separate repo.

mjrussell:
@nikgraf yeah I think Im going to have some free time the next few weeks so Im going to catch up on all the draftjs-plugins notifications and updates I've been missing and then update the docs since the plugin-creator has been working really well in production for us for a little while now

nikgraf:
@mjrussell nice, looking forward to that 👍

PR #267, opened 2016-05-18
Fix rendering empty avatars

alexkuz:
React v0.14 doesn't support `null`s as render return value in stateless components.

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/Entry/Avatar/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/Entry/Avatar/index.js
@@ -11,7 +11,7 @@ const Avatar = ({ mention, theme = {} }) => {
     );
   }
 
-  return null;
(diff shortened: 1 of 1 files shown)


nikgraf:
@alexkuz any chance you can upgrade to React 15? I believe we use `null` in a couple of places. Having React 15 as lowest dependency would be my favourite right now.

alexkuz:
I haven't try yet, but maybe it's time :)

alexkuz:
Turns out I can't switch right away (sigh). Anyway, it would be nice to make a temporary fix in current version (as long as it declares to support `react@0.14.8`) and switch to `15.0` in next major version.

nikgraf:
@alexkuz Just cut a new release & published it on npm https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/CHANGELOG.md#102---2016-05-24

PR #271, opened 2016-05-20
Support `onClick` handler in mentions

alexkuz:


nikgraf:
@alexkuz With your PR the onClick handler would need to be in the data. I would like to have a clear separation between behaviour related options & data.

We could do something similar to this PR for the mentions: https://github.com/draft-js-plugins/draft-js-plugins/pull/273/files

alexkuz:
Well, `href` is also kinda behaviour prop, but ok :)
Maybe, instead of props, allow to replace the component?
```js
const {
    mentionPrefix = '',
    theme = defaultTheme,
    positionSuggestions = defaultPositionSuggestions,
    mentionComponent
  } = config;

...

component: decorateComponentWithProps(Mention, { theme, mentionPrefix, mentionComponent }),
```
```jsx
  if (mentionComponent) {
    const Component = mentionComponent;
    return (
      <Component {...this.props} mention={ mention } />
    );
  }

  return (
    <span
      className={ theme.mention }
      spellCheck={ false }
    >
      { props.mentionPrefix }{ props.children }
    </span>
  );
```

nikgraf:
Quite an interesting idea! @antoinerey started to work on a way to pass in props for the link plugin here: https://github.com/draft-js-plugins/draft-js-plugins/pull/273. Might nice to have something useful & still consistent across all plugins.

What are upsides & sides between these two patterns? Is there a benefit of having both? Otherwise we just offer two ways of doing the same thing and that would make it more complex than necessary and harder to explain in docs.

antoinerey:
Correct me if I'm wrong. 

What you're suggesting here is, instead of giving each props manually, we could give a whole component with props already set on it, to the draft js plugin.

So instead of :

``` js
const linkify = linkifyPlugin({
  linkProps: {
    style: { color: 'red' },
    title: 'My link',
  },
})
```

We could have :

``` js
const linkify = linkifyPlugin({
  linkComponent: (
    <Link style={ ... } title="My link" />
  )
})
```

It would offer more flexibility but what if the developer gives a **display block** component instead of a **display inline** one ? This change will affect the layout which can be totally broken.

adrianmc:
I feel like we really should pick one. Perhaps all plugins should be given the option to use your own component.

alexkuz:
> It would offer more flexibility but what if the developer gives a display block component instead of a display inline one ? This change will affect the layout which can be totally broken.

Well, if you explicitly set block element, you probably know what you're doing? I'm not sure why it should be bulletproof here.

nikgraf:
It gets a bit tricky in the details. For example a mention contains subcomponents like `Avatar`, but on the other hand we don't have to expose all of them. Just having `mentionComponent` exposed as option might be enough.

@antoinerey I totally feel your concerns with breaking the layout. Some components have quite some tricky details. On the other hand I expect a ton of feature requests over the time asking for various cases like `onTouchStart` or a certain ARIA attribute. By exposing the component we give them all the freedom , but I think a super good documentation will be required showing how to use them properly.

Right now I'm super busy preparing for React Europe. I'm happy to do some tests afterwards. If you guys are up for some experiments feel free to open a PR 🙌

oyeanuj:
+1 for having your own component. I think it makes it easier to retrofit in existing apps where that functionality (dropdown or mention) might already have been built and is easier to just plug that in than use a component here and try to make it work exactly like something existing.

adrianmc:
Has nobody been working on this? I might have some time at the end of this month and I'd be glad to take this up.

alexkuz:
@adrianmc I have a version, actually, that I use in my project, just didn't commit it yet

nikgraf:
@alexkuz a version where you replace the whole component or something else?

@adrianmc in a week or two I should have more time for DraftJS Plugins again 😄

alexkuz:
@nikgraf yeah, the whole component.

antoinerey:
@alexkuz Is there any code on Github ?

I'm still afraid by replacing the whole component just to pass down one or two props. Maybe it could convince me that's the way to go ;)

nikgraf:
@alexkuz I'm pretty convinced now this is the way to go. Do you want open a PR with what you have?

alexkuz:
@nikgraf sorry for delay! I've updated PR, take a look.

alexkuz:
@nikgraf ping

nikgraf:
thx @alexkuz, looks great! 👍

PR #272, opened 2016-05-21
Plugin editor changes for wysiwyg plugins

bkniffler:
Hey, 
Here are some changes/additions I've created while working on the wysiwyg plugins.

I've removed the block decorator feature, as these will unmount/remount blocks on rerender.

I'm passing all plugin editor properties and methods to plugin methods. I know this has been discussed earlier, I now think this makes sense. Also I'm storing all plugins in this.plugins, which makes them available to plugin methods (for example to check if a certain plugin is available while initializing a plugin).

Also, I've created setReadOnly/getReadOnly to change readOnly through the state. This is crucial for nested editors, as while a nested editor is active the outer editor needs to be set to readOnly=true.

### DIFF
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -34,8 +34,8 @@ class PluginEditor extends Component {
   constructor(props) {
     super(props);
 
-    const plugins = [this.props, ...this.resolvePlugins()];
-    for (const plugin of plugins) {
+    this.plugins = [this.props, ...this.resolvePlugins()];
+    for (const plugin of this.plugins) {
       if (typeof plugin.initialize !== 'function') continue;
       plugin.initialize({
         getEditorState: this.getEditorState,
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -37,10 +37,7 @@ class PluginEditor extends Component {
     this.plugins = [this.props, ...this.resolvePlugins()];
     for (const plugin of this.plugins) {
       if (typeof plugin.initialize !== 'function') continue;
-      plugin.initialize({
-        getEditorState: this.getEditorState,
-        setEditorState: this.onChange,
-      });
+      plugin.initialize({ ...this });
     }
 
     // attach proxy methods like `focus` or `blur`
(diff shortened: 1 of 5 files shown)


nikgraf:
@bkniffler nice, glad to see the block decorator feature being deprecated!

Which methods/props do you need to access? I would rather prefer to expose them in an object than passing in the whole thing. Otherwise every internal change basically would lead to a breaking change.

Could we also expose: `getPlugins`? Saving the plugins in `this.plugins` is something we should really try to avoid as it is hard to identify the "source of truth". Not the same, but similar is described here in detail: https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html

bkniffler:
Yeah, sorry for having that introduced in the first place.

Your argument makes total sense. I need to expose all props though, so a nested editor can be initialized with the same props as the parent editor. What would you recommend? A `getProps` method? This doesn't seem like the most elegant solution.

I've named the method `getPluginMethods`, some alternatives might be `getEditorInterface`, `getPluginInterface`, `getEditorConnection`, `getPluginConnection`, ...

aeman1:
On 22/05/2016 7:31 am, "Nik Graf" <notifications@github.com> wrote:
>
> @bkniffler nice, glad to see the block decorator feature being deprecated!
>
> Which methods/props do you need to access? I would rather prefer to
expose them in an object than passing in the whole thing. Otherwise every
internal change basically would lead to a breaking change.
>
> Could we also expose: getPlugins? Saving the plugins in this.plugins is
something we should really try to avoid as it is hard to identify the
"source of truth". Not the same, but similar is described here in detail:
https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html

aeman1:
Saying thank you to all concerned putting and tweaking helping frustrated
bitch here. wont get now though over time ill learn in the most pleasant
way.
also have other explicit ideas if your possibly interested
Sincerely
Adam Reynolds
m. 0457193155 anytime anywhere lol

nikgraf:
@bkniffler `getPluginMethods` sounds fine to me & I can't come up with a better name right now 😄

Do we really need to pass in all the props? If we can make this a fixed list, would be cool, but not super important.

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -66,23 +65,28 @@ class PluginEditor extends Component {
...
     }
   };
 
   getEditorState = () => this.props.editorState;
+  getPluginMethods = () => ({
+    getPlugins: () => this.props.plugins.slice(0),
+    getProps: () => ({ ...this.props }),
+    setEditorState: this.onChange,
+    getEditorState: this.getEditorState,
+    getReadOnly: () => this.props.readOnly,
```
nikgraf: Further down we have `readOnly={this.props.readOnly || this.state.readOnly}`. Why is `readOnly` here just from the props? Why would plugins use this instead of just taking it from `getProps`?

nikgraf:
We also need to extend the data for `willUnmount` https://github.com/draft-js-plugins/draft-js-plugins/pull/287

nikgraf:
@bkniffler how can I get in touch with you? I need access to the plugins on `npm` to maintain them & publish new version.

PR #273, opened 2016-05-22
feat(props): Offer a way to customize the link component props

### LINKED ISSUE #269: Inline styles
I started to use your plugins today, and I'd like to apply some custom inline styles but from what I saw, it seems not possible. Is there any reason why the style overrides can only be made with CSS and not inline styles ? 

I went through some of your code, and it seems that you can use `decorate-component-with-props` to pass down any props from the `config` object.

Some code may be easier to understand :

``` js
const linkPlugin = (config = {}) => {
  // Any other key are considered as props
  let { theme, target, ...props } = config;

  theme = theme ? theme : defaultTheme;
  target = target ? target : '_self';

  return {
    decorators: [
      {
        strategy: linkStrategy,
        // We pass down the props to the Link component.
        component: decorateComponentWithProps(Link, { theme, target, ...props }),
      },
    ],
  };
};
```

This way, one can override styles with inline ones, add some ARIA attributes or whatever.

What are you thoughts about it ? I could make a PR if you want to.

PS: I didn't test my idea with real code, so I could have miss something :)

antoinerey:
The linkify plugins configuration now offers a way to specify custom
props to be passed down to the Link component. This way, one can
override the appearance with inline styles, setup event handlers or
whatever.

I kept the target configuration out of the linkProps to keep the current
API working.

Let me know if that's okay for you, and I'll make the same changes to
every plugins.

Closes #269

### DIFF
--- a/draft-js-linkify-plugin/src/index.js
+++ b/draft-js-linkify-plugin/src/index.js
@@ -14,13 +14,17 @@ const linkPlugin = (config = {}) => {
   // styles which needs a deep dive into the code. Merging also makes it prone to
   // errors when upgrading as basically every styling change would become a major
   // breaking change. 1px of an increased padding can break a whole layout.
-  const theme = config.theme ? config.theme : defaultTheme;
-  const target = config.target ? config.target : '_self';
+  const {
+    theme = defaultTheme,
+    target = '_self',
+    linkProps = {},
+  } = config;
+
   return {
     decorators: [
       {
         strategy: linkStrategy,
-        component: decorateComponentWithProps(Link, { theme, target }),
+        component: decorateComponentWithProps(Link, { theme, target, ...linkProps }),
       },
     ],
   };


adrianmc:
@antoinerey LGTM! Nice work. Do you think you can provide example usage in the docs as well?

draft-js-linkify-plugin/src/index.js
```diff
@@ -14,13 +14,17 @@ const linkPlugin = (config = {}) => {
...
+    target = '_self',
+    linkProps = {},
+  } = config;
+
   return {
     decorators: [
       {
         strategy: linkStrategy,
-        component: decorateComponentWithProps(Link, { theme, target }),
+        component: decorateComponentWithProps(Link, { theme, target, ...linkProps }),
```
nikgraf: Can you change it to `{ ...linkProps, theme, target }`? This way we make sure theme & target are not overwritten by values in the custom linkProps.

draft-js-linkify-plugin/src/index.js
```diff
@@ -14,13 +14,17 @@ const linkPlugin = (config = {}) => {
   // styles which needs a deep dive into the code. Merging also makes it prone to
   // errors when upgrading as basically every styling change would become a major
   // breaking change. 1px of an increased padding can break a whole layout.
-  const theme = config.theme ? config.theme : defaultTheme;
-  const target = config.target ? config.target : '_self';
+  const {
+    theme = defaultTheme,
+    target = '_self',
+    linkProps = {},
+  } = config;
```
nikgraf: Nice, way better 🙌

nikgraf:
Thx for the change @antoinerey 

Would you mind also update the docs & the changelog, no worries if it's too much hassle 👍

draft-js-linkify-plugin/src/index.js
antoinerey: Oh yup ! Good catch

antoinerey:
Yep, I'll do that.

nikgraf:
@antoinerey porting it to other plugins would be great! I suggest we do it one by one with separate PRs. Otherwise it gets easier to miss something in the review

nikgraf:
@antoinerey What do you think about @alexkuz latest idea in https://github.com/draft-js-plugins/draft-js-plugins/pull/271?

nikgraf:
I close this for now & we can discuss in #217

PR #274, opened 2016-05-23
Fixes and release notes

nikgraf:


draft-js-emoji-plugin/src/EmojiSuggestions/index.js
```diff
@@ -247,7 +247,7 @@ export default class EmojiSuggestions extends Component {
 
   render() {
     if (!this.state.isActive) {
-      return null;
+      return <noscript />;
```
gabeaudick: Curious: why return `<noscript />` instead of `null`?

nikgraf: One user is still using React 0.14. I cut a release to help him here. In the next version we will go back to `null`. 😄

PR #275, opened 2016-05-24
keep tooltip visible within window

jl-:
when tooltip get too wide for vertical-align to the target, it would overflow to the left of window, some parts of it become invisible. this patch sets `state.left = 0`
qq20160524-2 2x


jl-:
Sorry for the mess, I have no idea how the `93 commit` happened.
qq20160524-1 2x. I just edit it online. may be I would better just close this pr.

PR #276, opened 2016-05-24
fixed block type naming mistake

jl-:
[DraftBlockType](https://github.com/facebook/draft-js/blob/master/src/model/constants/DraftBlockType.js)


nikgraf:
Looks good to me 👍 
@bkniffler you are way more familiar with these parts, can you review?

PR #277, opened 2016-05-24
[wysiwyg:toolbar] keep tooltip visible within window

jl-:
when tooltip get too wide for vertical-align to the target, it would overflow to the left of window, some parts of it become invisible. this patch sets state.left = 0
qq20160524-2 2x

### DIFF
--- a/draft-js-toolbar-plugin/src/utils/portal.js
+++ b/draft-js-toolbar-plugin/src/utils/portal.js
@@ -39,21 +39,19 @@ class Tooltip extends Component {
       const refRect = ref.getBoundingClientRect();
       const scrollY = window.scrollY ? window.scrollY : window.pageYOffset;
       const scrollX = window.scrollX ? window.scrollX : window.pageXOffset;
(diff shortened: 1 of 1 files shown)


nikgraf:
@bkniffler what do you think?

bkniffler:
Thanks for the PR! Its true this hasn't been taken into account and I think such a functionality would absolutely make sense, though we'd need it for both sides left and right. What do you think @jl- ? Could you add right-side support to your solution?

jl-:
qq20160526-0 2x
The `toolbar-item` is left-floated, and their containing block `toolbar` has no `width` set, so the width of `toolbar` is the same as its containing block `data[react-root]`, which is `absolute` positioned, then its containing block is either the initial containing block or `body`(if positioned). So there should be no such case that the width of `data[react-root]` would exceed the width of viewport, instead, the `toolbar-item` would float in the second line, as shown in the picture, therefore, with `left` set to `0`, that should be enough. (correct me if I'm wrong 😄 )

Issue #259, opened 2016-05-09
Ability to test if popover is open or closed.


stefan1968:
Is there some way of testing if the mention or emoji window is open, some event handler?

Many thanks.

nikgraf:
@stefan1968 not right now, what's your use-case? When would it be useful?

stefan1968:
Ok, for example i'm inputting text and on my handlereturn i need to know weather to invoke the plugins handlereturn (or mine) depending if the popup is open, if i just call mine with the popup open because it resets the editor focus and creates a new state etc, it can cause strange behaviour with the plugin

mjrussell:
👍 for this. I ran into a similar issue when I wanted to handle the enter key in the component that wraps the plugin editor differently depending on if the popover was open or closed. In my case I wanted to submit a comment on enter if the popover is not open.

I found an entirely hacky/temporary solution to this by putting a ref on the suggestion component and doing the following:
```js
this.refs.suggestions._reactInternalInstance._renderedComponent._instance.state.isActive;
```

I think we need to add a `getWrappedInstance` to the `decorate-component-with-props` wrapper which returns the rendered ref (this is common in HOC components like redux's `connect`) and then just expose a public method on the suggestions component for querying if the state is open

crossman:
For what it's worth, I ran into the same thing and this was my temporary, hacky solution (along with a large comment)

```js
if (this.refs.editor.refs.editor.refs.editor.getAttribute("aria-haspopup") == "true")
        return
```

But yeah, I'd love to see a better way to handle this.

nikgraf:
What if we would introduce `onOpen/onClose` on MentionSuggestions itself? Would that solve the issue or did I miss something?

crossman:
That could work. I'd have to listen to those and keep track of the state internally for use in handleReturn, but it would create more opportunity to react to changes in the interface. Anything I can do to help? I haven't looked over the internals much but I really appreciate this project and would be happy to contribute.

mjrussell:
Yeah that should work, should this be a controlled state? 

Like passing the visible flag to it and using the onChange to update the parent state? Or is this just a helper/utility

nikgraf:
@mjrussell I though of it just as a callback and not as controlled state. The state should be controlled by the decorator. To me that's a core-value of the plugin, but I'm curious if there are other case.

@crossman a PR just exposing those in the Emoji & Mentions plugin would be great. Do you know how to do that or should we guide you?

[...]

PR #279, opened 2016-05-24
adding onOpen/onClose events for MentionSuggestions and EmojiSuggestions

crossman:
This takes care of #259 by providing hooks into when the Mention and Emoji suggestions popovers are opened or closed


nikgraf:
Well done @crossman - thanks for updating the Changelogs 🙌

PR #292, opened 2016-06-02
Fix emoji background styles.

gabeaudick:
I'm seeing mispositioned emojis in my projects and in the [draft-js-plugins](https://www.draft-js-plugins.com/) example editor as of an hour ago.

Notice the oversized, askew *(nerd)* and *(tada)* emojis.

I can repro in chrome and firefox - safari isn't polyfilled, errs out, and just renders the unicode character - and some slack peeps were able to repro as well.

I couldn't find any glaring, recent change that would have caused this but haven't had much time to look 😋 

My solution here is just to *center* and *contain* the emoji background images.

I'll follow up with `changelog` changes if this is indeed valid and useful :)

### DIFF
--- a/draft-js-emoji-plugin/src/emojiStyles.css
+++ b/draft-js-emoji-plugin/src/emojiStyles.css
@@ -1,13 +1,15 @@
 .emoji {
+  background-position: center;
   /* make sure the background the image is only shown once */
   background-repeat: no-repeat;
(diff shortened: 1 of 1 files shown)


nikgraf:
thx @gabeaudick - good job 👌

nikgraf:
just release a new version with your changes 🙌

PR #301, opened 2016-06-10
Feature/lerna

joehua87:


juliankrispel:
this seems really cool but massively out of date o_O, I'm thinking about closing this, any objections @joehua87 @nikgraf ?

nikgraf:
Yeah, sounds good. I wanted to keep it around as a reminder to do it at some point, but there is no point 😄

PR #304, opened 2016-06-14
Add groups to MentionSuggestion popup

tonypee:
Ive added the ability to add a 'type' field to the suggestions. 
- Types are split into groups in the popup
- groups are not shown when there is only one
- items without a type fall into an 'other' group

I tried to keep it simple by not creating a Group component and working directly in the MentionSuggestions. I could look to refactoring it into 'Groups' if you like. 

It might be nice to have the 'other' group always show up last

It should hopefully work as before when no 'types' are set 

All style, refactoring changes welcome, It's a pretty quick pass

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/index.js
@@ -5,6 +5,7 @@ import addMention from '../modifiers/addMention';
 import decodeOffsetKey from '../utils/decodeOffsetKey';
 import { genKey } from 'draft-js';
 import getSearchText from '../utils/getSearchText';
+import Immutable from 'immutable';
 
 export default class MentionSuggestions extends Component {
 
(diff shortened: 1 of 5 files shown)


draft-js-mention-plugin/src/mentionSuggestionsStyles.css
```diff
@@ -16,3 +16,13 @@
   box-sizing: border-box;
   transform: scale(0);
 }
+
+.mentionSuggestionsHeading:global(.first) {
```
nikgraf: Am I'm correct that `:global` would create conflict if people use multiple mention plugins with different themes.

draft-js-mention-plugin/src/MentionSuggestions/index.js
```diff
@@ -261,6 +262,19 @@ export default class MentionSuggestions extends Component {
...
+      const type = value.get('type') || 'other';
+      let items = data.get(type);
+      if (!items) {
+        items = Immutable.OrderedSet();
+      }
+      items = items.add(value);
+      return data.set(type, items);
+    }, Immutable.OrderedMap());
+
+    const groupTitles = Object.assign({ other: 'Other' }, this.props.groupTitles);
```
nikgraf: can you use the { ...this.props.groupTitles, other: 'Other' } instead of Object.assign?

draft-js-mention-plugin/src/mentionSuggestionsStyles.css
nikgraf: that should work and then you also can get rid of the `.first` class: https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child

nikgraf:
Nice, can you work on the proposed improvements?

nikgraf:
Hey @tonypee, still wanna work on this or rather drop it?

nikgraf:
I will close this for now, but feel free to open a PR in the future if you or someone else has time to clean it up 👍

Issue #289, opened 2016-05-30
Mention Suggestion Shows up in Unrelated Spaced Position


bekkou68:
I found a mention suggestion shows up in unrelated spaced position.
The problem seems occurred when I type a suggested name directly and type the space key.
And it seems solved when I type the enter key to fix a HTML element.

<image src="https://cloud.githubusercontent.com/assets/322852/15644877/67e67f98-2690-11e6-9c15-a80545843ef3.gif" width="50%" />

nikgraf:
hey @bekkou68 - this might be related to: https://github.com/draft-js-plugins/draft-js-plugins/issues/206#issuecomment-219192040 and https://github.com/draft-js-plugins/draft-js-plugins/blob/master/FAQ.md#why-is-there-no-popover-for-mentionsemoji-plugin

Your editor might be nested inside a styled element which causes the troubles. Can you verify if that's the case?

nikgraf:
fixed with https://github.com/draft-js-plugins/draft-js-plugins/pull/309 & going to be in the next release

nikgraf:
I close this issue for now. Let me know in case it didn't work after the next release.

hasnainshafqatmlt:
Hello guys, 
will you please help me how i can change it's position from bottom to top ?
i want to show mention list on top of the text instead of bottom.

![screen shot 2017-11-06 at 8 19 00 pm](https://user-images.githubusercontent.com/6399158/32448072-c217ad4c-c32f-11e7-8aa7-f349695c9218.png)

Issue #283, opened 2016-05-25
error with fixed parent


berdof:
First of all I am very excited with this plugin.
But I have found one issue. When you place editor into block with position fixed you ran into emoji modal strange positioning.
You can simply test it on demo site 
![image](https://cloud.githubusercontent.com/assets/1468790/15535469/31848454-2276-11e6-80ef-bd0b526e0c58.png)

mjrussell:
@berdof you can override the position suggestions function that is passed into the mentions plugin.

This FAQ explains how to override that function: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/FAQ.md#why-is-there-no-popover-for-mentionsemoji-plugin

Here is an example position suggestions function that worked for me:
https://github.com/draft-js-plugins/draft-js-plugins/issues/206#issuecomment-219192040

nikgraf:
@berdof glad you like it 😄 
thx @mjrussell 👍

brigand:
I just ran into this, with a position:relative parent. I'm not sure how to solve it without the dom node and walking up the parent tree searching for a non-position:static element. Is the dom node provided to positionSuggestions?

It'd be great if draft-js-plugins handled this case for us.

nikgraf:
fixed with https://github.com/draft-js-plugins/draft-js-plugins/pull/309 & going to be in the next release

Issue #206, opened 2016-04-18
SearchSuggestions position is not working properly when inside an overflowing div


Aleksion:
I'm not sure where exactly the issue is happening, but the SearchSuggestions PopOver's location doesn't seem to be working inside my overflowing scrollable div. The previous version of the mention plugin worked just fine position wise though. I'd like to debug it myself, but I can't quite wrap my head around the current position logic - and any click event (even ones occuring in chrome's console), dismisses the popover again. 

![screeenshot](https://cloud.githubusercontent.com/assets/3367641/14608135/c0b038bc-0584-11e6-843f-f281705c85e3.png)

nikgraf:
oh, thx for reporting @Aleksion 👍 

Seems like I didn't to the positioning properly. I will play with it & try to fix it. btw the position logic is here in case you have time to debug: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/src/SearchSuggestions/index.js#L42

 Can you elaborate a bit more on: "and any click event (even ones occuring in chrome's console), dismisses the popover again." Maybe a gif or video to demonstrate the issue?

nikgraf:
@Aleksion btw I still haven't forgotten your other requests 😄 sorry, other stuff was more important to fix first …

nikgraf:
@Aleksion I tried to put it into a div with `overflow-y: scroll; height:400px;`, but couldn't reproduce the issue. Can you create a working gist or so?

wbf1013:
I also have this problem, becasue of parent element have style of  `position:relative`  ,then the SearchSuggestions move other position!
<img width="500" alt="_20160508200317" src="https://cloud.githubusercontent.com/assets/1181144/15097737/f4d82540-1557-11e6-8e84-b931c00fcda4.png">

nikgraf:
@wbf1013 thx for investigating! That makes a lot of sense now. Ideally we could update [this function](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/src/utils/positionSuggestions.js) to make it work with `position: relative`.

Anyone interested to fix this?

mjrussell:
@wbf1013 @nikgraf 

[...]

PR #309, opened 2016-06-16
fix mentions positionSuggestions to work with non-static parents

Zhouzi:
This PR fixes mentions' `positionSuggestions` function to work with non-static parents, related to #283 #206 #289.

Note: it introduces a "regression" on screens smaller than 480px. There's used to have a condition that would align the popover to the left in such circumstances. The "universal" implementation would be more complex as it'd need to look for the relative parent's position in the viewport plus some other stuff. I wasn't sure if it's something that should be part of the default implementation. It's probably going to be a requirement for us though, so I'll add it anyway.

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/index.js
@@ -49,6 +49,7 @@ export default class MentionSuggestions extends Component {
         decoratorRect,
         prevProps,
         prevState,
+        popover: this.refs.popover,
         props: this.props,
         state: this.state,
       });
--- a/draft-js-mention-plugin/src/utils/positionSuggestions.js
+++ b/draft-js-mention-plugin/src/utils/positionSuggestions.js
@@ -1,30 +1,53 @@
-const positionSuggestions = ({ decoratorRect, state, props }) => {
-  const scrollTop = window.pageYOffset || document.documentElement.scrollTop;
-  const scrollLeft = window.pageXOffset || document.documentElement.scrollLeft;
-
-  let width;
-  let left;
-  let transform;
-  let transition;
-  if (window.innerWidth <= 480) {
-    left = '20px';
-    width = `${window.innerWidth - 40}px`;
-  } else {
-    left = `${decoratorRect.left + scrollLeft}px`;
+const getRelativeParent = (element) => {
+  if (!element) {
+    return null;
   }
 
-  if (state.isActive & props.suggestions.size > 0) {
-    transform = 'scale(1)';
-    transition = 'all 0.25s cubic-bezier(.3,1.2,.2,1)';
-  } else if (state.isActive) {
-    transform = 'scale(0)';
-    transition = 'all 0.35s cubic-bezier(.3,1,.2,1)';
+  const position = window.getComputedStyle(element).getPropertyValue('position');
+  if (position !== 'static') {
+    return element;
+  }
+
+  return getRelativeParent(element.parentElement);
+};
+
+const positionSuggestions = ({ decoratorRect, popover, state, props }) => {
+  const relativeParent = getRelativeParent(popover.parentElement);
+  const relativeRect = {};
+
+  if (relativeParent) {
+    relativeRect.scrollLeft = relativeParent.scrollLeft;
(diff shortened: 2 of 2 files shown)


nikgraf:
This is really good, thanks for your effort! 👌

nikgraf:
Looking forward to the improvements for the smaller screens.

Zhouzi:
I'll work on it tomorrow and push the changes ;)

flgarrid:
This should also work with the emojiplugin, right?

Zhouzi:
@flgarrid yeah, @nikgraf did port the implementation to the emoji plugin too via https://github.com/draft-js-plugins/draft-js-plugins/commit/7c983a2255d76b48ba8cab75adc6890faf24724a 👍

jssisodiyaPG:
@nikgraf I recently updated emoji plugin from `1.2.1` to `1.2.2`. After this it has started throwing this error in `postionSuggestions` on rendering smileys
screen shot 2016-07-04 at 3 28 16 pm
Previous version of emoji editor had no `suggestions` props. Let me know what is wrong here.

mjrussell:
I'm actually a bit curious if the else case is even necessary. I think the tweaks I added to mention and suggestions should make that when the suggestion list is empty it will set the state to closed.

Agree would be nice to consolidate these. One day maybe the auto complete plugin creator!

Zhouzi:
@jssisodiyaPG this just got fixed by bac8c30f5e324f1fa13b11eeecbaec9172adeb58. Depending on how you're tracking your dependencies, @nikgraf may have to publish a release.

@mjrussell I must admit that it's unclear to me why `positionSuggestions` is called if the list is empty so it's great if you've managed to remove those calls, we should clean out the function. While fixing the issue mentioned above, I've also seen that `positionSuggestions`'s signature is a bit different between the emoji and the mention plugin which is probably what tricked Nik ^^

jssisodiyaPG:
@Zhouzi Thanks for the quick resolution. I install my dependencies though npm packages

jssisodiyaPG:
@Zhouzi Let me know publish the fix to npm. Need to deploy the plugin to one of my feature.

nikgraf:
@jssisodiyaPG just published version 1.2.3 of the emoji plugin

nikgraf:
thx for the fix @Zhouzi! 👏

mkampik:
Just a heads up that version `2.0.0-beta-1` of the emoji plugin still throws the `props.suggestions === undefined` error described above. This is because `1.2.3` was released after `2.0.0-beta1`:

`$ npm view draft-js-emoji-plugin@2.0.0-beta1 time`

returns

```
{ modified: '2016-07-06T21:17:31.559Z',
  created: '2016-03-23T15:18:29.682Z',
  '0.0.1': '2016-03-23T15:18:29.682Z',
  '0.0.2': '2016-03-25T20:08:14.578Z',
  '0.0.3': '2016-03-26T05:02:33.226Z',
  '0.0.4': '2016-03-29T08:25:20.460Z',
  '1.0.0-beta1': '2016-04-16T01:15:20.875Z',
  '1.0.0': '2016-04-20T11:14:58.794Z',
  '1.0.1': '2016-04-29T14:45:27.619Z',
  '1.1.0': '2016-05-08T22:32:49.444Z',
  '1.1.1': '2016-05-23T23:34:29.113Z',
  '1.2.0': '2016-05-27T19:44:15.802Z',
  '1.2.1': '2016-06-05T20:06:38.106Z',
  '1.2.2': '2016-06-26T08:10:46.810Z',
  '2.0.0-beta1': '2016-06-28T22:29:10.951Z',
  '1.2.3': '2016-07-06T21:17:31.559Z' }
```

I would install `1.2.3`, however this throws 

`err  Installing npm:draft-js-emoji-plugin@1.2.3, no version match for npm:react@15.0.0-rc`

as this versions peerDependencies for both `react` and `react-dom` are still set to `>=0.14.0 || 15.0.0-rc`.

Would it be possible to update v `2.0.0-beta1` with @Zhouzi fix? Can't say I'm familiar with npm versioning... If not, would it be worth publishing a new version with the fix included? Not sure when you guy's are aiming to release v 2.0.0 (non-beta).

nikgraf:
@mkampik I will publish a new beta today

PR #318, opened 2016-06-21
fix mention's suggestions key to allow suggestion with the same name

### LINKED ISSUE #308: Suggestions with the same name shows a warning and drops "duplicates"
Hey there!

I am using the mention plugin with an api that returns users with different ids but sometimes the same name. So I may end up with: `{ id: 'john-doe', name: 'John Doe' }, { id: 'john-doe-1', name: 'John Doe' }` which leads to:

> Warning: flattenChildren(...): Encountered two children with the same key, `John Doe`. Child keys must be unique; when two children share a key, only the first child will be used.

It'd be great if we could provide a custom "keyGetter".

The current implementation is there: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/src/MentionSuggestions/index.js#L275

Zhouzi:
Fixes #308

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/index.js
@@ -8,6 +8,7 @@ import getSearchText from '../utils/getSearchText';
 import defaultEntryComponent from './Entry/defaultEntryComponent';
 
 export default class MentionSuggestions extends Component {
+
   static propTypes = {
     entityMutability: PropTypes.oneOf([
       'SEGMENTED',
@@ -284,7 +285,7 @@ export default class MentionSuggestions extends Component {
         {
           this.props.suggestions.map((mention, index) => (
             <Entry
-              key={mention.get('name')}
+              key={mention.has('id') ? mention.get('id') : mention.get('name')}
               onMentionSelect={this.onMentionSelect}
               onMentionFocus={this.onMentionFocus}
               isFocused={this.state.focusedOptionIndex === index}


jarindr:
agreed on this PR, it doesn't have to be `{this.key}` since even that we can't access the key in the object anyway. 
i'm currently having a problem to identify each mentioned person with its key to to create a notification for each person individually.
for now, you can do it with adding an ID field in the mention suggestion object field and access it in Entitymap object.
Any suggestion to  identify each mentioned person with its key other than this method?

Zhouzi:
A solution could be to add an option to provide a key generator: a function that would receive the mention and return an unique identifier.

draft-js-mention-plugin/src/MentionSuggestions/index.js
```diff
@@ -256,6 +255,24 @@ export default class MentionSuggestions extends Component {
     }
   };
 
+  renderSuggestion = (mention, index) => {
+    const { theme = {} } = this.props;
+    const key = genKey();
```
nikgraf: If we generate the key for every entry on every render React might not complain, but we end up with more DOM manipulations. The point is to use a unique identifier as key so React can reuse structures already existing in the DOM. Read more here: https://facebook.github.io/react/docs/multiple-components.html#dynamic-children

In addition the mention option is important for accessibility. That's the reason why we need this id attribute. Please don't change that.

Ideally we have some kind of mechanism that creates a unique ID for each mention.

Zhouzi: Yeah, you're right, I've been a little quick on that one. I'll add a key generator option with a default of `mention.get('name')` and fix the missing id ;)

Zhouzi:
@nikgraf I've added an option to provide a custom `getMentionKey` with a default of `mention.get('name')`, which is the current behavior. I also updated the documentation but I am not really sure if I have to go through other steps to generate it properly.

@jarindr I re-read your question and starting to wonder if you are really looking for a custom key getter? Would you mind clarifying what you mean by:

> Any suggestion to identify each mentioned person with its key other than this method?

Zhouzi:
@nikgraf I've synced the branch and it's now all clean, are you okay to merge it?

nikgraf:
@Zhouzi what about we check if the mention has an `id` and take the `id`. If not we fall back to `name`. I hope this would cover 99% of the use-cases especially if we use `id` in the examples as well. For the rest of the people we recommend to fork it and manipulate it as well.

I would like to keep such React specific functions away from the API surface, but maybe at some point in the future `key` might not be even needed … Does that make sense?

Zhouzi:
Makes sense, I'll make the changes ;)

Zhouzi:
@nikgraf here you go :)

nikgraf:
@Zhouzi sweet, merged

PR #319, opened 2016-06-21
add support for white space within mention

### LINKED ISSUE #316: Mention plugin to allow white spaces
I am dealing with an API that returns a lots of results so mentioning someone from their first name can be tricky. I'd like to be able to type `John Do` so `John Doe` pops out but right now the popover disappear after a white space is inserted.

I'm unsure how tricky that would be to implement, I've been looking at https://github.com/draft-js-plugins/draft-js-plugins/blob/9f94e5c94d4b08605d9ab227218f141d8bee2cf3/draft-js-mention-plugin/src/MentionSuggestions/index.js#L102 and will probably dig deeper but feel free to let me know if you have already worked on it.

Zhouzi:
This PR fixes #316 by adding support for white spaces within a mention. It's introducing a breaking change though: there's no limit to the number of white spaces allowed so given `hey @john doe how are you doing?` it's going to call `.onSearchChange` with `john doe how are you doing`. That's actually what Facebook does, the only thing to stop the autocomplete dropdown to show up is the service not returning results. A first step could be to add a limit of 3 white spaces so `john doe how` is ok while `john doe how are you doing` is not.

* [x] mentionSuggestionsStrategy should match only one space between words (`@john doe` is ok while `@john    doe` is not)
* [x] mentionSuggestionsStrategy should not match trailing white spaces if maxWordsSearch is disabled (set to 1)
* [ ] If there's a trigger character in a line, adding a new one further doesn't work


Zhouzi:
*This PR is not ready to be merged, I am still working on it but let me know what's your opinion.*

nikgraf:
I believe the default options should still be the current behaviour as this is what a lot of platforms like support e.g. Twitter, Slack

I'm happy to have this in as an additional option 👍

Zhouzi:
@nikgraf I've moved the white space support to an option called `maxWordsSearch ` with a default of `1`, meaning the current behavior is preserved. There's a little change to be noted, though:

Until now, the text search was returning the word surrounding the cursor, so in this situation: `@ma|tt` (`|` being the cursor) it was calling `onSearchChange` with `matt`. With this PR it'd call it with `ma` which is the text that's preceding the cursor. So if I type `@matt` and move my cursor around, the results will vary according to its position. That's what Facebook does, Slack simply hides the dropdown (when pressing the left arrow) and GitHub seem to search for the word surrounding the cursor (implementation before this PR). Seems to reflect what the user may be looking for by moving his cursor in the word but let me know what you think.

Oh and I'll indeed add the relevant documentation if this PR respects your vision.

nikgraf:
@Zhouzi ❤️ your attention to detail - didn't notice the Facebook behaviour so far …

Zhouzi:
Hehe 😄 

There's a few more things to be noted:

* [ ] The current mentionStrategy matches `@john    doe` (infinite number of spaces between words)
* [ ] The current mentionStrategy match trailing white space `@john ` (it wasn't the case before)
* [ ] Not sure about this one but the autocomplete doesn't seem to show up the first time the query is empty

I'll fix those points soon.

nikgraf:
@Zhouzi any news on the 3 points you mentioned? happy to do some pair programming /w you this week

Zhouzi:
I'll see if I can get those fixed in the weekend but if not we can set up a pair programming session 👍

Zhouzi:
Ok so I've fixed the remaining bugs except one that I'm struggling with. The issue is that typing the trigger character, not selecting a suggestion and typing a new one further on the same line doesn't open the suggestions. For example: `hey @matt` opens the suggestions but not the second @: `hey @matt did you get in touch with @nik` (won't suggest Nik Graf).

The problem comes from the fact that the new regular expression matches trailing white spaces:

```javascript
'hey @matt did you get in touch with @nik'.match(/(\s|^)@((\w+\s?){0,})?/g);
// [" @matt did you get in touch with "]
//
// whereas we'd like to have:
// [" @matt did you get in touch with ", " @nik"]
```

I tried to tweak the expression without success. The only workaround I found so far is to go through two steps in the mentionSuggestionsStrategy:

1. Find the last occurence of the mention with a simpler regexp (e.g `/(^|\s)@/g`)
2. In that last occurence, find the start/end of the interesting part with the white spaces aware regexp

Let me know what you think so we eventually work on that together this week.

nikgraf:
I'm actually surprised about this issue, usually it should apply the decorator at all of the placed where the match happens. Let's do some pair programming. What about Friday afternoon?

nishp1:
We recently had the need for supporting whitespaces as well to match Facebook's behavior where they match upto 20 characters instead of matching by N words. Just submitted PR at https://github.com/draft-js-plugins/draft-js-plugins/pull/402.

I think with API now in a good state, we should consider adding `suggestionsStrategy` prop so these customizations can be made in user land.

nikgraf:
The PR https://github.com/draft-js-plugins/draft-js-plugins/pull/402 from @nishp1 is very well executed. As mentioned in the other PR it introduces more complexity and right now I'm have a hard time maintaining the existing stuff.

How do you feel about creating a 2nd mention plugin coving the other use case or allowing to customise the behaviour. hmmmm, starting to reconsider if @nishp1 might be worth merging as a last feature for mention …

@nishp1 @Zhouzi what are your thoughts?

nikgraf:
I closed this PR for now, but we can continue the conversation here. I will read it and have more time from now on to follow up.

luisrudge:
I think that not supporting white spaces is a bummer :( It may work OK when you only think about people, because names are usually specific enough. But when you want to mention a list of other stuff (in my case, financial terms in portuguese), I need the white-space in order to find something more relevant. Example:

14269453_1478449902181528_1219971627_n

Zhouzi:
@luisrudge I believe that @nikgraf is aware of the value this feature could bring to the mention plugin but he had to make a tough choice here. The problem is that it's introducing a level of complexity that would at best delay the release of v2 and at worst compromise it. We need to get back on schedule to have some more time to think about such feature.

Issue #253, opened 2016-05-04
Multiple editor mention suggestions do not work 


nadimtuhin:

nikgraf:
Hey @nadimtuhin, do have a little bit more details? Some code you can share?

nadimtuhin:
I will be in a few hours. Thanks for coming back :)

nikgraf:
ping 😄

nikgraf:
@nadimtuhin any updates here?

hervejegou:
I am trying to use two mention editors suggestions plugins  in the same Editor component: similar to Twitter or Slack , with '@' for mentioning users and with '#' for mentioning channels...

I tried to initialize two mention plugin components :

``` 
const mentionPlugin = createMentionPlugin({ mentionPrefix: '@' });
const { MentionSuggestions } = mentionPlugin;
const mentionPlugin2 = createMentionPlugin({ mentionPrefix: '#' });
const { MentionSuggestions2 } = mentionPlugin2;
const plugins = [mentionPlugin, mentionPlugin2];
```
Also setup onChange2() and suggestions2 state required for the second mention plugin component and initialize properly. 

Then list the two plugins in the render area:
```
<Editor
editorState={this.state.editor}
onChange={this.onChange}
plugins={plugins}
ref="editor"
/>
 <MentionSuggestions
onSearchChange={ this.onSearchChange }
suggestions={ this.state.suggestions }
/>
<MentionSuggestions2
onSearchChange={ this.onSearchChange2 }
suggestions={ this.state.suggestions2 }
/>
```
does not work... it was a long shot anyway... Any advice on whether this is possible??

nikgraf:
@hervejegou the mentionPrefix is just for presentation purposes and not used for matching.

This is the current documentation. Any suggestion on how to improve it? Maybe even create a PR?
```
mentionPrefixBy default it is an empty String. For Twitter or Slack like mention behaviour you can provide an `@`
```

yjang1031:
I've added a trigger option, by adding mentionTrigger to config, passing mentionTrigger to the mentionSuggestionsStrategy, having mentionSuggestionsStrategy return a function instead. (https://github.com/yjang1031/draft-js-plugins/commit/7c8de90cd1522915dce0f2ad102e0cd8c441b533)

[...]

PR #320, opened 2016-06-21
adding mentionTrigger 

yjang1031:
Addresses https://github.com/draft-js-plugins/draft-js-plugins/issues/253#issuecomment-227144837.

Adds `mentionTrigger` to `config` with default value of `@`.
Adds and searches for mention entities with key of `'mention' + mentionTrigger`
`mentionStrategy` and `mentionSuggestionsStrategy` takes the trigger as a parameter.

Feedback welcome

### DIFF
--- a/draft-js-mention-plugin/src/index.js
+++ b/draft-js-mention-plugin/src/index.js
@@ -13,8 +13,11 @@ import defaultPositionSuggestions from './utils/positionSuggestions';
 
 const createMentionPlugin = (config = {}) => {
   const defaultTheme = {
+    // CSS class for mention text
     mention: mentionStyles.mention,
+    // CSS class for suggestions component
     mentionSuggestions: mentionSuggestionsStyles.mentionSuggestions,
+    // CSS classes for an entry in the suggestions component
     mentionSuggestionsEntry: mentionSuggestionsEntryStyles.mentionSuggestionsEntry,
     mentionSuggestionsEntryFocused: mentionSuggestionsEntryStyles.mentionSuggestionsEntryFocused,
     mentionSuggestionsEntryText: mentionSuggestionsEntryStyles.mentionSuggestionsEntryText,
--- a/draft-js-mention-plugin/src/index.js
+++ b/draft-js-mention-plugin/src/index.js
@@ -14,9 +14,7 @@ import defaultPositionSuggestions from './utils/positionSuggestions';
 const createMentionPlugin = (config = {}) => {
   const defaultTheme = {
     mention: mentionStyles.mention,
-
     mentionSuggestions: mentionSuggestionsStyles.mentionSuggestions,
-
     mentionSuggestionsEntry: mentionSuggestionsEntryStyles.mentionSuggestionsEntry,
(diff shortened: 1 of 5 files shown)


draft-js-mention-plugin/src/index.js
```diff
@@ -14,9 +14,7 @@ import defaultPositionSuggestions from './utils/positionSuggestions';
 const createMentionPlugin = (config = {}) => {
   const defaultTheme = {
     mention: mentionStyles.mention,
-
     mentionSuggestions: mentionSuggestionsStyles.mentionSuggestions,
-
```
nikgraf: The spaces where here by intention to make it easy to distinct which component is affected. Maybe a comment would have been better. Can you add comments?

draft-js-mention-plugin/src/mentionStrategy.js
```diff
@@ -1,12 +1,12 @@
 import { Entity } from 'draft-js';
 
-const findMention = (character) => {
+const findMention = (trigger) => (character) => {
   const entityKey = character.getEntity();
-  return (entityKey !== null && Entity.get(entityKey).getType() === 'mention');
+  return (entityKey !== null && Entity.get(entityKey).getType() === `mention-${trigger}`);
```
nikgraf: This would be a breaking change to all existing implementations. I think by default the type should still be `mention`. I suggest we introduce another parameter to provide a custom `type`.

draft-js-mention-plugin/src/index.js
yjang1031: I didn't realize I made these changes. I'll add some comments.

draft-js-mention-plugin/src/mentionStrategy.js
yjang1031: I can't think of any use case in which a custom `type` is needed on its own.

Even if the `mentionTrigger` and `type` functionalities are separated, I believe it'd still be a good idea to make sure that the entity type doesn't conflict with other plugins or generic entity types, by prefixing the `type` string with some sort of mention plugin identifier string or by some other way?

nikgraf: If we would change the default type everyone who used the `mention` type and saved it to the database would now need to re-write all save data-structures in their database. That's why the default should stick to `mention`.

nikgraf: I will merge it and fix it, that in case of `@` we keep the default of `mention` type.

draft-js-mention-plugin/src/mentionSuggestionsStrategy.js
```diff
@@ -1,9 +1,8 @@
 /* @flow */
 
 import findWithRegex from 'find-with-regex';
+import _ from 'lodash';
```
nikgraf: Just noticed this as well. This problematic as it would include the whole lodash library. By using `import escapeRegExp  from 'lodash/escapeRegExp'` we can reduce the imported code and skip all other lodash code not necessary for this plugin.

nikgraf:
Some related improvements here https://github.com/draft-js-plugins/draft-js-plugins/pull/344

PR #326, opened 2016-06-24
Let mentions render with plain JS Object

anderslemke:
When you try to use an `editorState` that was converted from a
`RawDraftContentState`, the assumption that a mention is a `Map` does not
hold.

````js
const contentState = convertFromRaw(instanceOfRawDraftContentState);
const editorState = EditorState.createWithContent(contentState);
this.state = {
  editorState: editorState,
}
````

I assume that calling `fromJS` on something that is already a `Map` is a no-op.
Which seems to be the case: https://github.com/facebook/immutable-js/blob/38019326a8d006a45ef48d9c4644e23ea476d298/src/fromJS.js#L28-L36

### DIFF
--- a/draft-js-mention-plugin/src/Mention/__test__/index.js
+++ b/draft-js-mention-plugin/src/Mention/__test__/index.js
@@ -21,4 +21,11 @@ describe('Mention', () => {
     const result = render(<Mention entityKey={ entityKey } theme={ Map() } />);
     expect(result).to.have.tagName('span');
   });
+
+  it('can render when mention is an Object', () => {
+    const mention = {};
+    const entityKey = Entity.create('mention', 'SEGMENTED', { mention });
+    const result = render(<Mention entityKey={ entityKey } theme={ Map() } />);
+    expect(result).to.have.tagName('span');
+  });
 });
--- a/draft-js-mention-plugin/src/Mention/index.js
+++ b/draft-js-mention-plugin/src/Mention/index.js
@@ -1,9 +1,10 @@
 import React from 'react';
 import { Entity } from 'draft-js';
+import { fromJS } from 'immutable';
 
 const Mention = (props) => {
   const { entityKey, theme = {} } = props;
-  const { mention } = Entity.get(entityKey).getData();
+  const mention = fromJS(Entity.get(entityKey).getData().mention);
 
   if (mention.has('link')) {
     return (


nikgraf:
@anderslemke this a really interesting approach … I was thinking about switching to non-immutable structures for 2.0 version.

I would like to discuss what are the up & downsides of doing so compared to your version:

Pro of you version:

- Works with Immutable Maps & simple objects

Con of you version:

- Doesn't force people to be consistent

It kind of reminds me of the robustness principle: `Be conservative in what you do, be liberal in what you accept from others`. @Zhouzi @mjrussell what are your thoughts?

nikgraf:
I merged it to cut a last 1.x.x release

mjrussell:
It does seem a little weird to me to "force" people into using ImmutableJS in a library, although since DraftJS is built upon ImmutableJS and requires it already it seems a little less heavy-handed than if you had done it in another library. 

I would generally prefer plain JS allowing people to use custom creators for Immutable if they want it, but I don't typically use Immutable in my projects so I am probably a bit biased.

anderslemke:
Thank you for merging @nikgraf 

As for your thoughts about switching to plain JS object, I think that discussion could be split into two parts. What do you want to use internally in the plugins, and what do you support from the outside.

As for "internally", I do not feel qualified to answer that question.

Externally, I think the current way of supporting both plain and ImmutableJS externally is good way to go.

People doesn't control what comes out of `convertFromRaw`.
So, you might have put `Map`s into `Entity` in the first place, but when you `convertToRaw` and then `convertFromRaw` you will get plain JS objects from `Entity.get(entityKey).getData()`. So being consistent, is not really something that is left to the developer. 

That was my thoughts. Thanks for some great plugins. 👍

Zhouzi:
I agree with @mjrussell about the fact that a library shouldn't force you to change your workflow but is acceptable in regards to DraftJS as the core library does itself forces you to use ImmutableJS.

That been said, @anderslemke raised some pretty valid points regarding the difference between internal mechanics and external API usage. If possible, I'd suggest to make a choice between the two - the external API should always provide and accept the same kind of objects (I'd personally vote for plain JavaScript object, there's no risk of users mutating the objects anyway).

nikgraf:
Sounds good to me! Accepting both is ideal, but if not possible we switch to plain objects. I added this discussion to the roadmap issue.

ismyrnow:
@Zhouzi DraftJS does include Immutable, but doesn't force you to use it explicitly (i.e. you don't have to `import Immutable from 'immutable'`). I was able to build user mention functionality right on top of DraftJS without having to explicitly bring in Immutable. My preference would be for plain objects.

PR #327, opened 2016-06-24
Add a way to provide a custom component for the suggestions' entry

Zhouzi:
This PR adds a `entryComponent` option to `MentionSuggestions` that provides a way to have a custom template for each suggestions' entry. You can have a look at what it looks like from the documentation (I've updated the custom mention editor example to include an usage example).

screenshot


nikgraf:
This is nice. Let me cut a release and merge this for a 2.0 release.

nikgraf:
flawless PR! I like that … 🙌

PR #333, opened 2016-06-26
Make link clickable

LFra:
Is there a reason for keeping the mention editable?
With it enabled I couldn't click the link.

### DIFF
--- a/draft-js-mention-plugin/src/Mention/index.js
+++ b/draft-js-mention-plugin/src/Mention/index.js
@@ -9,6 +9,7 @@ const Mention = (props) => {
   if (mention.has('link')) {
     return (
       <a
+        contentEditable: false,
         href={mention.get('link')}
         className={theme.mention}
         spellCheck={false}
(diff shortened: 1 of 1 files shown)


draft-js-mention-plugin/src/Mention/index.js
```diff
@@ -9,6 +9,7 @@ const Mention = (props) => {
   if (mention.has('link')) {
     return (
       <a
+        contentEditable: false,
```
LFra: contentEditable = {false}

nikgraf:
Mentions can contain full names and if they do you can remove parts of it like you can do on Facebook e.g. `Nik Graf` -> `Nik` and it still is a mention. The idea is that in edit mode you don't follow links. In read mode the links work. Twitter, Slack have the same behaviour. Does that make sense?

Still I'm curious about your use-case. Why would you want that?

tageecc:
@nikgraf hi nikgraf, what if i‘m using the markdown plugin？

[https://github.com/withspectrum/draft-js-markdown-plugin](https://github.com/withspectrum/draft-js-markdown-plugin)

it is a WYSIWYG markdown plugin, i can't just set `readOnly=true` , couse i neet to edit it.

thx~

Issue #329, opened 2016-06-26
Roadmap


nikgraf:
2.0

In this issue we can discuss & elaborate on plans for a 2.0 release. Feel free to add your ideas here and let me know if you want to pick up one of those tasks.
## Discussion
- [x] A lot of the plugins are actually just decorators. Should we name them e.g. `draft-js-dnd-plugin-decorator` or  `draft-js-dnd-decorator-plugin`? The names get quite long, but on the other hand these modules are not plugins. Thoughts? 
- [x] Create & export a `compose` functions (like in Redux) to combine decorators. The alternative is one decorator where you can activate/deactivate certain plugins.
- [ ] should get getEditorRef get the real DOM node or just the DraftJS Editor?

## All Plugins
- [x] Remove all `<noscript />` and replace with `null
- [ ] refactor: make sure decorated blocks are never in blockRendererFn
- [ ] Add info to all changelogs that npm2 is deprecated and `peerDepedencies` moved to `depedencies`
- [ ] Investigate which dependencies are not needed
- [ ] Add `npm3` as requirement to package.json
- [ ] mobile support
- [ ] aria support
- [ ] check for all todos
- [ ] move all state within the createPlugin functions

```
  "engines": {
    "npm": ">=3"
  },
```
## Editor
- [x] Investigate & fix: https://github.com/draft-js-plugins/draft-js-plugins/issues/338
- [ ] check if setReadOnly is use anywhere and even necessary

## Inline Toolbar

- [x] add advanced example with custom set of buttons

## Side Toolbar

- [ ] show an example with image adding

## Mentions
- [x] Investigate if we should switch to plain JS objects instead of using `Immutable.Map` for the mentions list. This change would make it play nicer together with `convertFromRaw`. see https://github.com/draft-js-plugins/draft-js-plugins/pull/326
- [x] Revisit of `mentionTrigger` https://github.com/draft-js-plugins/draft-js-plugins/pull/320/files is a good idea or if we should simply recommend @mjrussell autocomplete plugin creator
- [ ] Split Themed Example into Themed Example only updating CSS and a custom components example updating the entryComponent on MentionSuggestions
- [x] deprecate mentionPrefix since this can now be done with `mentionComponent`
- [ ] update docs to use none-immutable mentions
- [x] fix this bug: https://github.com/draft-js-plugins/draft-js-plugins/issues/633

## Stickers
[...]

PR #337, opened 2016-06-28
Remove all <noscript> tags

rockingskier:
Step towards #329


nikgraf:
yay! thx @rockingskier

PR #339, opened 2016-06-29
Separate buttons from the toolbar plugin

bryanph:
This introduces 3 new plugins with existing functionality and makes draft-js-toolbar-plugin simply accept Button components that it renders.

- The _draft-js-inline-plugin_ exposes BoldButton, ItalicButton, UnderlineButton and MonospaceButton
- The _draft-js-block-plugin_ exposes H1Button, H2Button, ..., BlockquoteButton, UlButton, OlButton, CodeblockButton,
- The _draft-js-link-plugin_  Exposes a LinkButton
- The _draft-js-toolbar-plugin_  "actions" config property just accepts an array of components to render. Hence doesn't take care of the logic of the buttons but just the presentation

There are still a few bugs that need to be fixed:

- fix linting
- clean up some files in the plugins


nikgraf:
@bryanph so cool, can you fix all the linting issues? This help a great deal to have consistent codebase. You can run the command locally with `npm run lint`

nikgraf:
@bkniffler I mentioned this today, curious to hear your thoughts about this PR. Any suggestions how to improve it?

bkniffler:
Thanks for your contribution @bryanph! As far as I see, you'd like to outsource the text actions (link, blocktypes, inlinestyles) into their own plugins and let them return components / markup for each action, which gets inserted into the toolbar. Correct?

I think it might be a good move to outsource the default text and link action into their own plugins. But I'd also like to push the idea of having multiple toolbars that can be switched effortlessly (inline/air/medium-like-toolbar, fixed-toolbar, left/right-sided toolbar, there are quite some possibilities for toolbars). Therefore, we might want to declare unopinionated actions (as objects) instead of defining components (buttons). Keep in mind that blocks are able to define their own actions too, so having an intermediate way to define text actions (inline-style, block-type, links) and block actions (align left/right, upload image) is imperative to move forward.

For example (from the top of my head), this is how using the toolbar plugin in a complex scenario could look like:

```jsx
var toolbarPlugin = createToolbarPlugin();
// Inline + Aside (like on medium for example)
toolbarPlugin.addToolbar(createInlineToolbar(), InlineStyleActions.ALL, ToolbarPlugin.SEPARATOR, LinkAction, ToolbarPlugin.SEPARATOR, ToolbarPlugin.BLOCK_ACTIONS);
toolbarPlugin.addToolbar(createAsideToolbar(), BlockTypeActions.ALL);
// Classic (like your average wordpress wysiwyg editor)
toolbarPlugin.addToolbar(createFixedToolbar(), InlineStyleActions.ALL, BlockTypeActions.ALL);
```
Ergo: Add as many toolbars to the toolbar plugin as you want and define what actions each toolbar should render, allow for separators (`ToolbarPlugin.SEPARATOR`) and define which toolbar should show block-actions (`ToolbarPlugin.BLOCK_ACTIONS`, imagine the upload-image action that an image-block might define).

@nikgraf @bryanph What do you think?

bryanph:
@nikgraf @bkniffler I probably should have made an issue before just going ahead and implementing it (I couldn't resist :)).

@bkniffler what would the different plugins like link plugin provide in this case? An object describing the functionality like in the current implementation instead of a component to be passed to the toolbar plugin?

My though process was that in the way it's implemented above, the plugins are useful by themselves since they simply expose buttons which provide the desired functionality. Then it's up to the user where to place them. Implementing a fixed toolbar in this case would simply be a list or something containing the buttons. And the air/medium-toolbar would just be responsible for rendering in the right position. 

With regard to custom actions, one option would be to provide higher order components returning a button for custom actions. So for example:

```
const Editor = import 'draft-js-plugins-editor'
Editor.createActionButton({
    label: 'Link',
    button: <span>Link</span>,
    style: 'link',
    active: (block, editorState) => {
      ...
      },
    toggle: (block, action, editorState, setEditorState) => {
      ...
    },
  },
);
```
And then this button could be rendered anywhere, including the toolbar. Any further toolbar code could also then return a component which the user can render somewhere. Let me know what you think.

nikgraf:
# General plan

I believe we should focus on making multiple multiple toolbar plugins (air, aside & fixed) and not build one that solves multiple toolbar use-cases. Each of them have very different UI & UX patterns and building one plugin to rule them all would require a lot of effort & time (delaying a 2.0 release). We can merge multiple plugins at some point in the future if we see common patterns or extract common logic into separate libraries.

I propose to have:

- draft-js-air-toolbar-plugin (focus for 2.0 release)
- draft-js-side-toolbar-plugin (focus for 3.0 release?)
- draft-js-fixed-toolbar-plugin (something already exists https://github.com/jasonphillips/draft-js-richbuttons-plugin)

# draft-js-air-toolbar-plugin

## Configuration

I believe we should make sure the toolbar is flexible and people can choose their own elements.

A list of built in buttons: BoldButton, ItalicButton, UnderlineButton, MonospaceButton, ParagraphButton, Headline1Button, Headline2Button, Headline3Button, BlockquoteButton, UnorderedListButton, OrderedListButton, CodeblockButton, LinkButton

## Simples Toolbar Example

```js
const createAirToolbarPlugin = import 'draft-js-air-toolbar-plugin';

// by default contains [bold, italic, link, separator, h1, h2, separator, blockquote, ul, ol]
const toolbarPlugin = createToolbarPlugin();
```

# Custom Toolbar Example

```js
const createAirToolbarPlugin, {
  BoldButton,
  ItalicButton,
  ParagraphButton,
  Headline1Button,
  Headline2Button,
  BlockquoteButton,
  UnorderedListButton,
  Separator,
} = import 'draft-js-air-toolbar-plugin';

const config = {
  structure: [
    BoldButton,
    ItalicButton,
    Separator,
    ParagraphButton,
    Headline1Button,
    Headline2Button,
    Separator,
    BlockquoteButton,
    UnorderedListButton,
  ]
}
const toolbarPlugin = createToolbarPlugin(config);
```

# Discussion

1. Is this a good plan?
2. Do we need a higher order component to make it easier creating Buttons? (who wants to build it)
4. Should the config be extend so tooling like the alignment or table toolbar can leverage the toolbar?

I was thinking to something like this:

```js
const config = {
  modes: [
    {
      rule: isTextValidation,
      components: [
        BoldButton,
        ItalicButton,
        Separator,
        ParagraphButton,
        Headline1Button,
        Headline2Button,
        Separator,
        BlockquoteButton,
        UnorderedListButton,
      ],
    }, {
      rule: isImageValidation,
      components: [
        AlignLeftButton,
        AlignCenterButton,
        AlignRightButton,
      ],
    },
  ]
}
```

@bryanph @bkniffler what do you guys think?

bkniffler:
Yes @nikgraf, I agree with your propositions and I'm convinced this is the right direction, great write-up.

@nikgraf and @bryanph about HOCs for action to button mapping.
I'm not yet convinced yet that we will need such functionality. It should be very easy to do this wherever you need it with a custom mapping function, either as part of your toolbar plugin or as part of your custom editor.

```jsx
actions.map(action => (
   <button onClick={action.toggle} className={classNames({'active': action.toggled})}>
      {action.label}
   </button>
));

// or
actions.map(action => (
   <li className={classNames({'active': action.toggled})}>
      <a onClick={action.toggle}>{action.label}</a>
   </li>
));
```

@nikgraf, about the tooling:
I think we need to distinguish different action types. We got text and general actions (make bold, upload a file) that depend on selection/caret-position, and we got block specific actions (stuff that appears when specific block types are focused like images -> align-left/right/center).

Currently, the first scenario is not extendable, so we need to implement this through the return value of plugins, by some kind of 'actions' key that has an array of actions with conditions (selection.length > 0 or selection.length === 0 or whatever). 
The block actions scenario is made possible by extending the this.props.actions array through every HOC, like the alignment wrapper. The final component, like the image component, will then hand the actions to the toolbar. Thats how it currently works. We could also try and make it possible through the plugin config like in scenario one, but I think HOCs should define the actions themselves, so they can make use of the toggle function directly. The alternative would be to write the actions (align left/right/center) in the plugin indexjs and use entityState communicate changes to HOCs.

bryanph:
@nikgraf, I agree with your plan. Seems like a straightforward API.
Especially the 2nd example seems very extensible.

Also, I agree with @bkniffler that a HOC for buttons is probably overkill.

For clarity, in your example; are the components array elements actual components or are they action descriptions (objects)?

nikgraf:
Closing this as I already rewrote the toolbar and merged it.

PR #348, opened 2016-07-10
update beta plugin usage instructions.

mkampik:
`$ npm install <plugin>@beta --save` also throws a `no version match` error.

Should have assumed as much but didn't get round to actually installing any plugins until now, sorry!


nikgraf:
thx 🙌

PR #350, opened 2016-07-11
pass search value to Entry component for user land customizations

nishp1:
Enables for customizations in user land:

screen shot 2016-07-11 at 2 38 42 pm


nikgraf:
@nishp1 this pretty cool! I reverted back the example in this PR https://github.com/draft-js-plugins/draft-js-plugins/pull/366, because I believe using dangerouslySetInnerHTML is not a good option here. In addition the custom themed example is becoming quite complex.

@nishp1 would you mind creating a 4th example at the bottom focusing just on this? Ideally we split the strings and don't use `dangerouslySetInnerHTML`.

PR #354, opened 2016-07-13
add onSelect mention

batmandarkside:
```javascript
const mentionPlugin = createMentionPlugin({
      onSelect: (mention) => {
        console.log(mention, 'onSelect');
      }
    });
```


nikgraf:
@batmandarkside can you explain me what's your use-case and what you try to achieve? I'm happy to dive and I'm confident we can find a way to make it happen without extending the API here.

PR #355, opened 2016-07-14
Remove  to prevent react-unknown-props warning

aunsuwijak:
@nikgraf PR as we discussed in slack

### DIFF
--- a/draft-js-linkify-plugin/src/Link/index.js
+++ b/draft-js-linkify-plugin/src/Link/index.js
@@ -15,7 +15,6 @@ export default class Link extends Component {
       theme = {},
       target = '_self',
       className,
-      ...props,
       } = this.props;
     /* eslint-enable */
     const combinedClassName = unionClassNames(theme.link, className);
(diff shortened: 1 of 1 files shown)


mjrussell:
@nikgraf I think mention has something similar which will warn in 15.2

nikgraf:
@aunsuwijak instead of removing the props completely I propose to take out all the custom ones and then pass the `elementProps` down to the Anchor element. See how it's done here: https://github.com/draft-js-plugins/draft-js-plugins/pull/367/files

aunsuwijak:
@nikgraf That solution looks better, I'll update this one soon.

aunsuwijak:
@nikgraf After discussed with some other guy, I still not sure do we really need to take out unused props into variables.

nikgraf:
@aunsuwijak any particular suggestion? I was thinking about using lodash's omit, but then that would need we need lodash as another dependency.

nikgraf:
I'm closing this one as we solved all the react-unkown-props warnings.

PR #358, opened 2016-07-14
Adds a check to set mouseDown property after component updates

AquiGorka:
I could not get the plugin to work properly when clicking on an item. After reviewing the code I found the mouseDown property was always set to false even though my "mouse was down". With this check, clicking on an item works as expected.

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/Entry/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/Entry/index.js
@@ -15,7 +15,7 @@ export default class Entry extends Component {
   }
 
   componentDidUpdate() {
-    this.mouseDown = this.mouseDown || false;
+    this.mouseDown = this.mouseDown || false;
   }
 
   onMouseUp = () => {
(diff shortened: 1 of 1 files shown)


nikgraf:
@AquiGorka what browser did you use? You change makes sense, but for me it even works without your change. I would like to investigate a bit further first.

AquiGorka:
@nikgraf I am using chrome. I actually fixed the problem on my side but I still believe this PR should be merged.

My problem: There was something triggering a render while my mouse button was down and that was resetting your component to mousedown = false. I figured what was causing the re-render and put an end to that.
Having said that, I still believe that such a re-render shouldn't reset your component's state to mousedown = false, I can't really think of a scenario where this behaviour should be expected right now though.

Obviously it is up to you, maybe right now this PR does not really add value but maybe down the road it will be needed.

Cheers

nikgraf:
@AquiGorka right now I wonder why we even added the 3 lines. Unfortunately I didn't add a comment. I can't think of a case and currently think of removing these 3 lines completely. I do some further testing later this weekend.

PR #370, opened 2016-07-25
Add propTypes for `mentions` list of MentionSuggestions/index.js. 

mzbac:
also added fixing unknown props of draft-js-mention-plugin/src/MentionSuggestions/Entry/defaultEntryComponent.js


nikgraf:
thx, this is nice! ✨

PR #379, opened 2016-08-03
Fix module not resolving for lodash escapeRegExp, causing linting error

adrianmc:
It seems that the lodash library is no longer organized such that you can import like this:

```js
import escapeRegExp from 'lodash/escapeRegExp';
```

Many of its functions are now separated into sub-folders for ease of categorization, so now you must do this instead:

```js
import escapeRegExp from 'lodash/string/escapeRegExp';
```


adrianmc:
I've made a mistake

Issue #383, opened 2016-08-06
The sticker plugin doesn't work on the demo website


Zhouzi:
I am unable to add a sticker here: https://www.draft-js-plugins.com/plugin/sticker

Clicking on one doesn't trigger anything apart form this error in the console:

```
app.js:3 Uncaught Error: Got unexpected null or undefinedn
  @ app.js:3value
  @ app.js:43_renderValidatedComponentWithoutOwnerOrContext
  @ app.js:38_renderValidatedComponent
  @ app.js:38_updateRenderedComponent
  @ app.js:38_performComponentUpdate
  @ app.js:38updateComponent
  @ app.js:38receiveComponent
  @ app.js:38receiveComponent
  @ app.js:6updateChildren
  @ app.js:38_reconcilerUpdateChildren
  @ app.js:39_updateChildren
  @ app.js:39updateChildren
  @ app.js:39_updateDOMChildren
  @ app.js:39updateComponent
  @ app.js:39receiveComponent
  @ app.js:38receiveComponent
  @ app.js:6updateChildren
  @ app.js:38_reconcilerUpdateChildren
  @ app.js:39_updateChildren
  @ app.js:39updateChildren
  @ app.js:39_updateDOMChildren
  @ app.js:39updateComponent
  @ app.js:39receiveComponent
  ...
```

mzbac:
that caused by blockReanderMap, because it  doesn't have 'sticker' blockTpye to map to element.
to solve this problem I think it needs to modify plugin-edit to support pass down blockReanderMap from plugins to draft-js

temporary fix can just add the below attribute in /docs/client/components/pages/Sticker/SimpleStickerEditor/index 's editor element
 blockRenderMap={DefaultDraftBlockRenderMap.merge(Map({'sticker':{ element: 'div'}}))}

mzbac:
I close this issue since it fixed by redeploy correct draftjs plugin version

PR #385, opened 2016-08-09
temporary fix for sticker

mzbac:
custom block should have mapping reference in blockRenderMap
refer to https://github.com/draft-js-plugins/draft-js-plugins/issues/383


nikgraf:
@mzbac this looks great!

One issue I'm thinking about is: with this update you have no chance to completely overwrite the blockRenderMap which is required by other plugins like https://github.com/icelab/draft-js-single-line-plugin.

Right now I'm thinking of adding a option to the editor: `defaultBlockRenderMap` which is a boolean an by default set to true and if manually set to false it won't be used.

What do you think about that? Also wanna implement it?

To finish the PR you or someone should:
- add tests for the editor for blockRenderMap
- update the Changelog of the editor
- update the Changelog of the sticker-plugin

nikgraf:
@mzbac thanks a lot!

PR #393, opened 2016-08-18
Correct typos in mention plugin FAQ

rickhanlonii:
## Overview

This PR corrects a few typos in the mention plugin FAQ.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -23,7 +23,7 @@ this.setState({ editorState: EditorState.createEmpty() })
 
 ## Why are mentions broken after using `convertFromRaw` and throwing an error?
 
-We design the API to accept an Immutable Map for a Mention. After saving your datastructure to the erver and using `convertFromRaw` the mentions in there are plain objects. I (Nik) believe this is a law in our design and the mention data should be a plain object. Hint: We might fix this with v2.0.0 f the mentions plugin.
+We design the API to accept an Immutable Map for a Mention. After saving your data structure to the server and using `convertFromRaw`, the mentions in there are plain objects. I (Nik) believe this is a law in our design and the mention data should be a plain object. Hint: we might fix this with v2.0.0 of the mentions plugin.
 
 What you can do now is fixing the datastructure before converting it:
 


FAQ.md
```diff
@@ -23,7 +23,7 @@ this.setState({ editorState: EditorState.createEmpty() })
 
 ## Why are mentions broken after using `convertFromRaw` and throwing an error?
 
-We design the API to accept an Immutable Map for a Mention. After saving your datastructure to the erver and using `convertFromRaw` the mentions in there are plain objects. I (Nik) believe this is a law in our design and the mention data should be a plain object. Hint: We might fix this with v2.0.0 f the mentions plugin.
+We design the API to accept an Immutable Map for a Mention. After saving your data structure to the server and using `convertFromRaw`, the mentions in there are plain objects. I (Nik) believe this is a law in our design and the mention data should be a plain object. Hint: we might fix this with v2.0.0 of the mentions plugin.
```
rickhanlonii: @nikgraf IMO using Immutable.js isn't the flaw here, but not accepting a plain JavaScript object is. Could this by corrected by converting the input to Immutable if it's a plain object. If so, then you're not changing the input contract so won't need a major version update.

Zhouzi: It's a discussion we started some time ago and I can't remember if we made a choice. There was three possibilities:

1. Allow JavaScript objects as well as Immutable objects
2. Allow only JavaScript objects
3. Allow only Immutable objects

I am wondering if allowing JavaScript objects wouldn't create the false assumption that the editor deals with JavaScript objects under the hood. The thing is that DraftJS is pretty tightly coupled to Immutable and it's great to keep that in mind or you'll likely come across issues that wouldn't make sense otherwise.

rickhanlonii: Thanks for the background 🍻 

> I am wondering if allowing JavaScript objects wouldn't create the false assumption that the editor deals with JavaScript objects under the hood

If it helps, I'd say that callers shouldn't worry about what goes on under the hood as long as everything is properly encapsulated and the contracts are clear. I think it's fair to tell callers that you can pass in either a plain object or immutable object, but if a mention is passed to a callback then it will be an immutable object.
  
> The thing is that DraftJS is pretty tightly coupled to Immutable  

 True, and DraftJS works when using `convertFromRaw` so I imagine they're already doing the conversion.

Zhouzi: Good points, thanks for sharing 🍻 

>  callers shouldn't worry about what goes on under the hood

Agreed but right now there are other places where you do need to use Immutable objects, e.g the suggestions. Maybe we should change those too.

> DraftJS works when using convertFromRaw so I imagine they're already doing the conversion

If DraftJS does it then I think it's safe to follow their example. All in all I'd personally be ok with allowing JavaScript and Immutable objects for the public api.

nikgraf: great conversion! very much what I was thinking about as well.

In general we could allow both, but I wonder if it creates more confusion than it actually help. I like it when there is only one way of doing things. Then docs and examples out in the wild are consistent. Initially I wanted to go with Immutable all the way, but due the nature of `convertFromRaw` plain JS objects make more sense.

Right now I'm thinking: Let's do both for now, convert everything to plain object examples/docs and eventually drop Immtuable (if not other API changes happened until then in Immutable or `convertFromRaw`.

nikgraf:
thx for you help @rickhanlonii ✨

FAQ.md
rickhanlonii: Makes sense. Happy to help!

PR #397, opened 2016-08-19
Decorators are not applied when props are received.

goodhyun:
Needed to deal with `componentWillReceiveProps`.
You sometimes lost decorators if your Editor is updated via props.


nikgraf:
hey @goodhyun - I only happens if you apply a new EditorState instance. I thought a lot about doing this, but then on the other hand it would break with the Draft.js default behaviour.

We added a guide on how to properly reset the EditorState here: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/FAQ.md#how-to-reset-the-content

Does that help? I will close this PR for now, but let me know if I missed something and you have good reasons why this still should go in. Happy to discuss and reopen again if needed.

goodhyun:
Sounds good, even though this is introducing a little bit of quirks for the library users. Thanks!

PR #402, opened 2016-08-25
add support for whitespace

nishp1:
Closes https://github.com/draft-js-plugins/draft-js-plugins/issues/316


nikgraf:
hey @nishp1, I checked out your code and it looks good! I really appreciate you contributing here, but currently believe this should be not in the scope of this plugin. I can understand the desire to have white-space support, but then limiting the amount of characters breaks the existing behaviour and is not how I or others use this plugin.

Let me propose an alternative suggestion: take this version of the plugin, create a new repo & npm package. I'm happy to share it on Twitter, link it in the Readme and even add link to the docs.

What do you think?

luisrudge:
@nishp1 can you publish this as a new plugin?

mathieumg:
@nikgraf I understand your position, but I just want to express my opinion that I don't see a mention plugin being very relevant if it doesn't support inputting at least a first + last name combination in it, which is probably the most common use case. That being said, thanks for the great work!

PR #405, opened 2016-08-28
Update emojione from v2.1.2 to v2.2.6

markusjwetzel:
I have just copied all the lists from the emojione repository, so these lists should be correct.


nikgraf:
awesome, thx @markusjwetzel 👍

nikgraf:
@markusjwetzel unfortunately I noticed a couple of the emojis were broken. It seemed to be related to the escape map, but didn't have time to investigate further. I feel like ideally we could contribute to the original project make sure it exposes the utility functions we need.

Are you interested to help out here?

mzbac:
@nikgraf  can you give us some details of which emojis were broken? then I can try to find out what's the cause..

nikgraf:
@mzbac this line returned undefined in a lot of cases: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-emoji-plugin/src/Emoji/index.js#L7 for example for `:heart:`

mzbac:
@nikgraf got it, by way as you said want to use original project. do you mean we can add emojione as dependency in draftjs-plugin-emoji?

markusjwetzel:
Since mzbac is working on this issue, I don't think my help is needed at the moment. Thank you, @mzbac! :)

The only reason I see for not using the emojione package as dependency is that we only need a subset of /lib/js/emojione.js from the original emojione project (unfortunately everything is in this one file). So we could reduce the size of this project by implementing the emojione lists manually.

nikgraf:
@mzbac yep, we could add emojione as depedency , but last time I checked it didn't support everything we needed …

@markusjwetzel that's true, still upgrading will be so much hassle. Ideally he splits up the library and we leverage sub-parts.

mzbac:
@nikgraf could you please have a look at https://github.com/draft-js-plugins/draft-js-plugins/pull/436 ?
I have test locally, seems everything is working.. no sure what missed on you last time checking of emojione..

nikgraf:
just taking a closer look 👍

PR #414, opened 2016-09-02
move to single lodash.escaperegexp package for mention

nikgraf:
@Zhouzi any downsides doing this? feels more lightweight now using the whole lodash chain for one function :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
     "immutable": "^3.8.1",
     "linkify-it": "2.0.0",
     "loader-utils": "^0.2.15",
+    "lodash.escaperegexp": "^4.1.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -102,6 +102,7 @@
     "immutable": "^3.8.1",
     "linkify-it": "2.0.0",
     "loader-utils": "^0.2.15",
+    "lodash.escaperegexp": "^4.1.2",
```
Zhouzi: Not sure if that's needed in the main package.json?

Zhouzi:
Looks like the best trade-off to me, as long as we don't start using other lodash stuff (and even then, it'd be a per-case concern).

PR #415, opened 2016-09-07
feat(Linkify): Allow the Linkify plugin to accept a custom component

antoinerey:
This adds a configuration key 'component', which can be used to enhance the default one
with additional properties like event handlers, or inline style.

In other words, this is the #271 for the Linkify plugin.

### DIFF
--- a/draft-js-linkify-plugin/src/Link/index.js
+++ b/draft-js-linkify-plugin/src/Link/index.js
@@ -9,25 +9,28 @@ linkify.tlds(tlds);
 // The component we render when we encounter a hyperlink in the text
 export default class Link extends Component {
   render() {
-    /* eslint-disable no-use-before-define */
     const {
       decoratedText = '',
       theme = {},
       target = '_self',
       className,
-      ...props,
+      component,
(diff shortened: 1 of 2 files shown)


nikgraf:
This is great, thx @antoinerey ✨ 👌

If I may ask for 2 things for future PRs: please update the changelog and ideally write some test. Without them it makes it harder to review and slows down development of the whole project.

draft-js-linkify-plugin/src/Link/index.js
```diff
@@ -9,25 +9,28 @@ linkify.tlds(tlds);
...
-        className={combinedClassName}
-        target={target}
-      />
-    );
+
+    const props = {
+      ...rest,
+      href,
+      target,
+      className: combinedClassName,
```
nikgraf: I like the approach where you still provide the combined className

antoinerey:
Oh yes, sorry about those. I'll do better in the future. :)

nikgraf:
all good @antoinerey, thx for your work

Issue #388, opened 2016-08-12
Mentions: Uncaught TypeError: clientRectFunctions.get(...) is not a function


jameslk:
When I try to use a mentionTrigger longer than 1 character, I start to get this error:

```
index.js:106 Uncaught TypeError: clientRectFunctions.get(...) is not a function
getPortalClientRect @ index.js:106
(anonymous function) @ index.js:77
MentionSuggestions._this.componentDidUpdate @ index.js:89
invokeComponentDidUpdateWithTimer @ ReactCompositeComponent.js:66
...
```

Steps to reproduce:

Set up a mentionTrigger of _test_,

``` javascript
    const mentionsPlugin = createMentionsPlugin({
        mentionTrigger: 'test'
    });
```

Then in the editor, type "test" and press backspace, which triggers the error.

This appears to be caused by the code in `MentionSuggestions` `componentDidUpdate`:

``` javascript
const decoratorRect = this.props.store.getPortalClientRect(this.activeOffsetKey);
```

It fails because `this.activeOffsetKey` is no longer in the store `searches` map. It was taken out from a prior call to the store's `unregister` function when I hit backspace.

gfx:
I have the same errors on my application.

Fortunately I can reproduce this issue on the demo site: https://www.draft-js-plugins.com/plugin/mention

1. Start a mention selection by typing `@`
2. Select the first one
3. Delete it by typing three backspaces
4. Type `@foo` and `<enter>`

Then you will see this error.

nikgraf:
Thanks for reporting @jameslk & @gfx. I'm quite busy right now. Does one of you have time to investigate & fix the issue? A Pull Request would be very welcome.

mzbac:
just quickly went through source code, it seems mention plugin only allow to  set single character to be mentionTrigger. no sure if needs a RP to modify it to support mulit-characters mentionTrigger.

paulyoung:
@mzbac I did something related to that in https://github.com/mjrussell/draft-js-autocomplete-plugin-creator/pull/7.

IMHO, it would be great if that project could be brought into this repo and the mentions plugin be implemented in terms of that.

cc @mjrussell

paulyoung:
@mzbac I also just proposed these changes: https://github.com/mjrussell/draft-js-autocomplete-plugin-creator/pull/11

mzbac:
[...]

PR #416, opened 2016-09-07
add checking logic for getPortalClientRect and onMentionSelect.

mzbac:
fix issues from https://github.com/draft-js-plugins/draft-js-plugins/issues/388

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/index.js
@@ -58,7 +58,9 @@ export default class MentionSuggestions extends Component {
           focusedOptionIndex: size - 1,
         });
       }
-
+      if(!this.props.store.getAllSearches().has(this.activeOffsetKey)) {
+        return;
+      }
       const decoratorRect = this.props.store.getPortalClientRect(this.activeOffsetKey);
       const newStyles = this.props.positionSuggestions({
         decoratorRect,
--- a/draft-js-mention-plugin/src/MentionSuggestions/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/index.js
@@ -206,6 +206,9 @@ export default class MentionSuggestions extends Component {
   };
 
   onMentionSelect = (mention) => {
+    if (!mention) {
+      return;
+    }
     this.closeDropdown();
     const newEditorState = addMention(
       this.props.store.getEditorState(),
--- a/draft-js-mention-plugin/src/MentionSuggestions/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/index.js
@@ -58,7 +58,7 @@ export default class MentionSuggestions extends Component {
           focusedOptionIndex: size - 1,
         });
       }
-      if(!this.props.store.getAllSearches().has(this.activeOffsetKey)) {
+      if (!this.props.store.getAllSearches().has(this.activeOffsetKey)) {
         return;
       }
       const decoratorRect = this.props.store.getPortalClientRect(this.activeOffsetKey);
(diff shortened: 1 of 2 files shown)


draft-js-mention-plugin/src/MentionSuggestions/index.js
```diff
@@ -206,6 +208,9 @@ export default class MentionSuggestions extends Component {
   };
 
   onMentionSelect = (mention) => {
+    if (!mention) {
```
nikgraf: In which situation can mention be `null` or `undefined`?

draft-js-mention-plugin/src/MentionSuggestions/index.js
```diff
@@ -58,7 +58,9 @@ export default class MentionSuggestions extends Component {
           focusedOptionIndex: size - 1,
         });
       }
-
+      if (!this.props.store.getAllSearches().has(this.activeOffsetKey)) {
```
nikgraf: I'm bit out of the this code. Can you elaborate why this is solving the problem? At best add a comment so we know in the future when revisit it.

mzbac: it doesn't solve the multi-character mention trigger problem.  it just a simple protection for the error when componentDidUpdate try to get new getPortalClientRect but key already deleted by previous action. ( right now, it only can happened when set the mention trigger to be multi-characters. current mention plugin logic is kind of handle mention trigger as single char. for example:
`onSearchChange = (editorState, selection) => {
    const { word } = getSearchText(editorState, selection);
    const searchValue = word.substring(1, word.length);
    if (this.lastSearchValue !== searchValue) {
      this.lastSearchValue = searchValue;
      this.props.onSearchChange({ value: searchValue });
    }
  };`
`const selectionIsInsideWord = leaves
      .filter((leave) => leave !== undefined)
      .map(({ start, end }) => (
        start === 0 && anchorOffset === 1 && anchorOffset <= end || // @ is the first character
        anchorOffset > start + 1 && anchorOffset <= end // @ is in the text or at the end
      ))`

no sure if we need add support for mention trigger of multiple character

draft-js-mention-plugin/src/MentionSuggestions/index.js
mzbac: user didn't select mention from mentionsuggestions, instead they can just by typing @xxx (invalidate mentions) then hit enter .. in that case mention will be undefined..

nikgraf: ohhh, good point

draft-js-mention-plugin/src/MentionSuggestions/index.js
nikgraf: ok, so then the plugin doesn't throw an error, but won't work anyway. is this correct?

mzbac: yes, it won't work for mention trigger more than 1 char. but at least the code doesn't break now..I think we may either add docs to tell user that mention trigger only can be set as single character or     add support for mulit-character mention trigger.

nikgraf:
@mzbac I updated the docs already in another PR

evankline:
@nikgraf @mzbac I believe a similar error is occurring when the MentionSuggestions are unfocused via a click or escape and the trigger character is still in the input. From what I observed it seems to be that the MentionSuggestionsPortal's searchPortal ref is no longer populated. I also noticed that if the same steps are taken, the same error occurs if a ref has been added to the Editor component as in the example below. On what may be a separate issue, or possibly a related issue, when the input is re-focused with the trigger character to the left of the cursor, would it be expected behavior to have the MentionSuggestions component opened as though the user had just typed the trigger character?

```
<Editor 
  ref={(el) => this.editor = el} 
  plugins={plugins}
  onChange={this.onChange}
/>

draft-js-mention-plugin/src/MentionSuggestions/index.js
nikgraf: I did that :)

PR #417, opened 2016-09-07
Fix Build for Mention Suggestions

brettjurgens:
I noticed that using the Mentions plugin from npm, the build version disregards the `entryComponent` prop. Seems to be related to `defaultProps` & the webpack build.

in this PR i removed the `defaultProps` and set entryComponent to be `props.entryComponent || defaultEntryComponent`

thanks


nikgraf:
interesting

PR #422, opened 2016-09-08
Fix incorrect PropType specification in CustomCounter

smeijer:
`PropTypes.function` is incorrect. It should be `PropTypes.func`.

Also a `CustomCounter` without a `count` function doesn't really make sense. So I've marked the `countFunction` as being required.


nikgraf:
good catch, thx for the fix 👍

PR #435, opened 2016-09-12
Mention plugin to aceept '-', '.', '_' while typing

jssisodiyaPG:


nikgraf:
nice one, can you also update the changelog of the mention plugin?

jssisodiyaPG:
@nikgraf Done 👍

jssisodiyaPG:
@nikgraf thanks for merging the PR. Let me know when you publish this to npm. I will update my package

nikgraf:
@jssisodiyaPG I just tested it an it breaks in Chrome. `.` must be escaped as well. Can you create a second PR with a fix?

mzbac:
@jssisodiyaPG I believe correct regex should be '\._\\-'

PR #436, opened 2016-09-12
replace hard code emojiList and escapeMap by emojione

mzbac:
introduce emnojione dependency and replace exist emojiList by emnojione's emojiList
https://github.com/draft-js-plugins/draft-js-plugins/pull/405
@nikgraf  could you help me to solve the package dependency issue? I have add emojione as dependency in draft-js-emoji-plugin, but still got Unable to resolve path to module 'emojione'


nikgraf:
works like a charm! all emojis I had trouble with are working fine

There is one issue that is worth fixing. In the existing version we re-ordered the list to make sure the first results when you just type `:` is an interesting mix of more often used emojis.

screen shot 2016-09-14 at 23 15 41


After this PR it would not be the case anymore:
screen shot 2016-09-14 at 23 15 22

mzbac:
@nikgraf  I added a priorityList which sit on top of rest of emojis in emojiList.. considering to expose that as part of config, then user can make their own choice to decide which emojis should show by default

nikgraf:
@mzbac great! yeah, let's expose the priority list. The idea is fantastic. I didn't even think about.

Did you see my comment regarding the packages.json?

mzbac:
@nikgraf  it's all done..

PR #440, opened 2016-09-14
Feature/video plugin

mzbac:
add video plugin, by default user can copy youtube or vimeo url  and paste on editor generating iframe video player


draft-js-video-plugin/package.json
```diff
@@ -0,0 +1,100 @@
...
+    "sinon": "^1.17.5",
+    "sinon-chai": "^2.8.0",
+    "style-loader": "^0.13.1",
+    "superagent": "^2.2.0",
+    "webpack": "^1.13.2",
+    "webpack-dev-middleware": "^1.7.0",
+    "webpack-hot-middleware": "=2.12.2"
+  },
+  "config": {
+    "commitizen": {
```
nikgraf: what's that? should we remove it or does it serve a purpose?

nikgraf:
@mzbac can you add page to the docs where I can try the plugin? If you are unfamiliar with the docs I'm happy to do it. I can create PR on top of this one to demonstrate how it works.

draft-js-video-plugin/package.json
mzbac: @nikgraf I am suing commitizen to commit my changes initially.. no sure if you want to use it in draftjs plugin project..

mzbac:
@nikgraf sure， I will add demo page in docs...

draft-js-video-plugin/package.json
nikgraf: as you like, if it's a tool useful to you I don't have any objections adding it …

nikgraf:
@mzbac ping me if you need help with the docs. we can also do a remote pair programming session

mzbac:
@nikgraf， I was quite busy over last weekend, will keep working on this tonight. Sure we can do pair programming session. Let's discuss it on slack..

mzbac:
@nikgraf I think I have done this.. can you review and merge it?

nikgraf:
sorry for the late reply, ok this is super close to be merged from my perspective 👍 

a few more nitpicks:

- Can you make sure either the videos keep the aspect ratio of youtube instead of the fixed height? Looks weird on widescreen monitors.

kapture 2016-10-09 at 17 01 51

- If you add a youtube video in the advanced example it creates a video tag with youtube as source (which doesn't work).

What do you think about avoiding the automatic parsing at all and using a button like the sticker plugin to add videos? I can imagine this automatic behaviour could be very confusing in case you paste in a block. I believe Slack & Facebook never did this and only added one block at the end as it would be to confusing if it breaks the text from a UX perspective.

mzbac:
@nikgraf didn't quite understand what you mean, If you add a youtube video in the advanced example it creates a video tag with youtube as source (which doesn't work). and problem with use a button to insert a video we have to provide some kind of user interface to do that..no sure if the video plugin should be responsible to that..I think maybe we can introduce a parameter in config object to disable automatic parsing and export addVideo https://github.com/draft-js-plugins/draft-js-plugins/blob/feature/video-plugin/draft-js-video-plugin/src/video/modifiers/addVideo.js#L1 to them let end user to decide how to insert new video block..

nikgraf:
@mzbac 

1. If you add a Youtube link in the second example on the docs it creates a video tag but uses the Youtube link as src. In the end you see a big white box.

2. Totally agree with exporting the the modifier to add a video like with the [Sticker plugin](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-sticker-plugin/src/index.js#L60). I would even go as far to export a button as well since we also have StickerSelect. But I guess at one point we could remove the button if there is a good sidebar plugin that makes it easy to just pass a modifier.

I'm pretty sure auto-converting video links is bad UX. I experimented with it years ago and it turned out to be super confusing to people nor have I seen it in any product. Happy to be convinced if you have a good example :)

nikgraf:
btw I really looking forward to get this merged!

mzbac:
@nikgraf for first issue, the wrapperComponent suppose let user decide how to present the video by their desired video player..therefore they should make their own logic to leverage video source which returned from getVideoSrc function..

and I will follow the StickerSelect see if I can make the button to insert video..

mzbac:
@nikgraf I made some changes, 1, add  autoHandlePastedText flag in config which default to false, it controls auto parsing. 2. plugin provides applyAddVideoComponent which is HOC component that expose addVideo function to end user therefore they can wire up their own login to insert new video block.

more details please refer the doc's video page.

nikgraf:
wahhh @mzbac I still wanna merge this. Sorry, wasn't on my focus list for now. Let's get back to it after 2.0?

mzbac:
@nikgraf  I closed this PR because I need to rebase the master branch..we cab do the after 2.0.  no worries..

PR #443, opened 2016-09-15
Fixes race condition with Japanese input

stevensacks:
When inputting Japanese characters (or any complex alphabet which requires hitting enter to commit the characters), that action was causing a race condition. By changing componentWillMount to componentDidMount, this issue is resolved and the component will unmount unregister and then properly mount and register after. Prior to this change, componentWillMount would not fire after componentWillUnmount even though it was still in the DOM, so it wasn't re-registering the offsetkey.


mzbac:
@nikgraf  can you review this change?

nikgraf:
@mzbac done!

PR #449, opened 2016-09-18
Update emoji

nikgraf:
@mzbac I made some changes on Emoji plugin: 

- removed the unicodeRegex & escape map (used the one from emojiOne)
- added the priorityList param to the docs
- cleaned up draft-js-emoji-plugin/package.json to align with all other plugins 
- removed double entry of emojione in package.json
- renamed `ns` to `emojione` (I assume they mean `namespace` for `ns` in the library, but that's not a good fit in here)
- renamed emojione to emojiList

Can you double-check the changes? Did I miss something? Potential for a bug?


mzbac:
@nikgraf, thanks for helping on those cleaning up...

PR #454, opened 2016-09-23
Update to support 16:9 ratio.

rivergod:
Update to support 16:9 ratio. (width: parrent width size)


nikgraf:
@mzbac how do you feel about this change?

mzbac:
@nikgraf looks good. I was waiting @rivergod to fix the lint error.. if he don't mind I can fix the lint errors and merge it

PR #461, opened 2016-09-27
Add babel polyfill to support older browser

### LINKED ISSUE #459: draft-js-plugins.com website down in IE11
If you go to https://www.draft-js-plugins.com/ in Internet Explorer 11, it simply says 
"Page not Found :("

An error that's thrown is "Object doesn't support property or method 'startsWith'"

What's interesting is that if you go to the website on Google Chrome (mac), that error doesn't appear, but the following error does: "Cannot read property 'replaceChild' of null"

Grsmto:
I added babel polyfill to the demo/website so it supports IE10:
screen shot 2016-09-27 at 10 53 16

Fix #459


nikgraf:
nice, thx 👍

Issue #457, opened 2016-09-23
Mentions: not work with Cyrillic


kortv:
onSearchChange event does not occur when typing Cyrillic character after @

ZozuliaDV:
in defaultRegExp.js add \u0400-\u0500
`exports.default = '[\\w\u4E00-\u9EFF\u3040-\u309F\u30A0-\u30FF\uAC00-\uD7A3\u3130-\u318F\u0400-\u0500]*';`

nikgraf:
@kortv we just merged a PR and going to release a new version today 😄

Issue #406, opened 2016-08-28
Mention plugin to handle some special characters like _ - .


jssisodiyaPG:
Usernames might have some special characters. Mention plugin regex should handle those as well.
Somthing like this - `/(\s|^)@[\w_\-.]*/g`

nikgraf:
Makes sense. Can you create a PR?

jssisodiyaPG:
Sure, here is the PR
https://github.com/draft-js-plugins/draft-js-plugins/pull/435

PR #463, opened 2016-09-27
add mentionUnicodeRegex support in config for mention plugin

mzbac:
it solve the issue for 
https://github.com/draft-js-plugins/draft-js-plugins/issues/406
https://github.com/draft-js-plugins/draft-js-plugins/issues/457

user can specify their own regex to let mention plugin to support any kind of Unicode as they want


mzbac:
@nikgraf could you please review it?

nikgraf:
@mzbac No that I see more and more stuff coming in I suggest we go a different route: We contain all of them as a default, but allow to overwrite the complete regex after the trigger. In the documentation we can explain our default regex and give examples if they want different use-cases.

There might be people only wanting auto-completion for japanese characters, but not alphabethic and so on.
What do you think?

mzbac:
@nikgraf sounds good.. let's close this PR..

PR #467, opened 2016-10-03
Adds optional `leftOffset` arg to toolbar decorator

poltak:
- the left absolute positioning calculation in `Tooltip` in `utils/portal.js` seems to work fine for standard text-selection toolbar
- left positioning seems to be off for decorated components in certain cases, however
- this optional arg to the decorator allows a number of pixels to offset the left positioning calculation by, in the case that it's a little bit off
- defaults to 0 to avoid unwanted offsetting
- tests passing locally

### DIFF
--- a/draft-js-toolbar-plugin/src/decorators/toolbar.js
+++ b/draft-js-toolbar-plugin/src/decorators/toolbar.js
@@ -10,7 +10,11 @@ const getDisplayName = (WrappedComponent) => {
 let number = 0;
 
 // HoverToolbar decorator will render a toolbar on hovering the WrappedComponent
-export default ({ theme, customRender }) => (WrappedComponent) => class FocusedToolbarDecorator extends Component {
+export default ({
+  theme,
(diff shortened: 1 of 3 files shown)


poltak:
After playing around with different versions of this package, it looks like the positioning issue I was facing was due to a change in the latest version of the package (`@2.0.0-beta5`). Downgrading to `1.0.5`, the toolbar positions itself perfectly (on both the standard text-selection toolbar and toolbar decorated components, like images). Going to close this PR and see if I can see if I can find the change that broke in the latest version instead (when I have time). I suspect it's related to some change in the positioning math logic in the `Tooltip` component in `utils/portal.js`.

nikgraf:
thx, any insights so far?

poltak:
@nikgraf No, sadly. I went through each of the changes made in commits between the npm package versions and could not reproduce it :S only when I explicitly npm update to the `2.0.0-beta5` version.

PR #471, opened 2016-10-05
Fix for DND onDropBlock overwriting custom Entity types

poltak:
- in the case where the Entity associated with the given block is of a custom type, the current onDropBlock function overwrites the Entity type with whatever type the ContentBlock is
- this commit allows optional specification of entity type in the addBlock modifier function, meaning the newly created Entity will be typed correctly as it was before being dragged-n-dropped
- if unspecified in the addBlock modifier function, it will fallback to the old behaviour of creating an Entity with whatever type the ContentBlock was using
- tests and linting passing; no build errors

### Example:
- I have an atomic-typed ContentBlock associated with a Entity of type 'blah'
- dragging and dropping that block somewhere else will recreate the Entity in the new selection but replace the type with 'atomic', as that's what the block is typed as (there is currently no separation of concern between ContentBlock type and Entity type)

### DIFF
--- a/draft-js-dnd-plugin/package.json
+++ b/draft-js-dnd-plugin/package.json
@@ -1,7 +1,8 @@
 {
-  "name": "draft-js-dnd-plugin",
+  "name": "@poltak/draft-js-dnd-plugin",
   "version": "2.0.0-beta5",
   "description": "Dnd Plugin for DraftJS",
+  "private": false,
   "author": {
     "name": "Benjamin Kniffler",
     "email": "benjamin@kniffler.com"
--- a/draft-js-dnd-plugin/package.json
+++ b/draft-js-dnd-plugin/package.json
@@ -1,8 +1,7 @@
 {
-  "name": "@poltak/draft-js-dnd-plugin",
+  "name": "draft-js-dnd-plugin",
   "version": "2.0.0-beta5",
   "description": "Dnd Plugin for DraftJS",
-  "private": false,
   "author": {
     "name": "Benjamin Kniffler",
     "email": "benjamin@kniffler.com"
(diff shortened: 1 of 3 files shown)


draft-js-dnd-plugin/package.json
```diff
@@ -1,7 +1,8 @@
 {
-  "name": "draft-js-dnd-plugin",
+  "name": "@poltak/draft-js-dnd-plugin",
```
nikgraf: Can you revert this back to `draft-js-dnd-plugin`?

draft-js-dnd-plugin/package.json
```diff
@@ -1,7 +1,8 @@
...
   "version": "2.0.0-beta5",
   "description": "Dnd Plugin for DraftJS",
+  "private": false,
```
nikgraf: Can you remove this line 😄

draft-js-dnd-plugin/src/modifiers/addBlock.js
```diff
@@ -1,7 +1,7 @@
 import { List, Repeat } from 'immutable';
 import { Modifier, CharacterMetadata, BlockMapBuilder, EditorState, ContentBlock, Entity, genKey } from 'draft-js';
 
-export default function (editorState, selection, type, data, text = ' ') {
+export default function (editorState, selection, type, data, entityType = undefined, text = ' ') {
```
nikgraf: What's the benefit to set this to undefined by default? Isn't it undefined except it is defined?

poltak: Good point. Silly oversight.

PR #479, opened 2016-10-11
[draft-js-mention-plugin] do open in componentWillReceiveProps() if and only if needed

gfx:
With draft-js-mention-plugin 2.0.0-beta5 (and 1.1.2), sometimes I can't use keyboard ↑ and ↓ on mention entities.

This is because once `openDropdown()` is called, keyboard event handlers are set and the events occurring are handled by them. This behavior, I guess, bans users from moving cursors up and down by keyboards.

This PR make `onEditorStateChange()` doesn't immediately call `openDropdown()`. Instead, `componentWillReceiveProps()` do so if `nextProps.suggestions` is not empty.

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/index.js
@@ -39,6 +39,11 @@ export default class MentionSuggestions extends Component {
   componentWillReceiveProps(nextProps) {
     if (nextProps.suggestions.size === 0 && this.state.isActive) {
       this.closeDropdown();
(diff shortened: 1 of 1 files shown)


nikgraf:
@gfx can you send me an example when that's the case? I wanna make sure it completely understand the issue to make an informed decision if this is the best way to solve that issue

gfx:
Because I've decided to implement my own completion plugin, I'm no longer interested in this PR.

Thanks.

nikgraf:
@gfx glad you fixed. Interested to publish your work? Might be helpful to others …

PR #480, opened 2016-10-11
[draft-js-mention-plugin] add support for korean characters

FourwingsY:
add korean trigger for mention plugin

### DIFF
--- a/draft-js-mention-plugin/src/mentionSuggestionsStrategy.js
+++ b/draft-js-mention-plugin/src/mentionSuggestionsStrategy.js
@@ -7,5 +7,6 @@ export default (trigger: String) => (contentBlock: Object, callback: Function) =
   // common chinese symbols: \u4e00-\u9eff - http://stackoverflow.com/a/1366113/837709
   // hiragana (japanese): \u3040-\u309F - https://gist.github.com/ryanmcgrath/982242#file-japaneseregex-js
   // katakana (japanese): \u30A0-\u30FF - https://gist.github.com/ryanmcgrath/982242#file-japaneseregex-js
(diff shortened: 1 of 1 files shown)


nikgraf:
hey @FourwingsY, happy to merge it. Do you have some kind of resource that demonstrates the unicode range?

I only found this, but it doesn't seem to match with your PR: http://stackoverflow.com/questions/38156300/regex-how-do-you-match-korean-hangul-letters-in-javascript-es6/38156301

FourwingsY:
Um, well, I just found the range in Wikipedia. [here](https://en.wikipedia.org/wiki/Korean_language_and_computers#Hangul_in_Unicode)
And, I will explain about that. I found some resources, but that was not in english. :(

There are 5 different ranges for Hangul, which is
- Hangul Syllables (AC00-D7A3) which corresponds to (가-힣)
- Hangul Jamo (1100–11FF)
- Hangul Compatibility Jamo (3130-318F)
- Hangul Jamo Extended-A (A960-A97F)
- Hangul Jamo Extended-B (D7B0-D7FF),
and the range in link, you found, seems like (1)merging all of these ranges, and (2)filtered for used letters. 
(1) I just used range for Syllables. because just a Jamo(letters) may not have any meaning, and doesn't make a word. and for last few days, I found some examples that using just Jamos for mentionable thing's name. I'll add that \u3130-\u318F range.
(2) while korean single character (like 한) is composed with three letters(ㅎ, ㅏ, ㄴ), so korean keyboards is made to type these letters, and OS will compose it. I can make any characters like 걃, but this character is never used in normal situations. so, some of fonts or codes like that doesn't include some characters.
And last, I won't add Hangul Jamo(\u1100–\u11FF) or Extended versions of it. these letters can't be inputed by korean keyboard. these are treated as special characters.

Thanks for review 😄

FourwingsY:
sorry, i misused this branch for my own use. i reset the branch as the merge-able state.

nikgraf:
This is a really good explaination! Thx man

PR #482, opened 2016-10-12
[draft-js-mention-plugin] add MentionSuggestions#isActive()

gfx:
This is useful if one should know the MentionSuggestions is open or not.

For example, when MentionSuggestions is active, any other keyboard events (i.e. ↑↓) should be handled only by MentionSuggestions.

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/__test__/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/__test__/index.js
@@ -71,11 +71,11 @@ describe('MentionSuggestions Component', () => {
     );
 
     suggestions.instance().openDropdown();
-    expect(suggestions.state().isActive).to.equal(true);
+    expect(suggestions.instance().isActive()).to.equal(true);
 
     suggestions.setProps({
       suggestions: fromJS([]),
     });
-    expect(suggestions.state().isActive).to.equal(false);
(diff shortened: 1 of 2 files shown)


nikgraf:
@gfx how do you use it in React then? Do you leverage the ref? What about callbacks for onOpen & onClose?

gfx:
Actually I use onOpen & onClose callbacks like this:

```jsx
onOpen={() => { this.isSuggestionActive = true; }}
onClose={() => { this.isSuggestionActive = false; }}
```

However, the `isSuggestionActive` seems the same as the `isActive` in the state. If `isActive` is a public accessor, `isSuggestionActive` is no longer required. Very simple.

nikgraf:
@gfx to use `isActive` you need to set a ref first and then call `isActive()` on the ref. Both ways require two lines of code to achieve the same thing. In addition calling instance methods is possible, but not really recommended in React except there is no other way.

Hmmm, maybe I'm missing the point, but using `onOpen` and `onClose` like you do looks good.

gfx:
@nikgraf That's because I'd like to know it is active or not. I'm not interested in the callbacks `onOpen`, `onClose`. 

BTW, I've decided to implement my own completion plugin, so I'm no longer interested in this PR and https://github.com/draft-js-plugins/draft-js-plugins/pull/479 .

Thanks.

PR #483, opened 2016-10-13
[draft-js-mention-plugin] mentionPrefix as part of draft editor node

YoonjiJang:
mentionPrefix is not in editorState since it's added as a plain text node, not draft editor node. This creates a problem when the plugin is used with CharCounter. (See example https://drive.google.com/open?id=0B3Ftrnd9yXx-SzQ2YndPR0g4YW8)

I'm not sure if this behavior is intentional. Feedback needed.

### DIFF
--- a/docs/client/components/pages/Mention/index.js
+++ b/docs/client/components/pages/Mention/index.js
@@ -153,7 +153,7 @@ export default class App extends Component {
           </div>
           <div className={styles.param}>
             <span className={styles.paramName}>mentionComponent</span>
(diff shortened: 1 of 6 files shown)


nikgraf:
This makes way more sense and was clearly not intentional. @yjang1031 can you add an entry to the Changelog of the mention plugin about this breaking change. I going to merge it then 🙌

yjang1031:
@nikgraf done

nikgraf:
nice, good job @YoonjiJang

lookapanda:
Is this merge already in a released version? I need this functionality.

nikgraf:
@lookapanda was released earlier today as 2.0.0-beta6

Issue #488, opened 2016-10-15
Allow for refreshing editorState


MarkNijhof:
In order to load new editorState after the pluginEditor has mounted we need a way to decorate the new editorState with the plugins loaded in the editor. If you just change the editorState in the props of the pluginEditor then it is not decorated atm and will load the editorState without the plugins applied. 

I have experimented with redecorating it in the componentWillRecieveProps but because of the cycle state.editorState -> pluginEditor.props.editorState -> onChange(editorState) -> this.setState(editorState) it becomes an unending loop.

My solution is to change:

```
  componentWillMount() {
    const compositeDecorator = createCompositeDecorator(
      this.resolveDecorators(),
      this.getEditorState,
      this.onChange);
    const editorState = EditorState.set(this.props.editorState, { decorator: compositeDecorator });
    this.onChange(moveSelectionToEnd(editorState));
  }
```

into:

```
    componentWillMount() {
        const editorState = this.decorateEditorState(this.props.editorState);
        this.onChange(moveSelectionToEnd(editorState));
    }
    decorateEditorState(editorState) {
        const compositeDecorator = createCompositeDecorator(
            this.resolveDecorators(),
            this.getEditorState,
            this.onChange);
        return EditorState.set(editorState, { decorator: compositeDecorator });
    }
```

So that in the component using the pluginEditor I can create/load a new editorState then decorate it with the plugins from the pluginEditor and put it in the props. For example like this:

```
    loadAst(ast) {
        const contentState = importer(ast);
        const editorState = EditorState.createWithContent(contentState);
        const decoratedEditorState = this.refs.pluginEditor.decorateEditorState(editorState);

        this.setState({
            editorState: decoratedEditorState
        });
    }
```

The downside of this is that you need a reference to the pluginEditor, but I don't see only one way how this can be prevented and that is if we can "discover" what plugins the editorState is decorated with and if that doesn't match the loaded plugins in the pluginEditor to re-decorate the editorState with the missing plugins.

If we can achieve that then this will also enable the loading of new plugins relatively easily as on each componentWillReceiveProps you would check if all needed plugins are loaded and load the missing ones. Maybe we can even unload plugins not loaded in the pluginEditor anymore. completely solving both this issue and #42 

I can see in the editorState that "decorator" is null so it should be easy to actually solve this in the componentWillReceiveProps the first time, but if we can add some meta-data that shows what plugins are loaded then we can do the loading and unloading automatically based on that list.

[...]

Issue #42, opened 2016-03-04
Allow to update the plugins and set new decorators


nikgraf:
This issue needs to be fixed first:

```
hellendag [12:06 PM] 
:simple_smile:

[12:07] 
The only thing to be careful of with setting a new decorator right now is that `beforeInput` allows native character insertion, so that's causing some problems

[12:07] 
I need to come up with a good solution for that

[12:07] 
So right now, if you set a new decorator in `onChange` and the new state is coming from a `beforeInput`, there can be some rendering bugs

nikgraf [12:08 PM] 
I think I had this issue

hellendag [12:08 PM] 
Double character rendering, mostly

nikgraf [12:08 PM] 
I type in character `a`and it results in 2 a `aa`

[12:08] 
exactly

hellendag [12:08 PM] 
Yeah

[12:08] 
So right now, the way around that is to propagate the new state with a `setTimeout`, to allow the native insertion

[12:08] 
It works, but it's not pretty

[12:09] 
And I need to come up with something better
```

MarkNijhof:
I would really like to see this happen :) and I would like to try out something myself:
`
componentWillMount() {
  this.loadPluginsAndDecorators();
}
componentWillReceiveProps(nextProps) {
  if (!Immutable.is(this.props.plugins, nextProps.plugins) { //make plugins an Immutable.List
    this.unloadPluginsAndDecorators(); // logic from the componentWillUnmount
    this.loadPluginsAndDecorators(); // logic from the constructor + componentWillMount
  }
}
componentWillUnmount() {
  this.unloadPluginsAndDecorators();
}
`
How does this look?

Btw I am not following why you do this in the constructor `const plugins = [this.props, ...this.resolvePlugins()];` why add this.props to the array?

juliankrispel:
Closing for issue cleanup - If you need this issue and think it should remain open please let us know why, thank you

PR #489, opened 2016-10-15
Auto decorate editState when no decorators are set plus update when plugins or decorators change

MarkNijhof:
This is my proposed fix for both these issues. It is a breaking change in the API as now plugins and decorators are expected to be an Immutable.List() vs plain array. Internally it is still converted back to a plain array using toArray(), but I think we can look into not doing that either.
- Allow to update the plugins and set new decorators #42 
- Allow for refreshing editorState #488


MarkNijhof:
Btw reason for using Immutable.List is easy comparison of this.props.plugins - nextProps.plugins and this.props.decorators - nextProps.decorators I guess this can be solved differently as well if you don't want a breaking change, but I understand 2.0 has breaking API changes anyway.

mzbac:
@MarkNijhof  can you do array -> list conversion inside plugin-editor? therefore we don't have API level changes.

MarkNijhof:
@mzbac yes that is of course possible, but that would mean the plugins and decorators array would be converted 2 times one each props change, once for the current props and once for the nextProps. Even if we were to store them into state it would still happen once on every props change. Currently this conversion still happens going back toArray() but I want to remove that need as well.

draft-js-plugins-editor/CHANGELOG.md
```diff
@@ -14,6 +14,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - `decorators` don't decorate plugins anymore.
 
+### Changed
+
+- *Breaking change* The `plugins` and `decorators` editor props need to be an Immutable.List()
```
nikgraf: I have concerns regarding the API design here. All other props are plain functions, but only the plugins need the list. That's inconsistent and I'm worried about that.

MarkNijhof:
I can revert the upgrade of draft-js

draft-js-plugins-editor/CHANGELOG.md
MarkNijhof: editorState is of course also a complex object. I choose to use Immutable.List for both plugins and decorators as it makes comparing them in the componentWillReceiveProps a lot more fault tolerant. And since this check happens on every props change I also wanted to optimise for that. Do you have a different suggestion on how to solve that part?

nikgraf: The issue I see is that even if you make the plugins a list a plugin might change it's behaviour and add or remove decorators. I wouldn't recommend it, but I can see this being useful for advanced editors. So in fact even if we compare the plugins the decorators might have changed.

You started to make me think in another direction.

```js
import Editor, { decoratorFromPlugins } from 'draft-js-plugins-editor';
import createImagePlugin from 'draft-js-image-plugin';
import createToolbarPlugin from 'draft-js-toolbar-plugin';
import { EditorState } from 'draft-js';

const imagePlugin = createImagePlugin();
const toolbarPlugin = createToolbarPlugin();
const plugins = [imagePlugin, toolbarPlugin];
const decorator = decoratorFromPlugins(plugins);

const editorState = EditorState.createEmpty(decorator);
```

nikgraf: I like it because it resembles the Draft.js API and the other hand it makes things more complicated.

Right now I have the feeling either we keep the current behaviour + add `decoratorFromPlugins` for advanced usage or we set the decorators on every new editorState and test that it's not causing performance issues especially with a lot of content.

MarkNijhof: I compare both the plugins and decorators, not sure if that is what you meant.

The proposed code looks nice as well, only downside I see that now you have two different places to work with plugins. Or do you want this to replace the plugins and decorators properties so you always set them outside the editor? 

I would like to "hide" this knowledge from the user, you set plugins via props then I expect them to update when the props value updates, this is how React generally works.

nikgraf: @MarkNijhof Totally agree! The trouble we face is that the decorator is actually part of the EditorState and not passed to the Editor in Draft.js.
see https://github.com/facebook/draft-js/blob/master/examples/tweet/tweet.html#L49

If I'm not mistaken you can't compare the decorators by reference, because they are not an immutable object. We could also change that, but then it this would diverge from Draft.js and be inconsistent with the rest of the plugin architecture. It's a tricky edgecase.

MarkNijhof: I am not sure yet how to deal with decorators set outside of the editor, need to think about that a bit

MarkNijhof:
Added the latest changes from master into this one, I am __really__ eager to find a solution for this as it is needed for my project (this or a solution that achieves the same results).

MarkNijhof:
Going to redo this from scratch based on the conversations I have had on Slack

PR #494, opened 2016-10-19
[draft-js-mention-plugin][feature] Add onSelect props to get the mention object selected

dimaslz:
Hi team!

We are using draftjs and some plugins. And, in this case, we have a need with mentions. We need that when you select a mention, can get the object that is selected.

I think that could be util this feature :)

Thanks!


nikgraf:
@dimaslz why do you need to know when a mention has been selected? what's the value here?

In any way since onselect in html means something completely different I believe it should be either `onChange` or `onMentionSelect`

dimaslz:
Hi @nikgraf 

We need that to save the users data mentioned in the message in other object.

Example:
```js
// when the user type a message with a mentions
bla bla bla @username1 bla bla @username2
// we want keep it in a object like
let montionsInMessage = [
  {
    username: "username1",
    userId: "123",
    ...
  }, {
    username: "username2",
    userId: "134",
    ...
  }
];
```
or something like this.

About the naming, yes, I agree. I will update!

nikgraf:
sounds like a good usecase. one more question. Sorry for requesting another change, but from you comments I noticed it's more about when the mention is added not selected in the Dropdown. Can you change it one more time to `onAddMention` or is this more confusing from your perspective?

dimaslz:
Hi @nikgraf 

For me the action `onMentionSelected` is totally correct. For me the idea to `onAddMention` and onMentionSelect is the same. But if you think that could be `onAddMention` for special something, I can fix it.

nikgraf:
Let's go with `onAddMention` please 👍  and can you resolve the merge conflict? Sorry made some changes this weekend.

dimaslz:
Sorry for the delay, I agree. I will change it and fix the conflict.

Thanks, and don't care, it normal try to discuss good naming.

dimaslz:
DOne, let me know if you need something.

Thanks!

john-osullivan:
Is this going to get merged anytime soon?  I was just about to sit down and write up my own version of the same feature, my app needs to hook in and know when a mention gets added.  Thanks to all of you for writing up a great component!

nick-gudumac:
Also, can we add onRemoveMention props as well? I my use case, I am saving mentions apart from the editor content, so there is need for triggering adding / removal of mentions while editing the content.

dimaslz:
@nick-gudumac mmm, I don't thought this case, but could be good.

I will check it and add in this PR o maybe in other one.

Thanks!

dimaslz:
@nikgraf Hi!

Can you check it and let me know if it is ok? I trying to do the suggestion by @nick-gudumac but I prefer do that in other PR, because the context is other and I guess that I need time to find the good way to do.

Thanks!

nikgraf:
@dimaslz sorry, took me a while since I focused on getting a couple other plugins ready

PR #497, opened 2016-10-20
Sidebar plugin

Nihaux:
Doc is [here](https://github.com/Nihaux/draft-js-plugins/tree/sidebar-plugin/draft-js-sidebar-plugin), Demo [there](http://5.39.83.92/plugin/sidebar)

feedbacks are welcome!

Edit: demo show full embed implementation using iframely as proxy
Edit2: tested on chrome, some issues to fix on firefox.


nikgraf:
@Nihaux I just merged a side toolbar I worked on for the last couple days. Wanna combine our efforts and improve it together? I'm super curious what you think about the structure

Nihaux:
@nikgraf hello, sry for the delay, work stuff and all.
Just tried the sidebar on my project. 

First thing that comes to mind is that the design is not adapted for my use case. You must have a 80px+ marge on the left side of the editor which does not work is my project. I think the "medium like" toolbar (button on the left which opens a menu over the editor) is more all-purpose.
But this is just a css / customization remark.

Regarding the functionnalities, our toolbars have different purpose.
Mine is a way to inject custom entities wrapped in an atomic block, yours is a way to switch the block type.

I could try to integrate what I have done into this one. but have you looked at what my toolbar does, and are you ok with it ? Don't want to work at something for nothing...

nikgraf:
Hey @Nihaux, while I'm not 100% sure yet I envisioned the sidebar as a tool to contain one or many dropdown/buttons. More like the Facebook notes editor:

screen shot 2016-11-15 at 00 54 11

As you said while most of it would be simply css / customization I care about a good default experience. The need of multiple buttons is probably not idea. We could stack them vertically, but I'm not sure if there is a way this can look nice.

If you could help out creating an add button and add one example to the image documentation page (image is not simple an atomic block as well) that would be extremely cool. But before we start we better probably brainstorm what a good experience would look like.

My goal for an ideal sidebar:

- By default "Add new custom blocks"-feature and "Modify text blocks"-feature
- Customisable in the sense that people can add their own dropdowns or buttons

Any ideas how to best achieve that and make it work nice on mobile & desktop?

nikgraf:
@Nihaux ping :)

PR #516, opened 2016-10-27
added callback on link found for linkify

cristiandan:


nikgraf:
@cristiandan why do you need it? What's the use-case? 🙂

cristiandan:
@nikgraf if you want to call some custom code when there are links in the textbox, and don't want to run a second linkify to call your code when there's a link. 
I mean, it's more efficient to have just one linkify running, so I was thinking: why not use the existing one.
I need it for an app that display the images from the sait page if there is a link in the text box. But the code in to display the image needed to use a separate linkify object to search for links

johnmpotter:
I would love to have a callback for when links are found as well. 🙂

My use case is essentially the same as the social post textbox found on Facebook, Linkedin or Twitter. When a link is pasted in the textbox, we scrape the url and display a 'card' for the url.

I have no feedback on whether this is the right implementation because I don't know the code well enough, but thought i'd share my use case. Thanks again for all your work on this 👍

nikgraf:
Hey @cristiandan & @johnmpotter … after some consideration I suggest a fork of the plugin. Your use-case makes a lot of sense, but I have seen only very few people in need of this. Considering every feature complicates the docs even further I rather leaves this out. I hope you can understand my reasoning here.

PR #521, opened 2016-10-31
Add onRefChanged as a callback

AnilRedshift:
Components that keep UI outside of the <Editor> should not assume that the ref is immutable during the lifetime of the plugin.

This change adds an `onRefChanged` that behaves similarly to `onChange` when the editor ref changes

### DIFF
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -123,6 +123,15 @@ class PluginEditor extends Component {
     getEditorRef: this.getEditorRef,
   });
 
+  editorRefChanged = (editor) => {
+    this.editor = editor;
+    this.resolvePlugins().forEach((plugin) => {
+      if (plugin.onRefChanged) {
+        plugin.onRefChanged(editor);
+      }
+    });
+  };
+
   createEventHooks = (methodName, plugins) => (...args) => {
     const newArgs = [].slice.apply(args);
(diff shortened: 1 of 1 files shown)


nikgraf:
ohhhh, this I didn't consider.

@AnilRedshift is it really necessary? If we use the getEditorRef function it always should return the latest ref. What's your use-case?

max-winderbaum:
@nikgraf We need to attach `mousemove` handlers onto the editor component to track hover, and therefore need to know when and if the component changes so we can reattach them.

IMO this is a good idea to include as part of an API so users can avoid polling for changes :)

juliankrispel:
Looking into this right now and struggling to come up with use-cases. When does your editor ref actually change?

AnilRedshift:
I'm no longer working on this project. @max-winderbaum: Do you have a need to continue this PR? Otherwise we can go ahead and close it.

max-winderbaum:
@juliankrispel The editor ref would change if, for example, you are using `react-router` and switching to and from a page that had a `draft-js` editor on it.

juliankrispel:
@max-winderbaum do you have an example somewhere by any chance? Like a codesandbox so I can reproduce the problem? Sorry for the late reply :(

juliankrispel:
@max-winderbaum I'm going to close this for now since I'm struggling to reproduce the problem tbh. If you find a way to reproduce the issue in a sandbox I'd be happy to have another look @max-winderbaum  :+1:

PR #525, opened 2016-11-04
fix broken image from Adrian Li

tobiaslins:
changed to link from current twitter profile image from @adrianmcli


nikgraf:
thx @tobiaslins 👊

PR #526, opened 2016-11-04
Add reloadDecorators function to pick up new decorator props manually

max-winderbaum:
There is a use case for loading different plugins and decorators after the PluginEditor mounts for the first time.

Accepting new `plugins` props from a `componentWillReceiveProps` hook is a difficult problem to solve. This is because because the `plugins` prop is just a plain JavaScript object and performing deep equality checks isn't possible every time a component will receive props.

We *could* do a strict equality check, but this would be a breaking change as to be consistent we would have to enforce that all consumers of draft-js-plugins-editor use immutable data structures. This is less than ideal.

For now, I'm proposing that we let the consumer of the editor decide when to manually reload decorators. I believe there is a use case for reloading plugins in general, but I want to run this general concept by @nikgraf to see his thoughts before continuing to develop further.

### DIFF
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -85,6 +85,28 @@ class PluginEditor extends Component {
     }
   };
 
+  getPlugins = () => this.props.plugins.slice(0);
+  getProps = () => ({ ...this.props });
+
+  // TODO further down in render we use readOnly={this.props.readOnly || this.state.readOnly}. Ask Ben why readOnly is here just from the props? Why would plugins use this instead of just taking it from getProps?
+  getReadOnly = () => this.props.readOnly;
+  setReadOnly = (readOnly) => {
+    if (readOnly !== this.state.readOnly) this.setState({ readOnly });
+  };
+
+  getEditorRef = () => this.editor;
+
+  getEditorState = () => this.props.editorState;
+  getPluginMethods = () => ({
+    getPlugins: this.getPlugins,
+    getProps: this.getProps,
+    setEditorState: this.onChange,
+    getEditorState: this.getEditorState,
+    getReadOnly: this.getReadOnly,
+    setReadOnly: this.setReadOnly,
+    getEditorRef: this.getEditorRef,
+  });
+
   loadDecorators = () => {
     const decorators = this.resolveDecorators();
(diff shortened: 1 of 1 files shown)


MarkNijhof:
We had a short discussion about this on Slack already,

I don't like this solution, I think we should not be to afraid to break the API if it serves the greater good. And I think this case is such a case. I like this solution if we would completely get rid of the props.plugins and pass the plugins into the .loadPlugins(plugins) function that way there is one way to load plugins and do the same for the decorators. Then there is no difference between first load or sequential loads.

Having said that I like the thing we discussed as well even better where we would make reloading of plugins possible via props change if the plugins array is of type Immutable.List if it is that then compare and reload the plugins if it is a plain Array then ignore changes in the props.plugins. This is both backwards compatible and enables new functionality. The change is minimal and I'll fix it this way in my PR #489 The logic is almost identical to this PR except the .loadDecorators() is called in the componentWillReceiveProps().

It will be slightly more complex as we have to deal with plain arrays needing to be converted to Immutable.List in componentWillMount so I think plugins and decorators should move to state and update them if the props have changed. I'll need to think a bit more about how to do this exactly.

nikgraf:
@MarkNijhof I'm happy to break the API in a 3.0 version, but then I would love to explore more options. Currently I'm focusing solely on finishing up the Image support and toolbars to cover the use-cases of  most of the people using draft-js-plugins. Sorry dude! btw always feel free to for the editor and roll your own version on npm.

@max-winderbaum your approach sounds reasonable for the current API. Wanna proceed?

MarkNijhof:
Makes sense, and I'll probably do that.

Btw @max-winderbaum I think that you are not initing plugins on reload

max-winderbaum:
@nikgraf Yes please - let me write some tests and investigate creating a similar function to reload plugins - right now we're only reloading decorators.

Thanks for reviewing! I should have the updated code in the next couple days.

nikgraf:
@max-winderbaum do you have rough idea when you get to it?

max-winderbaum:
I'm sorry I let this slide - I just haven't had the time and it looks like I won't into the near future. I'm going to close this PR but someone else should feel free to take up the gauntlet here.

nikgraf:
@max-winderbaum too bad, good luck with your work 👍

PR #534, opened 2016-11-10
add missing props to render of MentionSuggestions

Arinteck:
Without this props in render method of MentionSuggestions component you may have warning:
_`Warning: Unknown props `onOpen`, `onClose` on <div> tag`_


nikgraf:
thx @Arinteck 🙌

PR #536, opened 2016-11-12
Fix `visisble` typo in toolbar plugins

Rosey:


nikgraf:
thanks for submitting this fix @Rosey 👌

PR #553, opened 2016-11-17
feat(mention-suggestions): Adds a property `triggerOnEmpty`

lookapanda:
This prop allows onSearchChange to be triggered, even if the search value is empty. Useful for dynamic remote data fetching.

Out of the documentation entry:

> If provided, the search will be triggered even if the search value is empty. This can be useful if you want to fetch remote data whenever the user hits `@`. By default, this is set to `false`.


nikgraf:
I discussed it with @lookapanda and we learned it's actually a bug that he discovered. Thx for the effort though! I fixed it here: https://github.com/draft-js-plugins/draft-js-plugins/pull/567

PR #556, opened 2016-11-19
Moved entityMutability default value for consistency

Radivarig:


nikgraf:
good one @Radivarig - love it when people care about consistency and actually take the effort to submit a PR!

PR #561, opened 2016-11-22
linkify transform

sam-jay:
Makes linkify accept an optional transform function in the config, and then applies this transformation to all hrefs.

### DIFF
--- a/draft-js-linkify-plugin/README.md
+++ b/draft-js-linkify-plugin/README.md
@@ -20,6 +20,14 @@ const linkifyPlugin = createLinkifyPlugin({
 });
 ```
 
+You may optionally provide a transformation function to transform the links:
+
+```js
+const linkifyPlugin = createLinkifyPlugin({
+  transform: (href) => { return href; // identity transformation }
+});
+```
+
 ## Importing the default styles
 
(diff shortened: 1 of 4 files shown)


nikgraf:
@sam-jay interesting proposal! 'm curious about your use-case. Why do you need to transform hrefs?

sam-jay:
I'm using draft.js in an electron app and clicking the links in the editor currently causes the electron renderer window to navigate to that link. I want the link to be opened externally in the users default browser. I can globally intercept all navigation events and prevent the default behavior but I need to restrict this just to the links inside the draft.js editor. If I can transform the hrefs so that they start with 'open-external://' then I will be able to selectively prevent the default behavior when I intercept navigation events.

nikgraf:
@sam-jay can you add the change to the CHANGELOG.md of the linkify plugin. In addition please add it to the docs so the `transform` config shows up in the docs here: https://www.draft-js-plugins.com/plugin/linkify. Here the file: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/docs/client/components/pages/Linkify/index.js

In the README as well as in the docs please mention your use-case with a real world example on how the transform should look like for electron.

PR #563, opened 2016-11-22
Fix toolbars package names in docs

mercury2269:
It appears that docs are referencing old package names which don't exist in npm.


nikgraf:
@mercury2269 ups, thx for the fix and especially fixing the line endings! Makes my life so much easier if people care about making CI pass 👍

PR #587, opened 2016-12-06
Update docs README, side-toolbar-plugin, hashtag-plugin

StevenIseki:
- Incorrect css import in side-toolbar-plugin
- Fix custom hashtag example
- Update docs README

Note a few other small doc updates have yet to be pushed live to draft-js-plugins.com

Thanks


nikgraf:
amazing, thx for looking into this @StevenIseki 👍

PR #588, opened 2016-12-06
Add theme support to side and inline toolbar plugins

StevenIseki:
- Add theme support to side-toolbar-plugin
- Add theme support to inline-toolbar-plugin


mzbac:
@nikgraf  could you please review?

nikgraf:
@StevenIseki this is so good! thx for taking the effort! btw I will add another example to both where you can choose the set up buttons in the toolbars

nikgraf:
@mzbac thx for the reminder here :)

PR #596, opened 2016-12-09
Patch blockStyleFn

sugarshin:
Hi,

I think that you should exclude null in the return value of `blockStyleFn` .

refs:

- https://github.com/facebook/draft-js/blob/master/src/component/contents/DraftEditorContents.react.js#L152
- https://github.com/facebook/fbjs/blob/64208e08a409b9b4a078aa8d09a5ab3977d3b085/packages/fbjs/src/core/joinClasses.js#L23


nikgraf:
sounds reasonable to me, thx for your contributon 👍 

can you add this as a fix in here as well https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-plugins-editor/CHANGELOG.md

sugarshin:
Please check, Thanks

nikgraf:
perfect, thx @sugarshin

PR #597, opened 2016-12-09
Updated usage in docs [undo]

Radivarig:
Removed onChange and editorState props
Added passing undoPlugin to editor prop plugins


nikgraf:
thx @Radivarig

Issue #549, opened 2016-11-15
Moving React dependency to peerDependencies


chrisui:
Hey,

Just installed draft-js-plugins and noticed it pulls in the *latest* version of `react`. Typically I would've expected this to be a peer dependency since the consumer is in the charge of managing the versioning there.

Two questions:

1. Can we move `react` and `react-dom` to `peerDependencies` in `package.json` (perhaps there are other deps which should live there too?)
2. Does draft-js-plugins support previous versions of React? (.14 in my case)

Thanks!

nikgraf:
@chrisui sorry for the late reply. I kind of stopped looking at issues for the last month to actually focus on code and finish up 2.0

1. Didn't peerDependencies become obsolete with npm3?
2. Nope, we don't support 0.14.x

nikgraf:
ohhh, didn't know that. Sounds good to me. Interested to open up a PR?

sugarshin:
+1

bj00rn:
close?

PR #600, opened 2016-12-09
Move react to `peerDependencies`

sugarshin:
Hi,

Please check.

ref: https://github.com/draft-js-plugins/draft-js-plugins/issues/549


nikgraf:
@sugarshin so sorry, but can you re-do the PR? I merged some other changes and split up a plugin into two.

sugarshin:
@nikgraf I already push. Could you check this ?

nikgraf:
hmmmm, should we add this to every changelog? If people didn't have React in their dependencies npm would throw an error, correct?

sugarshin:
I think that as a behavior of a library that depends on React it is correct.

PR #601, opened 2016-12-09
Changed export default from varname to expression

Radivarig:


nikgraf:
@Radivarig would be awesome if you could submit a PR for all the other plugins as well 🙂

Radivarig:
Sure, #604

PR #606, opened 2016-12-12
Add image button to side toolbar

StevenIseki:
Add image button to side toolbar


StevenIseki:
draft-plugins-image-button

StevenIseki:
This should be pretty close, it works by passing in a `addImageFile` function to the store, which is triggered when the AddImage button is clicked. `parseUrl` is an optional function to pass in where you can parse the url from the file and possibly upload to S3 then get the actual file url to pass back to the image. At least that is my use case, most would be the same.

We need to publish a new version of the buttons with this, In `DefaultBlockTypeSelect/index.js` I updated the buttons import to `from '../../../../draft-js-buttons/src/';` so it will work but this will have to change back

tlenclos:
What do we need to merge this ?

marekolszewski:
+1

StevenIseki:
Just a note, this functionality is now present in the draft-js-sidebar-plugin in https://github.com/vacenz/last-draft-js-plugins as shown [here](http://lastdraft.xyz/v3) so this pr can be closed if required.

nikgraf:
Definitely check out https://github.com/vacenz/last-draft-js-plugins!

PR #607, opened 2016-12-12
Toggle side toolbar on click

StevenIseki:
Toggle side toolbar on click

### DIFF
--- a/draft-js-buttons/src/utils/createBlockAlignmentButton.js
+++ b/draft-js-buttons/src/utils/createBlockAlignmentButton.js
@@ -6,6 +6,7 @@ export default ({ alignment, children }) => (
 
     activate = (event) => {
       event.preventDefault();
+      event.stopPropagation();
       this.props.setAlignment({ alignment });
     }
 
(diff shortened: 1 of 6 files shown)


StevenIseki:
draft-plugins-click

StevenIseki:
Issues still to look at - 
- The clicking of the menu, is this better than on hover, my opinion it is and is a more common action, or do we want both and an option can be passed to the side-toolbar hover or click.
- The menu needs to hide when clicking back into the editor, currently a bug, should be fixable.

We need to publish a new version of the buttons with this, In `DefaultBlockTypeSelect/index.js` I updated the buttons import to `from '../../../../draft-js-buttons/src/';` so it will work but this will have to change back

StevenIseki:
Just a note, this functionality is now present in the draft-js-sidebar-plugin in https://github.com/vacenz/last-draft-js-plugins so this pr can be closed if required.

nikgraf:
Definitely check out https://github.com/vacenz/last-draft-js-plugins!

PR #610, opened 2016-12-13
Add link button to inline-toolbar/linkify

StevenIseki:
Add link button
Update inline-toolbar with addLink function
Add new strategy and LinkAdd component to Linkify
Update CustomInlineToolbarEditor example

### DIFF
--- a/draft-js-linkify-plugin/src/LinkAdd/index.js
+++ b/draft-js-linkify-plugin/src/LinkAdd/index.js
@@ -1,0 +1,130 @@
+import React, { Component } from 'react';
+import ReactDOM from 'react-dom';
+import linkifyIt from 'linkify-it';
+import modifier from '../modifiers/addLink';
+import styles from './styles.css';
+
+const linkify = linkifyIt();
+
+export default class LinkAdd extends Component {
+  // Start the popover closed
+  state = {
+    url: '',
+    open: false,
+    linkError: false
+  };
+
+  // When the popover is open and users click anywhere on the page,
+  // the popover should close
+  componentDidMount() {
+    document.addEventListener('click', this.closePopover);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('click', this.closePopover);
+  }
+
+  onPopoverClick = () => {
+    this.preventNextClose = true;
+  }
+
+  onKeyDown(e) {
+    if (e.keyCode === 13) {
+      e.preventDefault();
+      e.stopPropagation();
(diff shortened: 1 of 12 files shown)


StevenIseki:
draft-js-plugin-link

anton-aleksandrov:
Why was this feature removed? Is there equivalent in current version?

draft-js-linkify-plugin/src/LinkAdd/index.js
```diff
@@ -0,0 +1,130 @@
+import React, { Component } from 'react';
+import ReactDOM from 'react-dom';
+import linkifyIt from 'linkify-it';
+import modifier from '../modifiers/addLink';
+import styles from './styles.css';
```
mkhazov: this file should be renamed to `../linkAddStyles.css` in order to include it in `plugin.css` built with [concatCssFiles](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/scripts/concatCssFiles.js)

StevenIseki:
Just a note, this add link functionality is now present in the last-draft-js-toolbar-plugin in https://github.com/vacenz/last-draft-js-plugins as shown [here](http://lastdraft.xyz/v3) so this pr can be closed if required.

StorytellerCZ:
Few thoughts when looking this over:
I'm trying to implement a custom button with similar functionality. Doing that and looking at this PR I think it would be nice to have some easy way to build your own complex buttons. Either more types of button containers can be added or some general container that is going to expose more about the toolbar states would be nice.

In my case I need a button that is similar to the link button. Once click it open a field where user can type. In my case it searches through the DB and offers suggestions to the use. Once user selects an option it closes and decorates the selection. And so on.

If we can somehow extract what is in this PR into some general pattern that would allow the creation of complex buttons I think that would be a great service. This would make creating mentions or emoji buttons super easy.

I might give it a try, but I'm not sure if I'll have the time and knowhow (I'm not familiar with the codebase enough) to pull this off, so any help would be appreciated.

StorytellerCZ:
@StevenIseki Found a bug in the demo of last-draft. Once you add a link the toolbar stays open even if you select elsewhere. If you start typing it starts re-writing the text of the link and you are unable to escape.
BrowserL Firefox 53

stevewillard:
What's the status of this PR? Can it be merged?

StorytellerCZ:
@stevewillard Seems to me that this was superseded by #729.

PR #612, opened 2016-12-15
Fix bug if mentionTrigger length more than 1 character

itoldya:


nikgraf:
@itoldya with the current implementation the mention trigger is supposed to have only a lenght of 1. If you test it with a couple of edge cases this PR would fail. If we want to support that it would require a rewrite. I have an implementation in mind which is not based on decorators, but don't have the time right now to do a reimplementation.

Issue #531, opened 2016-11-07
inline-Toolbar wrong position when using twitter bootstarp.


navidmesbah:
Position of the inline-toolbar is calculated wrong when using twitter bootstrap as the skeleton of the css...

![screenshot_2016-11-07_22-48-13](https://cloud.githubusercontent.com/assets/10755936/20072085/5344b08e-a53c-11e6-9e7d-7cf8001aeb29.png)

Rosey:
I've noticed this too, not with twitter bootstrap but just in general - it seems to always set the position relative to the window, which could be wrong if it has a relative/absolutely positioned parent. The way emoji/mentions are positioned seems way more reliable, maybe that functionality could be ported over to the inline-toolbar as well?

I do find if you set `position: fixed` on the toolbar the position "works" but in the case of a scrollable parent, it doesn't scroll with its parent anymore, so I think it could still be made better by emulating how emoji and mention suggestions are positioned :)

jrock17:
I'm seeing the same thing (using Zurb Foundation). Any recommended fixes?

![2016-11-15_09-27-06](https://cloud.githubusercontent.com/assets/1270524/20316198/bc4a1f22-ab15-11e6-8361-91b0db4bf544.png)

jebarjonet:
I am facing a similar issue when the draft js input is nested in `div` that have `position: relative`, the toolbar position is messed up and is displayed somewhere else on the screen. When I set those parents `div` in `position: static`, it works perfectly. Unfortunately I need this `position: relative` property on those parents.

jrock17:
I'm still having the same problem after the beta7 update.

![toolbar](https://cloud.githubusercontent.com/assets/1270524/20676561/ffc880de-b544-11e6-8414-a54839b75ca7.png)

mkampik:
I've encountered the same problem. You can fix the inline-toolbar's position by ensuring that **all** of it's parent `divs` have their css `position` property set to its default value: `static`.

Would be nice to be able to position parent `divs` however you like. Not sure what the best implementation would be tho. Perhaps worth exploring in a separate issue?

jebarjonet:
[...]

PR #617, opened 2016-12-19
Fixing alignment for Side/Inline Toolbar

myyellowshoe:
- Added position fixed, to fix issue with incorrect position of toolbar per this issue (https://github.com/draft-js-plugins/draft-js-plugins/issues/531)

- Also fixed a small style issue where separator was missaligned, and had a white background.


dracudakid:
It's still in the wrong position :(

nikgraf:
@myyellowshoe thanks for looking into it! I merged this PR to solve the issue: https://github.com/draft-js-plugins/draft-js-plugins/pull/629

PR #629, opened 2017-01-04
Fix inline toolbar positioning to work with a relative parent fixes #531

### LINKED ISSUE #531: inline-Toolbar wrong position when using twitter bootstarp.

tomkelsey:
The inline toolbar plugin is currently positioned absolutely but the offset assumes there are no containers between the toolbar and the document body with a position other than `static`

This PR re-uses the `getRelativeParent` function from the mention and emoji plugins to correct the positioning of the toolbar to fix #531

Whilst the alternative fix in PR #617 is much better than it is at the moment, using `position: fixed` means that if the user scrolls whilst something is selected the toolbar follows them down the page.

I think/hope this PR creates a nicer experience.

### DIFF
--- a/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
+++ b/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
@@ -59,7 +59,7 @@ export default class Toolbar extends React.Component {
       <div
         className={theme.toolbarStyles.toolbar}
         style={this.state.position}
-        ref={toolbar => this.toolbar = toolbar}
+        ref={(toolbar) => { this.toolbar = toolbar; }}
       >
         {this.props.structure.map((Component, index) => (
           <Component
--- a/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
+++ b/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
@@ -4,6 +4,19 @@ import { getVisibleSelectionRect } from 'draft-js';
 // TODO make toolbarHeight to be determined or a parameter
 const toolbarHeight = 44;
 
+const getRelativeParent = (element) => {
+  if (!element) {
+    return null;
+  }
+
+  const position = window.getComputedStyle(element).getPropertyValue('position');
+  if (position !== 'static') {
+    return element;
+  }
+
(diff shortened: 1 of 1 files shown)


Rosey:
Thanks so much for doing this!

There's probably a good answer for this but for my own education can you explain why you wrote a custom `getRelativeParent` instead of using `offsetParent`? 😄

tomkelsey:
No worries!

A fair question - I'd imagine `offsetParent` would work and be considerably tidier! I went with `getRelativeParent` just because I dipped into the emoji plugin source code and saw how it was done there. No good reason really 😄

Rosey:
Ah yeah sorry, I read your commit more carefully after and saw you mentioned that you had copied it from elsewhere!

ramsgoli:
I could definitely use this! Any idea as to when this fix will be merged?

gavinwahl:
@tomkelsey Are you going to update this to use offsetParent? Maybe that would have a better chance of being merged.

Is this project maintained?

bharney:
Currently using Bootstrap and this fix worked perfectly. I did have to make some slight modifications to the code @tomkelsey submitted, but it was a breeze given his guidance. ill post my exact changes when I get a chance.

image

Rosey:
So awesome 😊 this fix is important to me too so I bump on getting it merged! (Maybe after @bharney's fixup though?)

asdine:
Any idea on when this PR will be merged ? 🙈

Rosey:
@nikgraf pleasssss 🙏

nikgraf:
Well done @tomkelsey! 👍

@Rosey @ramsgoli @gavinwahl @bharney @asdine thanks for your involvement and review! sorry that I was so unresponsive. I started to do too many things and currently swamped in 300 Github issues. If any of you is interested to help out long-term with DraftJS Plugins let me know. Would be very much appreciated 😃

jackychou00:
I have the same issue with side toolbar, is this also work for side toolbar?

nikgraf:
@jackychou00 I think so, wanna pick this up and create PR?

PR #635, opened 2017-01-11
[emoji-plugin] Make emoji highlightable

AndrewHamm:
Currently if you highlight the emojione emoji's, the highlight bar disappears. With these changes however you get a nice continuous highlight so you can actually see what you are selecting.


nikgraf:
Thanks @AndrewHamm, I just merged you PR after removing a span we don't need anymore with this change!

PR #642, opened 2017-01-23
add link to inline toolbar

marktron:
Based off https://github.com/draft-js-plugins/draft-js-plugins/pull/610, this should allow us to add a link button to the inline editor bar.


marktron:
Sorry, pushed to the wrong place!

PR #647, opened 2017-01-25
Take toolbar buttons out of the tab order

gavinwahl:
I have two editors side by side like this:

```xml
  <Editor />
  <InlineToolbar />
  <Editor />
  <InlineToolbar />
```

I expect to be able to tab from the first editor to the second, but instead I have to tab a bunch of times to get through all the [invisible!] buttons in the inline toolbar. You can't tab through inline toolbar buttons anyway, so we might as well completely remove them from the tab order.

### DIFF
--- a/draft-js-buttons/src/utils/createBlockAlignmentButton.js
+++ b/draft-js-buttons/src/utils/createBlockAlignmentButton.js
@@ -26,6 +26,7 @@ export default ({ alignment, children }) => (
             onClick={this.activate}
             type="button"
             children={children}
+            tabIndex="-1"
           />
         </div>
       );
--- a/draft-js-buttons/src/utils/createBlockStyleButton.js
+++ b/draft-js-buttons/src/utils/createBlockStyleButton.js
@@ -39,6 +39,7 @@ export default ({ blockType, children }) => (
             onClick={this.toggleStyle}
             type="button"
             children={children}
+            tabIndex="-1"
           />
         </div>
       );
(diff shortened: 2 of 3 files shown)


nikgraf:
Thanks for reporting the issue @gavinwahl. Makes sense to make the buttons not clickable in case they are not shown. Nevertheless setting tabindex to -1 permanently is bad for blind people using screenreaders. It should be switched with the visibility of the inline bar.

Are you interested to help out here and improve the PR?

gavinwahl:
@nikgraf Does it currently work with screenreaders? I thought about this concern but since you can't tab to the buttons anyway (tabbing unselects the text which hides the toolbar), I thought the same would happen in a screenreader.

nikgraf:
@gavinwahl It might now, but it's the goal to make it work with screenreaders. That's why it's still in beta 😄

davidsonsns:
I'm having the same problem, when I was contributing vizualizei this pull request. Any update forecast?

MilanLempera:
This is really unpleasant behavior. What prevents the release of this fix?

juliankrispel:
@gavinwahl we'd like to not deprecate this feature - buttons should be navigatable via tabbing. Can you make the change to only disable  tabbing when buttons are invisible/disabled?

gavinwahl:
Tabbing never worked. The buttons only appear when you have text selected, and pressing tab with text selected unselects the text which removes the buttons.

Anyway, I'm no longer working on the project that needed this. Maybe @MilanLempera or @davidsonsns can take over.

MilanLempera:
@juliankrispel I think the right solution is to hide inline toolbar with css properly when is not visible. I created new pull request https://github.com/draft-js-plugins/draft-js-plugins/pull/809

juliankrispel:
addressed in #809

Issue #469, opened 2016-10-03
Mentions: update to 'suggestions' prop while dropdown is closed and suggestions is empty doesn't trigger dropdown reopen


maxhr:
This can be reproduced in the Remote Data Mention Example:
https://www.draft-js-plugins.com/plugin/mention

Type `@m` fast, before remote fetch returns.
`setState` is called, `suggestions` prop is updated, but the dropdown opens only after `editorState` changes somehow.

This happens because `componentWillReceiveProps`of `MentionSuggestions` checks for empty suggestions and closes the dropdown:
https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/src/MentionSuggestions/index.js#L40

However triggering openDropdown happens only inside `onEditorStateChange` 
https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/src/MentionSuggestions/index.js#L84
which doesn't get called until some change happens inside the editor.

nikgraf:
Thanks for the detailed bug report. Would you be interested to submit a fix? I'm quite busy with other plugins right now.

tal-moshayov:
+1 any remedy?

jameskraus:
Is there any particular reason why the MentionSuggestions class closes itself if there are no suggestions? Can that just be disabled?

nikgraf:
Just merged this PR which should fix the issue: https://github.com/draft-js-plugins/draft-js-plugins/pull/659

PR #659, opened 2017-02-14
Reopens mentions dropdown if new suggestions are available.

jameskraus:
Here's an ultra simple fix for bug https://github.com/draft-js-plugins/draft-js-plugins/issues/469


nikgraf:
Looks good to me!

PR #662, opened 2017-02-18
Update to Draft-js v0.10.0

nil1511:
Updated Webpack
Updated Draft-js
Updated other user dependencies


nikgraf:
@nil1511 this is amazing! Do you have time fix all the linting issues and update the PR?

nil1511:
@nikgraf Sure I will fix the linting issue. There also an another issue with webpack-2 and react-static-webpack-plugin due to which Uglify is failing in build.

https://github.com/iansinnott/react-static-webpack-plugin/issues/37

nikgraf:
ohh no, maybe it's time to move away from it 🤔

nikgraf:
@nil1511 this is such a fantastic PR! Interested to become a core contributor to the project?

nil1511:
Sure.

nikgraf:
@nil1511 anything in particular you would like to tackle next?

nil1511:
Now custom entities types are in small case(eg. image) and some plugin uses [draft-js-utils](https://github.com/draft-js-utils/draft-js-utils) where entities are in uppercase I think we should have one standard.

Is there anything else you have in mind?

nikgraf:
good point, let's change it to all be uppercase. I saw you changed it in the serialized content state in the custom image example, but that broke it. Wanna clean that up?

My top priority list:
- Implement + Docs page for File upload Drag'n'Drop Plugin
- Docs page for Drag'n'Drop Plugin
- Make sure the new plugins work with Screenreaders (Aria support built in)
- Test on Mobile
- Update the homepage (add a video and full example)

Interested to pick any of those up or help out with issue and bug fixes?

nil1511:
Thank you for the list. I will update the entities type to use with draft-js-utils then will look at bugs.
will keep you updated when working on the list.

nikgraf:
thanks for helping 🙌

tedmx:
Real life saver since Draft is urgently pushing toward new API, thank you so much @nil1511

PR #663, opened 2017-02-18
Adding rel config

tagraha:
https://dev.to/ben/the-targetblank-vulnerability-by-example

### DIFF
--- a/draft-js-linkify-plugin/src/Link/index.js
+++ b/draft-js-linkify-plugin/src/Link/index.js
@@ -13,6 +13,7 @@ export default class Link extends Component {
       decoratedText = '',
       theme = {},
       target = '_self',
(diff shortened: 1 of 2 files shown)


nikgraf:
Good to know, thx! Glad that by default it uses `target="_self"` which means if people don't modify target they are still fine.

Would you mind updating the docs and the changelog for linkify?

- docs: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/docs/client/components/pages/Linkify/index.js#L90
- changelog: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-linkify-plugin/CHANGELOG.md

juliankrispel:
hey @tagraha, any chance you could make these updates? x

Issue #666, opened 2017-02-21
Mention - Cannot read property 'getBoundingClientRect' of null


ismyrnow:
When updating search suggestions, MentionSuggstionPortal loses its reference to `this.searchPortal`, which is a ref for the element.

<img width="740" alt="draft-js-mention-plugin-bug" src="https://cloud.githubusercontent.com/assets/1127410/23185694/4cf3631a-f852-11e6-90e1-52891d4e5f02.png">

The problem seems to have arose when switching from string refs (`ref='searchPortal'`) back in beta2. I have a fix for this and will open a PR shortly.

ismyrnow:
Could I have gotten a worse issue number?

ismyrnow:
The emoji plugin has code almost identical to this plugin which could be broken as well. I'm not using that plugin at the moment, so I don't know for sure if it's an issue there.

nikgraf:
hehe, bad luck with the number 😄

nikgraf:
Solved with #667

PR #667, opened 2017-02-21
Move searchPortal ref function to constructor

ismyrnow:
Fixes issue #666.

I'm not entirely sure why this makes a difference. I suspect it's a timing issue in React. Ideally, we shouldn't be accessing refs outside of the React lifecycle functions, so accessing this in a closure like this may cause problems again in the future if React internals change.


nikgraf:
I'm really confused that this would actually solve the bug. I mean it's exactly the same code except one time the function is attached to `this` and the other time it's an anonymous function.

Any idea why this would make a difference?

ismyrnow:
I'm confused about that too. I actually think it could be a bug with React,
however, because they don't recommend using refs outside of lifecycle
functions, I doubt opening an issue with them would go far.

PR #671, opened 2017-03-01
Support popoverComponent on MentionSuggestions

samdroid-apps:
This allows you to override the `<div />` used to render the mention
suggestions in.  A use case is to replace the `<div />` with MaterialUI's
`<Paper />` for design coherency.

### DIFF
--- a/docs/client/components/pages/Mention/index.js
+++ b/docs/client/components/pages/Mention/index.js
@@ -187,6 +187,13 @@ export default class App extends Component {
               <span className={styles.paramName}>onAddMention</span>
               <span>A callback which is triggered whenever the mention is about to be added. The first argument of this callback will contain the mention entry.</span>
             </div>
+            <div className={styles.param}>
+              <span className={styles.paramName}>popoverComponent</span>
+              <span>Component to be used as the template for the popover (the parent of entryComponent).  Defaults to a div.</span>
+            </div>
+            <div className={styles.param}>
+              <span>Additional properties are passed to the <InlineCode code="popoverComponent" /></span>
+            </div>
           </div>
           <Heading level={3}>Additional Exports</Heading>
           <div>
--- a/draft-js-mention-plugin/src/MentionSuggestions/__test__/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/__test__/index.js
@@ -39,37 +39,37 @@ const mentions = fromJS([
   },
 ]);
 
-describe('MentionSuggestions Component', () => {
-  it('Closes when suggestions is empty', () => {
-    const callbacks = {
+
+function defaultProps() {
+  return {
+    suggestions: mentions,
+    callbacks: {
       onDownArrow: sinon.spy(),
       onUpArrow: sinon.spy(),
       onTab: sinon.spy(),
       onEscape: sinon.spy(),
       handleReturn: sinon.spy()
-    };
-    const store = {
+    },
+    store: {
       getAllSearches: sinon.spy(() => ({ has: () => false })),
       getPortalClientRect: sinon.spy(),
       isEscaped: sinon.spy(),
       resetEscapedSearch: sinon.spy(),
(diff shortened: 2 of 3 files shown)


draft-js-mention-plugin/src/MentionSuggestions/__test__/index.js
```diff
@@ -80,4 +80,47 @@ describe('MentionSuggestions Component', () => {
     });
     expect(suggestions.state().isActive).to.equal(false);
   });
+
+  it('The popoverComponent prop changes the popover component', () => {
+    const PopoverComponent = ({ children, ...props }) => {
+      return <div data-test-test {...props}>{children}</div>;
+    };
```
samdroid-apps: eslint gives an odd error here:

    85:58  error  Unexpected block statement surrounding arrow body  arrow-body-style

IDK what's wrong with this... the code seems reasonably idiomatic to me.

nikgraf: The Airbnb setup expects you to do this:

```
const PopoverComponent = ({ children, ...props }) => (
  <div data-test-test {...props}>{children}</div>;
);
```

draft-js-mention-plugin/src/MentionSuggestions/index.js
```diff
@@ -317,31 +318,29 @@ export default class MentionSuggestions extends Component {
...
-        }
-      </div>
+    return React.cloneElement(
+      popoverComponent,
+      {
+        className: theme.mentionSuggestions,
+        role: 'listbox',
+        id: `mentions-list-${this.key}`,
+        ref: (element) => { this.popover = element; },
+        ...elementProps
```
nikgraf: You flipped the order here and spreaded the `elementProps` at the end. Was this in purpose? Which one of those would you want to overwrite?

docs/client/components/pages/Mention/index.js
```diff
@@ -187,6 +187,13 @@ export default class App extends Component {
               <span className={styles.paramName}>onAddMention</span>
               <span>A callback which is triggered whenever the mention is about to be added. The first argument of this callback will contain the mention entry.</span>
             </div>
+            <div className={styles.param}>
+              <span className={styles.paramName}>popoverComponent</span>
+              <span>Component to be used as the template for the popover (the parent of entryComponent).  Defaults to a div.</span>
+            </div>
+            <div className={styles.param}>
+              <span>Additional properties are passed to the <InlineCode code="popoverComponent" /></span>
```
nikgraf: If you provide a custom component do you even need custom props? Is there a use-case where we can't do without them?

samdroid-apps: Yes, we could probably remove the custom pops.  However, that would break api comapt - so I just wanted to make the smallest change possible.

draft-js-mention-plugin/src/MentionSuggestions/index.js
samdroid-apps: Not intentional - just a mistake.  Force pushed a fix.

docs/client/components/pages/Mention/index.js
nikgraf: What do you mean by breaking the api comapt? If you provide just the component wouldn't that be even a smaller change?

samdroid-apps: Some people might rely of passing extra params - it is currently part of the public api.

If we remove this feature, the users will need to go and change their code, even though it is quite trivial.  That would mean we are breaking the api compatibility.

samdroid-apps: @nikgraf, is this good to merge?

samdroid-apps:
Hey @nikgraf, just checking on the status of this patch.  Is there anything you would like me to do before it is merged?

nikgraf:
I was holding off, because I didn't get what it was doing at first. Looks good

Issue #660, opened 2017-02-15
Draft JS Plugin Editor and IE11: Expected Object error


tomkelsey:
Hi,

Using the Draft JS Plugin Editor I was getting an "Expected Object" error in IE11.

I fixed this by replacing every instance of a `(for x of y)` loop with a `forEach` loop in `/Editor/index.js`

As per: https://github.com/tomkelsey/draft-js-plugins-editor/blob/master/src/Editor/index.js

Due to how the project is structured I haven't had time to shape this into a PR but hopefully it's useful for anyone else that comes across this issue.

amannn:
I had the same problem. I'm using polyfill.io for polyfills but apparently the babel transpiled version of `for of` needs at least these two polyfills:

```
import 'core-js/es6/map';
import 'core-js/fn/symbol';
```

Using them instead of the ones from polyfill.io solved the problem for me.

nikgraf:
That's a pity!

@amannn would you mind opening a PR? We could add core-js as dependency. Correct?

amannn:
So you'd bundle those polyfills with the editor? Not sure if that's the best way. I could also add a note about this particular problem to the [FAQ](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/FAQ.md#the-editor-throws-errors-in-internet-explorer-11). Maybe there are other polyfills for Map and Symbol that work and if people use those, we'd unnecessarily increase their bundle size with additional polyfills. It could also be that the polyfill.io polyfills will be updated in the near future and will work properly. So personally I'd leave the polyfills out and try to solve the problem with documentation.

What do you think?

nikgraf:
Sound good to me, an update to the FAQ or even very early in the Readme sounds like a good approach …

tomkelsey:
Hi guys,

Just jumping back in and noticed the linting config extends airbnb's - I'm not entirely sure as to their rationale but they seem to recommend not using for-of loops:

https://github.com/airbnb/javascript#iterators-and-generators

I don't know how extensively they're used in the codebase but it could be worth removing them rather than adding a caveat. Either way, great job with these plugins @nikgraf 👍

nikgraf:
@tomkelsey we using them as a way to make the plugin system as performance as possible and bail out asap. Maybe that was over engineered …

amannn:
[...]

PR #696, opened 2017-03-24
Replaced `for of` loops with alternatives.

amannn:
As discussed in #660.

I've changed all the `for of` loops were possible to `forEach`. I noticed there were some loops where there was a return statement. As these can't be handled with `forEach`, I've changed them to basic `for` loops that iterate over a counter variable.

The file `src/Editor/index.js` seems to be the only file where `for of` loops occur – I've searched through the code base with a regex.

Currently the test "MentionSuggestions Component: The popoverComponent recieves the children" fails, but that seems to be unrelated to this PR.

### DIFF
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -126,27 +126,21 @@ class PluginEditor extends Component {
   createEventHooks = (methodName, plugins) => (...args) => {
     const newArgs = [].slice.apply(args);
     newArgs.push(this.getPluginMethods());
-    for (let i = 0; i < plugins.length; i++) {
-      const plugin = plugins[i];
-      if (typeof plugin[methodName] !== 'function') continue;
-      const result = plugin[methodName](...newArgs);
-      if (result === true) return true;
-    }
 
-    return false;
+    return plugins.some((plugin) =>
+      typeof plugin[methodName] === 'function'
+        && plugin[methodName](...newArgs) === true
+    );
   };
 
   createHandleHooks = (methodName, plugins) => (...args) => {
     const newArgs = [].slice.apply(args);
     newArgs.push(this.getPluginMethods());
-    for (let i = 0; i < plugins.length; i++) {
-      const plugin = plugins[i];
-      if (typeof plugin[methodName] !== 'function') continue;
-      const result = plugin[methodName](...newArgs);
-      if (result === 'handled') return 'handled';
-    }
 
-    return 'not-handled';
+    return plugins.some((plugin) =>
+      typeof plugin[methodName] === 'function'
+        && plugin[methodName](...newArgs) === 'handled'
+    ) ? 'handled' : 'not-handled';
   };
 
   createFnHooks = (methodName, plugins) => (...args) => {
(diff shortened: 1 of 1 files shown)


nikgraf:
@amannn thanks for putting in the time!

I would definitely merge the forEach looks, but the for loops seems to be step back even compared to the `for of` loops. Can you revert them or is there another way how to restructure the code there?

amannn:
Hmm, the requirement is that we need to return from inside the loop to the function the loop lies within.

I think we have the following options:

1. Use a `for of` loop. This leads to the problem described in #660 and causes a bug in IE11.
2. Use an array method like `Array.prototype.some` to do something like this:
```js
const wasHandled = plugins.some(plugin => {
  if (typeof plugin[methodName] !== 'function') continue;
  return plugin[methodName](...newArgs) === 'handled';
});
if (wasHandled) return 'handled';
```
3. Use a `for` loop that iterates over a counter variable as it's currently the case. Not too pretty, but does the job.
4. Use a [`for in` loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in). However the MDN reference says:
> Because the order of iteration is implementation-dependent, iterating over an array may not visit elements in a consistent order. Therefore it is better to use a for loop with a numeric index (or Array.prototype.forEach() or the for...of loop) when iterating over arrays where the order of access is important.

Personally I think that number 2 adds a little bit of indirection, but it's not too bad. If you like it, I can change the loops to that structure. Otherwise I think that number 3 isn't too bad given the requirements and options.

What do you think?

nikgraf:
Thanks for the detailed write-up! The benefit of `for of` was the early bail out if I remember correctly. By using some it would go over all of them again. In addition on MDN it's not clear how well it is supported in browsers like Safari. Even for Chrome the Yes is in brackets. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some?v=example

Let's stick to the counter. Can you add a comment above each counter loop to this PR so we have a reference to these options and why we chose a counter loop.
 Apart from that the PR is good to go :)

amannn:
Ok, I'll add the comments!

Just FYI: `some` can also bail out early, as it's enough when one function invocation returns true. E.g.  the code `[1,2,3, 4].some(i => { console.log(i); return i === 2; });` prints only `1` and `2` to the console. And I think that `(yes)` notation in MDN mean that the feature was supported since the first version. E.g. [parseInt](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Browser_compatibility) also has it.

nikgraf:
Woah, ok then let's go with `some`!

amannn:
@nikgraf Alright, the PR is now up to date with `some` loops! I double checked that this is a purely internal change to the functions so there shouldn't be any observable difference when using the editor.

nikgraf:
awesome @amannn! 🙌

nikgraf:
btw @amannn are you interested to contribute more in general? I'm quite busy and this project clearly needs more help from maintainers 😃

amannn:
Thanks for the offer – I feel honored 😀.

However, I don't know yet how much I'll work with Draft.js in the future, as the project I'm currently using it on is coming to an end. But if I stumble over other things in this repo that I think could be improved, I'll definitely open more PRs. Maybe then I'll reconsider devoting more time to this repo. Hope that's ok 🙂

nikgraf:
Works for me 🙌

PR #706, opened 2017-04-04
[mentions]: prevent selection commit on Tab/Enter if dropdown is not rendered

alexfedoseev:
I tried out mentions plugin and noticed that it inserts selected item on `Enter` or `Tab` when there's no trigger in sight and dropdown is not rendered. After reading the code, this is the first solution I came up with. Also, I was thinking about using `null` value for `focusedOptionIndex` when dropdown is not visible, but it's a bit implicit and I've already implemented the current fix, so I'd get feedback from maintainers first.

There's a chance that both directions are suck :) due to lack of my familiarity with this code base (and with draft js in general, yet). So I'll be happy to implement it in the right way after feedback from maintainers.

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/__test__/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/__test__/index.js
@@ -100,7 +100,7 @@ describe('MentionSuggestions Component', () => {
     let called = false;
     const PopoverComponent = ({ children, ...props }) => {
       called = true;
-      expect(React.Children.count(children)).to.equal(mentions.length);
+      expect(React.Children.count(children)).to.equal(mentions.size);
       return <div {...props}>{children}</div>;
     };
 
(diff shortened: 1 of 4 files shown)


mbseid:
This is important for our team as well. I'm going to see if I can help fixing those failing tests.

mbseid:
I fixed the tests and submitted a PR here: https://github.com/alexfedoseev/draft-js-plugins/pull/1

@alexfedoseev  if you merge it, the tests should pass and then the admins can merge it in. In the meantime I'm probably going to push a beta built to my organization NPM repo.

mbseid:
For anyone who needs this now, you can use our NPM package here: https://www.npmjs.com/package/@naytev/draft-js-mention-plugin

alexfedoseev:
@mbseid Thanks! Merged 👍 I was about to publish it to npm today, now I'll use yours. Thanks again!

nikgraf:
Looks good to me, Thanks for looking into it!

PR #707, opened 2017-04-05
fix 'addComponentAsRefTo...' bug in third party libs

ku8ar:
as in title...
[ref](https://github.com/draft-js-plugins/draft-js-plugins/pull/600)


nikgraf:
@ku8ar how does this help? in each package react is already a peerDependency!?!

PR #714, opened 2017-04-12
Fix the positioning of the alignment plugin to work the same way as the inline-toolbar works

Discordius:
We fixed the positioning of the inline toolbar a while ago to work well with relative parents, but the alignment plugin toolbar still had the same problem. So I changed the positioning logic of the alignment plugin to work the same way.

### DIFF
--- a/draft-js-alignment-plugin/src/AlignmentTool/index.js
+++ b/draft-js-alignment-plugin/src/AlignmentTool/index.js
@@ -12,6 +12,19 @@ import buttonStyles from '../buttonStyles.css';
 // TODO make toolbarHeight to be determined or a parameter
 const toolbarHeight = 44;
 
+const getRelativeParent = (element) => {
+  if (!element) {
+    return null;
+  }
+
+  const position = window.getComputedStyle(element).getPropertyValue('position');
+  if (position !== 'static') {
+    return element;
+  }
+
+  return getRelativeParent(element.parentElement);
+};
+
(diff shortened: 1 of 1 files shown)


Discordius:
Hmm, no idea why the mentions plugin check is failing. I didn't make any changes to any of the relevant code, I think.

nikgraf:
Thanks for looking into it @Discordius 👌

amannn:
If possible, it would be great to have a release for this in the near future as it solves a critical bug for me in IE11 👍

nikgraf:
@amannn I release beta11 about an hour ago :)

amannn:
Already saw it :). Thanks a lot!

HaNdTriX:
I know this PR is already merged. Just to clarify the [draft-js-side-toolbar-plugin](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-side-toolbar-plugin/src/components/Toolbar/index.js#L37-L50) could need this positioning technique as well.

Is there a WIP or are you open for PRs?

nikgraf:
@HaNdTriX no WIP or PR I'm aware of. Are you interested to contribute one?

HaNdTriX:
@nikgraf actually I was but now I switched from draft.js to slate. 

Nevertheless thanks a lot for the awesome work you guys did. 
I really enjoyed working with draft-js-plugins.

PR #716, opened 2017-04-13
Fix mentions

natebunnyfield:


nikgraf:
@natebunnyfield did this help to fix an issue?

natebunnyfield:
@nikgraf I believe so. We have since upgraded everything and are in the clear. Thanks for checking in though.

PR #719, opened 2017-04-24
Update packages

ngs:
Hi,

I've updated yarn package references because I could not run dev server with current master codebase.


nikgraf:
thanks @ngs

Issue #491, opened 2016-10-17
Cannot render stickers from raw content state 


timschambers:
Howdy,

First, thank you all for the great work on these set of plugins. They're really helping to reduce my pain points with draft-js.

Second,  I'm running into a problem where ContentStates that are "convertFromRaw" and stored as JS strings, can't render their stickers. Instead, any sticker blockType throws an 'Invariant' exception and does not show up in the editor. 

I am using draft-js-plugins-v2.0.0@beta5, draft-js-0.9.1, and React@15.3.2.

All stickers run into this problem, while other block types, like text, render perfectly fine.

I have attached screenshots of the debug output, particularly the line where it fails:

<img width="514" alt="screen shot 2016-10-17 at 9 29 20 am" src="https://cloud.githubusercontent.com/assets/1759595/19439630/d6756472-944c-11e6-8c19-57d2012b1e07.png">

The getEntityAt(0) returns a `null` that is passed into Entity.get, causing the error shown below.

<img width="1157" alt="screen shot 2016-10-17 at 9 34 34 am" src="https://cloud.githubusercontent.com/assets/1759595/19439684/077c1aca-944d-11e6-9d79-548edb5ced2f.png">

And this is JSON of a sample ContentState being loaded, which throws the 'Invariant' exceptions:

``` js
{
    "entityMap": {
        "0": {
            "type": "sticker",
            "mutability": "IMMUTABLE",
            "data": {
                "id": "/static/media/images.97b53447.jpeg"
            }
        },
        "1": {
            "type": "sticker",
            "mutability": "IMMUTABLE",
            "data": {
                "id": "/static/media/plague.2d0a9da3.gif"
            }
        }
    },
    "blocks": [{
        "key": "7m54a",
        "text": "",
        "type": "sticker",
        "depth": 0,
        "inlineStyleRanges": [],
        "entityRanges": [{
            "offset": 0,
            "length": 0,
            "key": 0
        }],
        "data": {}
    }, {
        "key": "fa7d4",
        "text": "",
        "type": "sticker",
        "depth": 0,
        "inlineStyleRanges": [],
        "entityRanges": [{
            "offset": 0,
            "length": 0,
            "key": 1
        }],
        "data": {}
    }, {
        "key": "f6o38",
        "text": "",
        "type": "unstyled",
        "depth": 0,
        "inlineStyleRanges": [],
        "entityRanges": [],
        "data": {}
    }]
}
```

[...]

PR #721, opened 2017-04-25
Update addSticker.js

matthiasn:
This small change fixes issue #491 for me, where my application would crash when a sticker was used and the content was recreated from raw.


matthiasn:
The failed test seems unrelated, it's the same one that fails on master.

nikgraf:
awesome @matthiasn - thanks for looking into it!

PR #725, opened 2017-05-04
Emoji select

bashkos:
Hi! I'm added a new component to `draft-js-emoji-plugin` EmojiSelect
gif

**I'm also**
* moved all components in `draft-js-emoji-plugin` to folder `components`;
* has expanded function `addEmoji`, so that it is possible not only replace text fragment but also insert emoji;
* added `sass-loader` because it's hard to control pure CSS when it contains many related rules
* separated suggestions and select props;
* supplemented the documentation.


bashkos:
Fix for tests https://github.com/draft-js-plugins/draft-js-plugins/pull/726

tomkelsey:
This looks awesome, great job 👍

nikgraf:
@bashkos this is suuuuper cool!

Can you also document the Mode option you introduced?

draft-js-emoji-plugin/src/index.js
```diff
@@ -92,18 +121,21 @@ export default (config = {}) => {
   // breaking change. 1px of an increased padding can break a whole layout.
   const {
     theme = defaultTheme,
-    positionSuggestions = defaultPositionSuggestions,
     imagePath = defaultImagePath,
     imageType = defaultImageType,
     allowImageCache,
-    priorityList,
+    positionSuggestions = defaultPositionSuggestions,
+    suggestionsPriorityList,
```
nikgraf: @bashkos can you change it back to `priorityList`. With these changes I like your naming more, but I think it's an unnecessary breaking change that would cause more confusion than it helps.

nikgraf:
@bashkos I just wanted to merge, but noticed I'm getting these two warnings in your branch. One seem to come from react-motion the other I don't know. Can you fix them?

screen shot 2017-05-08 at 23 35 25

bashkos:
@nikgraf these warnings from `react-scrollbar` package, that I used https://github.com/souhe/reactScrollbar/issues/85 . I do not know what to do with this, except use fork, because the owner does not react to issues and PR.

nikgraf:
@bashkos makes sense, hmmm - are there any similar projects we could use or even extract the minimal case we need into this repo?

The problem betting on an abandoned project is that we can't upgrade React at some point in the future and then often the only choice is to make breaking changes. I would like to avoid that.

bashkos:
@nikgraf I replaced the scrollbars component on a https://www.npmjs.com/package/react-custom-scrollbars. It is kept up-to-date.

.flowconfig
```diff
@@ -6,6 +6,7 @@
 .*/node_modules/.*.json
 .*/node_modules/draft-js/.*
 .*/node_modules/react/.*
+.*/node_modules/react-motion/.*
```
nikgraf: Can you remove this entry again?

bashkos: Sorry, my mistake

nikgraf:
awesome work @bashkos 👏

PR #730, opened 2017-05-09
fix: video plugin styles

echenley:
- Adds className prop to DefaultVideoComponent (needed by focus plugin).
- Enables customizing config.theme and allows focus plugin to work with videos.
- Move inline iframe styles to theme to allow for more customization.
- Removes unnecessary DefaultVideoComponent.css file.
- Removes dragNDropPlugin from example since it doesn't work.


nikgraf:
@mzbac can you have a look at this?

Antherkiv:
Hello the compiled video plugin (rc-02) in npm, has a problem, it try to find videoStyles.css, but it doesn't exist, thanks, great work!

Antherkiv:
.babelrc was missed, compile (not well) but not working at all, I added it #760

PR #732, opened 2017-05-10
Fix positioning of side toolbar in IE11.

amannn:
The positioning of the side toolbar was broken in IE11, as `window.scrollY` isn't available there. This fixes the issue.


nikgraf:
Thanks!

PR #737, opened 2017-05-11
Fix #702

### LINKED ISSUE #702: Draft-Js dependency incorrect?
Since 3ba2d3, I believe the draft-js dependency in each plugins' package.json file should have been bumped to v0.10.0. I just ran into a bunch of issues because the resizable plugin was pulling in an older version of draft js but calling the new getEntity function.

Related to this, it may make sense to move the draft-js dependency into peerDependencies in each of the plugins.

StorytellerCZ:
## Description

Upgrade `draft-js` dependency to 0.10.0 and move it to `peerDependencies` as per #702.

### DIFF
--- a/draft-js-alignment-plugin/package.json
+++ b/draft-js-alignment-plugin/package.json
@@ -31,7 +31,6 @@
   "license": "MIT",
   "dependencies": {
     "decorate-component-with-props": "^1.0.2",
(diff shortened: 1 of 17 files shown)


nikgraf:
Thanks @StorytellerCZ 🎉

Issue #718, opened 2017-04-24
UndoButton/RedoButton cannot be initialized before Editor


terryoy:
Hi,

I'm writing a toolbar with DraftJS, and it will be rendered before the Editor.

```jsx
<div>
<Toolbar><RedoButton/><UndoButton/></Toolbar>
<Editor .../>
</div>
```

I think it's because the RedoButton/UndoButton didn't consider the situation that the Editor has not been initialized before the buttons are rendered.

```
Uncaught TypeError: this.props.store.getEditorState is not a function
    at UndoButton.render (index.js:59)
    at ReactCompositeComponent.js:796
    at measureLifeCyclePerf (ReactCompositeComponent.js:75)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext 
```

Exception line: 
https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-undo-plugin/src/RedoButton/index.js#L21

When the method "getEditorState" is actually initialized:
https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-undo-plugin/src/index.js#L30

AverageZ:
+1

nikgraf:
> I think it's because the RedoButton/UndoButton didn't consider the situation that the Editor has not been initialized before the buttons are rendered.

You are right, interested to create a PR?

terryoy:
@nikgraf I have moved on to other projects and didn't maintain the work space, so I just added a simple fix without testing it. You can accept it if it's appropriate.

https://github.com/draft-js-plugins/draft-js-plugins/pull/740

nikgraf:
Ouch, just read you added the fix without testing it. Let's hope it works

hoppula:
It doesn't work yet as store always exists: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-undo-plugin/src/index.js#L14

I think https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-undo-plugin/src/RedoButton/index.js#L22 would need another check that ensures `this.props.store.getEditorState` actually exists.

terryoy:
Agree with @hoppula

Hongbo-Miao:
It also happens to LinkButton.

Hongbo-Miao:
This also happens to LinkButton.

minhdang241:
[...]

PR #740, opened 2017-05-17
Fix the Undo/Redo Button initialize issue

terryoy:
UNdo/Redo Button is disabled when => store is not initialized, or redo stack is empty.

## Implementation

Added an existence check of store to the Undo/Redo button

## Demo

https://github.com/draft-js-plugins/draft-js-plugins/issues/718


nikgraf:
Hmmm, I'm bit worried that this creates weird side-effects, but not sure right now. Let's see :)

nikgraf:
Thanks for taking the time to create a PR!

Issue #341, opened 2016-07-04
Styling issues due to the autocomplete dropdown not being appended to the body


Zhouzi:
While I am referring to the mention plugin in particular, this issue applies to other plugins that have an autocomplete dropdown. For now, the component is added to the portal, directly in its container element. It introduces a few styling issues:
- positioning the suggestions is a bit tricky (we managed to deal with it just fine but the current implementation is not perfect as it may be positioned out of the screen)
- part of the element may be hidden if its parent has an `overflow: hidden`

That's just two of the main issues but even from a theoretical point of view, those elements are "out of the flow", on top of the rest and depends on the viewport size. It'd make sense to move them as a direct child of the body. The good news is that those components are already instantiated through portals so it should be pretty straight forward.

**Update:** I've been looking into it and it's far from being trivial. I've found nice screenshots showing an example of the issues we can face when not appending the element to the body:

![suggestions overflowing](http://jamesknelson.com/wp-content/uploads/2015/03/Screen-Shot-2015-03-01-at-11.22.40-pm-800x232.png)

Instead of:

![correct behaviour](http://jamesknelson.com/wp-content/uploads/2015/03/Screen-Shot-2015-03-01-at-11.26.04-pm-800x220.png)

The screenshots are taken from an article explaining how to append a React component to the body: [Rendering React components to the document body](http://jamesknelson.com/rendering-react-components-to-the-document-body/). It's from March 2015 so the code is no more relevant but the point is still valid.

nikgraf:
Yep, this is quite tricky. Ideally we would implement smart absolute positioning like http://tether.io

I'm very opposed to render Elements outside of the the JSX where they are placed. This can cause way more trouble with other libs manipulating DOM and unfortunately there are a ton of libs doing a lot of crazy stuff to the DOM 😄 Happy to be convinced otherwise with a great PR, but just from experience I learned to reply with: "Unfortunately we didn't cover your case yet as we haven't had it ourselves. As a first resort I recommend to find a solution without `overflow: hidden`. A PR is very welcome, but currently not my focus."

Zhouzi:
Agreed, I think doing that would be optimal but in the end it's all a matter of if it's worth the effort or not. Right now, given the complexity and time it'd take, I believe it's not worth the value it would bring. So yeah, your reply seems to be the way to go.

That been said, if you're using plugins that blocks you from adding elements to the `<body>`, an `overflow: hidden;` is the last of your concerns 😄

PR #746, opened 2017-05-24
add react-portal to emojis and mentions

mattkrick:
## Implementation

In #341 you said you were open to a PR, so here's a minimal example of what is possible when we use a portal to break out of the DOM tree. 

## Demo

Nothing fancy to show, it looks the same!

## Use-case

As you can see in the example, even though the parent div is `position: relative` & `overflow: hidden`, the full menu still shows. This is the same strategy that facebook uses. If we don't make this change, then a designer could very easily bork the entire menu by simply changing the styles of a nearby component. A good summary of this issue (albeit with a different conclusion) is here: https://medium.com/@david.gilbertson/modals-in-react-f6c3ff9f4701. Ideally, we can switch over all plugins that produce a modal.

If closing animations, multiple menus, or more complex logic is required, we could switch to a more complex portal component that I made (https://github.com/mattkrick/react-portal-hoc), but as the name implies that is a higher-order component & I figured I'd start simple.

### NOTE BREAKING CHANGE FOR `popoverComponent`
The current implementation of `popoverComponent` accepts a component _instance_. 
This is troubling for a couple reasons:
- The instance _must_ be an `HTMLElement` (span, div, etc.)  
- The instance _must_ handle the props passed into it & any extra props in future versions will cause a breaking change since they may not be suitable for the element
- the instance _must_ handle `ref` appropriately, otherwise the modal won't appear
- The `ref` is may be stale & set to `null`. This makes portals impossible & was the main reason for fixing it

Reading the history of the feature, I see it's for using material-ui. Unfortunately, as it is currently written, this is about the only 3rd party package it is useful with because most others don't expect a `ref`.
The path to upgrade would look something like this:
```
const popoverComponent = ({popoverRef, children, closePortal, ...props}) => <Paper ref={props.popoverRef} {...props}>{children}</Paper>
```

With that small tweak, it'll work for everybody! (portals included)

### DIFF
--- a/docs/client/components/pages/Emoji/SimpleEmojiEditor/index.js
+++ b/docs/client/components/pages/Emoji/SimpleEmojiEditor/index.js
@@ -27,7 +27,7 @@ export default class SimpleEmojiEditor extends Component {
 
   render() {
     return (
-      <div style={{position: 'relative', overflow: 'hidden'}}>
+      <div style={{ position: 'relative', overflow: 'hidden' }}>
         <div className={editorStyles.editor} onClick={this.focus}>
           <Editor
             editorState={this.state.editorState}
--- a/docs/client/components/pages/Emoji/SimpleEmojiEditor/index.js
+++ b/docs/client/components/pages/Emoji/SimpleEmojiEditor/index.js
@@ -27,7 +27,7 @@ export default class SimpleEmojiEditor extends Component {
 
   render() {
     return (
-      <div>
+      <div style={{position: 'relative', overflow: 'hidden'}}>
         <div className={editorStyles.editor} onClick={this.focus}>
           <Editor
             editorState={this.state.editorState}
--- a/docs/client/components/pages/Mention/SimpleMentionEditor/index.js
+++ b/docs/client/components/pages/Mention/SimpleMentionEditor/index.js
@@ -38,7 +38,7 @@ export default class SimpleMentionEditor extends Component {
 
   render() {
     return (
-      <div className={editorStyles.editor} onClick={this.focus}>
+      <div className={editorStyles.editor} onClick={this.focus} style={{ position: 'relative', overflow: 'hidden' }}>
         <Editor
           editorState={this.state.editorState}
           onChange={this.onChange}
--- a/draft-js-emoji-plugin/package.json
+++ b/draft-js-emoji-plugin/package.json
@@ -35,11 +35,11 @@
     "emojione": "^2.2.7",
     "find-with-regex": "^1.0.2",
     "immutable": "~3.7.4",
-    "prop-types": "^15.5.8",
     "lodash.keys": "^4.2.0",
+    "prop-types": "^15.5.8",
     "react-custom-scrollbars": "^4.1.1",
-    "react-portal": "^3.1.0",
     "react-icons": "^2.2.3",
+    "react-portal": "^3.1.0",
     "to-style": "^1.3.3",
     "union-class-names": "^1.0.0"
   },
--- a/draft-js-emoji-plugin/package.json
+++ b/draft-js-emoji-plugin/package.json
@@ -38,6 +38,7 @@
     "prop-types": "^15.5.8",
     "lodash.keys": "^4.2.0",
     "react-custom-scrollbars": "^4.1.1",
+    "react-portal": "^3.1.0",
     "react-icons": "^2.2.3",
     "to-style": "^1.3.3",
     "union-class-names": "^1.0.0"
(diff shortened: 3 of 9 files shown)


nikgraf:
@mattkrick thanks for investing time and especially starting small! What do you mean by 

> As you can see in the example, even though the parent div is position: relative & overflow: hidden, the full menu still shows.

Can you do a screenshot what problem this solves?

mattkrick:
sure, i'll screenshot what it looks like if portals aren't used, 1 min

mattkrick:
bad

nikgraf:
Hmmm, just tested and for me it looks like this 

screen shot 2017-05-25 at 06 41 39

What browser are you using?

mattkrick:
heh, that's the point :smile: 

To make that screenshot I removed the `Portal` locally & snapped the shot.

If you're on master, you'll need to add `<div style={{ position: 'relative', overflow: 'hidden' }}>` around the Editor.

Either of those will cause the break.

nikgraf:
Ok, so if someone has `position: 'relative', overflow: 'hidden' ` then it wouldn't show up. But why would you want to add these styles if you don't want overflow hidden anyway? Sorry, seems like I'm missing something here :(

mattkrick:
ah! alrighty lemme use a real world example:

I have 2 lanes of todo items.
Each todo item is a card.
Each card should stay in its lane.
So, I use overflow: hidden on the column.

However, the mention menu is a _child_ of one of those cards and a grandchild of the column, and as a  grandchild, it has to obey the rules of its parents (including overflow!).

So, the result is something like this:
screenshot from 2017-05-24 22-32-21

That means if we don't use portals, we have to be _very very careful_ that a parent div does not use `overflow: hidden`. 

Another example might be using this inside a modal.
We want all the contents to stay _inside_ the modal, but it's OK if the menu goes outside.
Lemme know if that makes sense! If not I'll try to come up with another example.

nikgraf:
Ahhh ok, yeah that makes sense. Since the MentionsSuggestion is a Component you could put it into a Portal yourself

```jsx
  render() {
    return (
      <div className={editorStyles.editor} onClick={this.focus}>
        <Editor
          onChange={this.onChange}
          plugins={plugins}
          ref={(element) => { this.editor = element; }}
        />
        <Portal>
          <MentionSuggestions
            onSearchChange={this.onSearchChange}
            suggestions={this.state.suggestions}
            onAddMention={this.onAddMention}
          />
        </Portal>
      </div>
    );
  }
```

The benefit here would be that we don't force developers to use a portal. Having something breaking out of the DOM might cause in various setups. Still we should probably recommend it in the docs! In addition we have one dependency less.

What do you think? btw thanks a lot for investing the time to explain it to me.

mattkrick:
that _could_ work, but I would advise against it for 2 reasons:
- if i want a plugin, i just want it to work :smile: that's the beauty of the plugin ecosystem that you created!
- whether the portal is opened or closed is a function of the internal `isActive` state, which is a child of the Portal. So the example above doesn't work because I'd need to write something like `<Portal isOpened={this.state.isActive}>`. Since parents can't manage the state of their children, I'd have no way to manage that state. 

Now, I _could_ just leave the portal open all the time, but that means I open up a portal for every single plugin on every single draftjs instance. for example, i might have 30 draft-js instances with 3 portal-driven plugins. that'd be 90 extra DOM nodes! that doesn't scale well.

nikgraf:
Makes sense. Let's wait with this change for a bit to avoid confusion and we can do a breaking change when Fiber comes out (which supports Portal out of the box … see https://github.com/facebook/react/blob/master/src/renderers/dom/fiber/__tests__/ReactDOMFiber-test.js#L272)

mattkrick:
Does that mean this package would have a hard peer dependency on React v16? I'm OK with that, but I know the road to v16 is a long one, so it may take some folks awhile to make the migration.

nikgraf:
Yeah, as soon as React 16 is out I would like to upgrade so we don't block people from upgrading. I hope React 16 is coming sooner than later. Facebook is using it internally already.

jimyaghi:
Is there any more progress on this? I'm using the plugin in a overflow: auto (scrolling) container and i wish the mentions menu would show on top of the container's lower boundary rather than inside the draft editor and cut by the outer container.

image

comatory:
@jimyaghi I'm on react 15.4.2 and to solve this issue I used `react-portal` on version 3.1.0. It works well for most of it (I had to wrap `MentionSuggestions` component in another `<span>` so I could add `className` prop on it to manipulate `z-index` property.

It's not super elegant but it works. I'm kind of 50/50 on this. On one hand, you're bound to bump into this problem sooner or later with any kind of pop-over, on the other hand I can see how authors do not want to add this feature in. Thankfully I was able to solve this with `react-portal`.

I can see two options here:

- add configuration prop to render mentions/dropdowns outside of parent and utilize native `React.createPortal`. So limiting this feature for newer version of React is OK with me (since there is a solution for older versions)
- do not add any configuration but add a solution to this problem to documentation

PR #762, opened 2017-06-04
onMentionFocus wasn't properly setting state using `setState`

bradbirnbaum:
## Implementation
Using the latest versions of React and DraftJs when using the up/down arrow keys it wasn't properly selecting the appropriate items


nikgraf:
Thanks for the fix @bradbirnbaum 🙌

Issue #773, opened 2017-06-14
Limit the amount of suggestions


davidsonsns:
In mentions, limit the amount of suggestions

nikgraf:
Hi @davidsonsns, for this use-case I recommend you to pass in your own filter function that limits the suggestions for you. I hope that helps.

PR #774, opened 2017-06-14
Limit the amount of suggestions in mentions

davidsonsns:
## Implementation
Adjust for issue #773, limit the amount of suggestions in mentions

## Demo
```
  onSearchChange = ({ value }) => {
    this.setState({
      suggestions: defaultSuggestionsFilter(value, mentions, 3),
    });
  };
```

## Use-case
For cases where we need more or less than 5 listed suggestions


nikgraf:
Hi @davidsonsns, for this use-case I recommend you to pass in your own filter function that limits the suggestions for you. I hope that helps.

PR #782, opened 2017-06-28
Add a default rel property; limit props to component to valid attributes

bmsterling:
React was throwing the unknown prop warning, this fix is to address that but limiting the props that can be added to a link.

Also addresses the `target="_blank"` vulnerability.


bmsterling:
Just noticing that there was already a PR for part of what this PR fixes (#663) Let me know if you'd like me to break out just the prop changes.

PR #783, opened 2017-06-29
Update dependencies for linkify-plugin

sugarshin:
## Implementation

Update `linkify-it` to `^2.0.3` and `tlds` to `^1.189.0` for linkify-plugin.

## Demo

Nothing.

## Use-case

Nothing.


nikgraf:
Thanks @sugarshin

Issue #593, opened 2016-12-09
webpack bundle has two copies of immutable


lonelyclick:
I used source-map-explorer look at the webpack bundle, found that there are two copies of the immutable. One from draft-js to ~3.7.1, the other from the ^3.8.1 in the draft-js-plugins, so we can make the two library unified version of it?

nikgraf:
good point, interested to open a PR to change the dependencies to 3.8.1

lonelyclick:
#598

J22Melody:
Cool, same issue. Wonder when and which npm version will this commit be published so I can use it via npm?

danii1:
Still different versions of immutable.js....

AleksandrZhukov:
via Webpack you can do:
```
  resolve: {
    alias: {
      ...
      immutable: path.join(__dirname, '../node_modules/immutable')
    }
  }
```

danii1:
I'm using Yarn, so I ended up adding 
```javascript
"resolutions": {
    "draft-js/immutable": "^3.8",
    "draft-js-plugins-editor/immutable": "^3.8",
    "draft-js-export-html/immutable": "^3.8"
}
```
to package.json

PR #794, opened 2017-07-12
Allow immutable 3.8.x as well

tarjei:
Please tell if the `~` is intended :)

## Implementation
Allow not just immuatable 3.7.4 but alos 3.8.x etc

### DIFF
--- a/draft-js-plugins-editor/package.json
+++ b/draft-js-plugins-editor/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "decorate-component-with-props": "^1.0.2",
     "find-with-regex": "^1.0.2",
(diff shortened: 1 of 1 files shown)


nikgraf:
Thanks @tarjei

whazap:
just noticed immutable being imported twice in some project;
immutablejs versions:
* 3.7.4
* 3.8.2
i think this commit is causing that.
draft-js has immutable "~3.7.4" as a dependency, (https://github.com/facebook/draft-js/blob/master/package.json#L37)
so does every draft-js-plugin, except for draft-js-plugins-editor, which uses "^3.7.4".
related: #593

PR #796, opened 2017-07-13
Fix initialization of blockRenderMap to address problem with Maps ins…

richtera:
…ide of blockRenderMap.

## Implementation

The problem with a plain maps.merge is that immutable will completely wrap the item into a Map object. Then when the draft-js editor gets the blockRender item from the blockRenderMap it will receive a Map object rather than the expected plain javascript object.

## Demo

This is hard to demo since nothing shows when it's broken and the custom renderMap item works when it's fixed. Basically it will look for Map.wrapper to have a value inside of draft-js to determine whether it needs to do custom rendering and it just doesn't.

## Use-case

Creating a plugin with a blockRenderMap will work after this patch. Before the renderMap was being ignored.

### DIFF
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -272,7 +272,12 @@ class PluginEditor extends Component {
   resolveblockRenderMap = () => {
     let blockRenderMap = this.props.plugins
       .filter((plug) => plug.blockRenderMap !== undefined)
-      .reduce((maps, plug) => maps.merge(plug.blockRenderMap), Map({}));
+      .reduce((maps, plug) => {
+        Object.keys(plug.blockRenderMap).forEach((key) => {
+          maps = maps.set(key, plug.blockRenderMap[key]);
+        });
+        return maps;
+      }, Map({}));
     if (this.props.defaultBlockRenderMap) {
       blockRenderMap = DefaultDraftBlockRenderMap.merge(blockRenderMap);
     }


richtera:
Checking the test scripts but they might be wrong.... NOTE: This is how draft-js currently utilizes the blockRenderMap:
```javascript
	        decorator: decorator,
	        direction: direction,
	        forceSelection: forceSelection,
	        key: key,
	        offsetKey: offsetKey,
	        selection: selection,
	        tree: editorState.getBlockTree(key)
	      };

	      var configForType = blockRenderMap.get(blockType);
	      var wrapperTemplate = configForType.wrapper;
              // <<<<<<==== NOTE here it doesn't expect a Map object to come back. It expects
              // <<<<<<==== configForType to be a plain javascript object.

	      var _Element = configForType.element || blockRenderMap.get('unstyled').element;

	      var depth = _block.getDepth();
	      var className = this.props.blockStyleFn(_block);

	      // List items are special snowflakes, since we handle nesting and
	      // counters manually.
	      if (_Element === 'li') {
	        var shouldResetCount = lastWrapperTemplate !== wrapperTemplate || currentDepth === null || depth > currentDepth;
```

richtera:
There is another way to fix this problem. The documentation should show that it's required to create the blockRenderMap as a immutable map when using it as part of a plugin. This will also forgo the conversion to a full Map of each child object. I noticed this is how the test scripts were doing this and this works.

PR #801, opened 2017-07-16
fix video plugin build

teux:
Lacks some configuration in video plugin build. 
As a result, css is missing in the package and exception occurs.

I'm copy .babelrc from image plugin.
Please update npm package draft-js-video-plugin


nikgraf:
Thanks for looking into it! Just merged it and going to cut a release soon

PR #804, opened 2017-07-20
Fix #639 & #763 : Alignment plugin , deleting the block still shows the toolbar

### LINKED ISSUE #639: AlignmentToolbar is not hidden when focused item is deleted
The issue can be observed in the `Alignment + Resize + Focus + Drag&Drop Example` sample on the `Image` plugin page (https://www.draft-js-plugins.com/plugin/image).

1. Click on the image to provide it focus. The alignment toolbar will appear.
2. Press the `Backspace` or `Delete` key on the keyboard. The image will be deleted, but the AlignmentToolbar element will remain.

The AlignmentToolbar will remain visible until another image (or other element that supports the AlignmentToolbar) is focused and unfocused.

### LINKED ISSUE #763: Alignment plugin: deleting the picture still shows the toolbar
If I go to the demo https://www.draft-js-plugins.com/plugin/alignment and delete the image, the toolbar still displays the alignment toolbar, as if the image still exists.

image

How to reproduce:

1. Go to https://www.draft-js-plugins.com/plugin/alignment
2.  Scroll down to the example
3. Delete the image using the backspace key or the delete key
4. The alignment toolbar still exists. It should not, since the image no longer exists.

ramrami:
## Implementation

Fix #639 & #763 : deleting the block still shows the toolbar
ezgif com-video-to-gif 2

## Demo
ezgif com-video-to-gif

### DIFF
--- a/draft-js-alignment-plugin/src/createDecorator.js
+++ b/draft-js-alignment-plugin/src/createDecorator.js
@@ -28,6 +28,11 @@ export default ({ store }) => (WrappedComponent) => class BlockAlignmentDecorato
     }
   }
 
+  componentWillUnmount() {
+    // Set visibleBlock to null if the block is deleted
+    store.updateItem('visibleBlock', null);
+  }
+
   render() {
     const {
       blockProps,


nikgraf:
Thanks for looking into it. Love such simple PR that fix problems like this ✨❤️

PR #807, opened 2017-07-28
Fix #633 mentions dropdown props update

### LINKED ISSUE #633: Once suggestions prop contains an empty list mentions dropdown closes and doesn't reopen
The example for remote data uses a JSON file, however, when the data is not cached (and thus near instantaneous lookup) then the suggestions box will not display reliably (more often than not, it does not display). For example, this issue appeared when I switched from a JSON file to Algolia for my mention lookups.

You can easily recreate this by either putting the JSON data for the remote example on to a service like Algolia and loading it that way, or you can use a setTimeout/lodash.delay for every call and set it to around 100-200ms.

juliankrispel:
fixes #633


nikgraf:
@juliankrispel can you also add that to the plugin changelog?

juliankrispel:
@nikgraf coming up!

PR #809, opened 2017-07-28
hide inline toolbar if is not visible using css visibility

MilanLempera:
- it takes toolbar buttons out of the tab order, when are hidden

## Implementation
Hide toolbar with visibility property, browser then ignore buttons whenpage is walked throught tab. 

## Demo
same as https://github.com/draft-js-plugins/draft-js-plugins/pull/647


juliankrispel:
@MilanLempera thanks for submitting this, I'll test it and get back to you!

juliankrispel:
thanks @MilanLempera much appreciated!

PR #817, opened 2017-08-02
Hide inline toolbar on editor blur

Kornil:
## Implementation

- Fixed inline toolbar not disappearing when editor loses focus (blur) [Video of the bug](https://drive.google.com/file/d/0B4McFs23KX_VdnFsbmc2c0hDX1E/view?usp=sharing)
- Fixed test for inline toolbar

## Demo

[Demo of the fix](https://drive.google.com/file/d/0B4McFs23KX_VdWlNU1N5LVpMTGc/view?usp=sharing)

## Use-case

//

### DIFF
--- a/draft-js-inline-toolbar-plugin/src/components/Toolbar/__test__/Toolbar.js
+++ b/draft-js-inline-toolbar-plugin/src/components/Toolbar/__test__/Toolbar.js
@@ -28,7 +28,7 @@ describe('Toolbar', () => {
       unsubscribeFromItem() {},
       getItem: (name) => ({
         getEditorState: () => ({
-          getSelection: () => ({ isCollapsed: () => true })
+          getSelection: () => ({ isCollapsed: () => true, getHasFocus: () => true })
         })
       }[name])
     };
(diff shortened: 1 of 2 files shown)


juliankrispel:
@Kornil can you me instructions to reproduce this? Ideally [a codesandbox](https://codesandbox.io/) I can't reproduce it it seems to work fine for me.

Kornil:
Hi @juliankrispel, thanks for your reply, I created a [codesandbox](https://codesandbox.io/s/yrxR60A1P?autoresize=1&hidenavigation=1&module=%2FHello.js&moduleview=1) to show the bug, steps to replicate:

- focus on the edit
- write something random
- highlight the text you wrote to show the inline toolbar
- click on the `<p>` "some random text" or in any other place outside the editor but still on the page.

What should happen is that the inline toolbar attaches the element just above it, in this case the `p` tag, if you remove that element it attaches to the `h1` above.

juliankrispel:
thanks @Kornil weird that we can't reproduce it on the draft js homepage, that said in firefox we can reproduce it o_O

juliankrispel:
Just realised that this breaks the newly added link plugin since the overrideContent is an input for adding a link...

juliankrispel:
so I'll have to change it to
```
const isVisible = (!selection.isCollapsed() && selection.getHasFocus()) || overrideContent;
```

that's more appropriate I suppose. Lmk if you have any thoughts on that cc @nikgraf

PR #729, opened 2017-05-09
Add feature to inline toolbar that allows for children to override the content.

amannn:
## Implementation

This allows buttons of the inline toolbar to render dynamic content into the toolbar. An example is the ability to add links directly from the toolbar (see demo).

This could be a first step for a robust plugin to add links. Currently we have #610, but that touches quite a few files and got released in a [separate repository](https://github.com/vacenz/last-draft-js-plugins) anyway. This PR is inspired by the implementation of the toolbar plugin over there, but is agnostic to what kind of content can be rendered in the toolbar.

## Demo

draft

In this case, I've got a `AddLinksButton` component in the `structure` for the toolbar. When the button is clicked, it calls `this.props.onOverrideContent(AddLinkForm);` to render the form. The form can use the same method for reverting back to the normal state.

## Use-case

I needed this to implement adding links directly from the toolbar.

## Allow editors for maintainers

- [x] Enable "Allow edits from maintainers" for this PR

## TODO
- [x] Add a test
- [x] Add docs

PR #819, opened 2017-08-03
Add link plugin

amannn:
## Implementation

As discussed in PR #729 here comes the link plugin 👍. There are quite a few files in the PR, but most of them are docs.


stevewillard:
You rock! This is awesome ❤️ 🎉

draft-js-link-plugin/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "prepublish": "npm run build"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "decorate-component-with-props": "^1.0.2",
+    "immutable": "~3.7.4",
+    "prepend-http": "1.0.4",
+    "prop-types": "^15.5.8",
+    "union-class-names": "^1.0.0",
+    "url-regex": "4.1.1"
```
amannn: The failing test complains about this dependency missing. Not sure why though. I noticed the package.json in the root folder has a lot of dependencies that are specific to plugins. Should I add this dependency to that package.json as well?

amannn: I've added them to the main package.json now. Looks like on CI only those dependencies are installed.

clwilliams:
this is exactly what I'm looking for, thank you! How long does it usually take for these requests to be ok'd?

jrock17:
Agreed. Would love to see this merged!

draft-js-link-plugin/src/components/Link/index.js
```diff
@@ -0,0 +1,34 @@
...
+  className: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  entityData: PropTypes.object,
+  entityKey: PropTypes.string,
+  getEditorState: PropTypes.func
+};
+
+const Link = ({ children, entityData, entityKey, getEditorState, className }) => {
+  if (!entityData) {
+    if (!getEditorState) return <span />;
```
juliankrispel: what's this check for? Surely getEditorState should always be defined

draft-js-link-plugin/src/components/Link/index.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+  const href = (entityData && entityData.url) || '';
+
+  return (
+    <a
+      className={className}
+      title={href}
+      href={href}
+      target="_blank"
+      rel="noopener noreferrer"
```
juliankrispel: nice :+1:

draft-js-link-plugin/src/components/Link/index.js
```diff
@@ -0,0 +1,34 @@
...
+    entityData = entity ? entity.get('data') : undefined;
+  }
+  const href = (entityData && entityData.url) || '';
+
+  return (
+    <a
+      className={className}
+      title={href}
+      href={href}
+      target="_blank"
```
juliankrispel: shouldn't be the default. Maybe make it an option?

draft-js-link-plugin/package.json
nikgraf: yes, for caching reasons the CI installs only from the root package. Eventually using yarn workspaces or similar would be great. Something for post 2.0

juliankrispel:
also @amannn if you could run yarn and include the updated lock file

draft-js-link-plugin/src/components/Link/index.js
amannn: Yep, that's true. There were some left over checks that I used in a similar component in order to work better with [draft-convert](https://github.com/HubSpot/draft-convert/). But they're actually not necessary here and only add noise. I've removed them now.

amannn:
Good points, thanks for the review! The last commit should fix the mentioned issues.

juliankrispel:
Awesome dude. Sick pr! 🙌

Julian Krispel
UI Engineer & Consultant
http://reactrocket.com
+447988116178

nikgraf:
This is so rad! @juliankrispel any objections left or should we merge?

juliankrispel:
@nikgraf mergio! Can you release it as part of rc4? Since that plugin hasn't been published yet we should get away with it right?

jrock17:
Would love to see this merged. Great work!

juliankrispel:
Oh it'll get merged shortly 

Julian Krispel
UI Engineer & Consultant
http://reactrocket.com
+447988116178

juliankrispel:
So, on second thought, I'm actually thinking it'd be best to treat the Link Button like any other Button and add docs for [draft-js-buttons](https://github.com/draft-js-plugins/draft-js-plugins/tree/master/draft-js-buttons). I appreciate the work you've already put in and so I'd be happy to take over from here. Whatcha think @amannn @nikgraf <3

juliankrispel:
Actually I just figured that we of course have tons of components like that, for example - undo/redo, alignment etc. I'll try and figure out a solution going forward but for now let's just merge :+1:

juliankrispel:
so I'd like to release the link plugin but by the looks of it that name is already taken by one of @StevenIseki's [excellent plugins](https://www.npmjs.com/package/draft-js-link-plugin)

@amannn wanna suggest an alternative name? How about `draft-js-anchor-plugin`? `draft-js-links-plugin`, idk 😳  cc @nikgraf

jrock17:
Did you guys pick and name and publish to NPM for this? Thanks!

StorytellerCZ:
Maybe we ought to add some common namespace to the name when the plugin concerns the toolbar.

amannn:
Actually I implemented the plugin as a draft-js-button initially. But there were three things that made me move the code to a standalone plugin:

- It has two external dependencies. They’re not huge, but I thought its helpful to not bloat the package size of the other buttons.
- There’s also a need for a decorator which can render the created links. These could also be two separate packages, but I thought it might be easier to use if it’s included in a single one.
- The styles for the draft-js-buttons live inside the inline toolbar. Therefore you’d have to create the styles for the link button and the links in the content there and also offer configurations for that options there.

I’m open to change though if it makes more sense. What do you think @juliankrispel? 

I like the idea from @StorytellerCZ. Maybe draft-js-inline-toolbar-link-plugin is a good option, as it’s intended to be used alongside the inline toolbar.

juliankrispel:
@amannn so I'm thinking for now let's leave it as is. But here's what I'm thinking for 2.0

1. The reason I don't want to namespace it as a toolbar-link-plugin is that it'd be useful as a standalone thing as well (after a little refactoring)
2. For the 2.0 release I'm thinking about making Buttons official building blocks of the draft-js-plugins repo. They're quite powerful and it'd be nice to enable users to use them in whatever way they see fit. Right now they're kind of hidden and don't have any docs. 

Also, for now I've renamed the plugin to `draft-js-anchor-plugin`, there's [a pr here](https://github.com/draft-js-plugins/draft-js-plugins/pull/838) lmk if you're cool with the change or would like something else...

amannn:
@juliankrispel Alright, sounds good to me! :)

juliankrispel:
Beautiful, I'll release Friday along with @mxstbr's changes 🙌

Julian Krispel
UI Engineer & Consultant
http://reactrocket.com
+447988116178

PR #821, opened 2017-08-08
Fix hashtag as part of url bug

verbaleks:
## Implementation
When you use hashtag plugin it finds hashtags at url, what is wrong. Example: https://minimalistbaker.com/vegan-roasted-red-pepper-pasta-gf/#_a5y_p=2099867

## Demo

wrong
correct


juliankrispel:
@verbaleks it's the order that's causing the issue. If you have your plugins array like this: `[linkifyPlugin, hashTagPlugin]` it should work. The first plugin wins

juliankrispel:
We're addressing this in #824, @nikgraf is making a pr right now for the docs

Issue #781, opened 2017-06-28
EmojiSelect button has default type attribute


dglinyanov:
When placing EmojiSelector inside form, click on emoji selector causes form submit at least in Chrome (and maybe other browsers).

dglinyanov:
By the way emoji inside EmojiSelect Popover have same behavior - they are implemented as buttons and trigger form submit.

juliankrispel:
oh o_O

PR #829, opened 2017-08-13
Add type="button" to emoji select to prevent submitting when placed inside a form.

alanwflood:
Fix for issue #781 

By default buttons inside nested forms when clicked submit the form. Adding type="button" fixes the issue.


dglinyanov:
@alanwflood, if you had time to create PR, please, add one more line here:

https://github.com/alanwflood/draft-js-plugins/blob/7118bb402980cf9b2387198d64f18c4485b38adc/draft-js-emoji-plugin/src/components/EmojiSelect/Popover/Entry/index.js#L64

Otherwise you will be able to open Popover, but click on emoji image will trigger submit.

juliankrispel:
@alanwflood great catch! I'll make the update @dglinyanov

PR #830, opened 2017-08-14
Allow passing extra data with addImage call

mxstbr:
We need to keep a reference to the actual file object around in our app, which is currently impossible as we have to use `URL.createObjectURL` to get a URL suitable to be passed as a `src`.

This patch allows us to keep that reference in `entity.data` like so:

```JS
onChange(addImage(state, URL.createObjectURL(file), { file: file }));
```

Without this we can't upload the images to a remote server (in our case S3) since we can't do that with just the URL. By allowing extra data to be passed through to the entity this becomes possible.

### DIFF
--- a/draft-js-image-plugin/src/modifiers/addImage.js
+++ b/draft-js-image-plugin/src/modifiers/addImage.js
@@ -3,10 +3,10 @@ import {
   AtomicBlockUtils,
 } from 'draft-js';
 
-export default (editorState, url) => {
+export default (editorState, url, extraData) => {
   const urlType = 'image';
   const contentState = editorState.getCurrentContent();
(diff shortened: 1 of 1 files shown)


juliankrispel:
actually the docs for the image plugin are so bad that they need a general overhaul anyway, so I'll defer adding the docs and merge this in.

mxstbr:
Oh sorry, totally missed your comment, thanks Julian! Would appreciate a release in the meantime 😊

juliankrispel:
Working on it 🙌

It'll probably be Friday when I have a check-in with nik 👌

Julian Krispel
UI Engineer & Consultant
http://reactrocket.com
+447988116178

PR #831, opened 2017-08-14
Add missing plugins to list

ddelrio1986:
## Implementation

Added some plugins that were missing in the list of plugins shown in the README.


nikgraf:
Oh, I had in mind to create a section mentioning the new ones in 2.0. Nevermind this works as well :) Thanks @ddelrio1986

PR #835, opened 2017-08-15
Draft js static toolbar plugin

juliankrispel:
Not sure if this is a great name, maybe it should just be Toolbar ? 🙄 

## Implementation

I mostly took the inline-toolbar plugin as a starting point here and changed it. Simples. I also made some changes to enable rendering buttons before the Editor.

## Demo
toolbar-plugin

## Use-case

I need a standard wysiwyg toolbar, not inline or anything like that. So here it is.


draft-js-inline-toolbar-plugin/src/index.js
```diff
@@ -36,6 +36,7 @@ export default (config = {}) => {
 
   return {
     initialize: ({ getEditorState, setEditorState }) => {
+      console.log('init inline');
```
nikgraf: should be removed

draft-js-buttons/src/utils/createInlineStyleButton.js
```diff
@@ -18,7 +19,11 @@ export default ({ style, children }) => (
 
     preventBubblingUp = (event) => { event.preventDefault(); }
 
-    styleIsActive = () => this.props.getEditorState().getCurrentInlineStyle().has(style);
+    // we check if this.props.getEditorstate is undefined first in case the button is rendered before the editor
+    styleIsActive = () => {
+      console.log('does the editor store have the current style?', this.props.getEditorState && this.props.getEditorState().getCurrentInlineStyle().toJS());
```
nikgraf: should be removed

juliankrispel:
@nikgraf addressed all the feedback and removed the editorstate caching. plz review and let's merge and release :+1:

PR #837, opened 2017-08-15
Fixes Inline toolbar overflow on left and wrap on right

FrankieFabuloso:
## Fixed Issue

screen shot 2017-08-15 at 12 36 11 pm

When a user selects text that is on the far left the toolbox overflows off the left side of the window.

screen shot 2017-08-15 at 12 37 28 pm
When a user selects text that is on the right of the toolbox buttons wrap and stack.


antoinerousseau:
why not merged?

PR #846, opened 2017-08-19
[draft-js-anchor-plugin] PropTypes from 'prop-types' instead of React

stevewillard:
Noticed I got a warning in my project. I haven't tested it (I just made the change through the Github UI), but it matches the other component in this plugin.


stevewillard:
Ah, I assumed it was already updated for the other components. Let's close this, and I'll try to carve out some time to migrate all of them.

PR #850, opened 2017-08-24
blockStyleFn should return '' instead of false.

kgdev:
Other with the DOM will have <div class="false" ...


nikgraf:
Thanks for investigating the issue @kgdev!

Can you also update the changelog file https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-plugins-editor/CHANGELOG.md and add a test for this?

juliankrispel:
@kgdev mind responding to this? <3

PR #852, opened 2017-08-30
Sidebar position if there is a relative parent

mbutsykin:
## Implementation

New param added.
`createSideToolbarPlugin({relative: true})` to calculate position relatively to editor element

## Demo

So for now parent element can have position relative
image
```
createSideToolbarPlugin({
  theme: { buttonStyles, toolbarStyles, blockTypeSelectStyles },
  relative: true
})
```

### DIFF
--- a/docs/client/components/pages/SideToolbar/CustomSideToolbarEditor/editorStyles.css
+++ b/docs/client/components/pages/SideToolbar/CustomSideToolbarEditor/editorStyles.css
@@ -7,6 +7,7 @@
   margin-bottom: 2em;
   box-shadow: inset 0px 1px 8px -3px #ABABAB;
   background: #fefefe;
+  position: relative;
 }
 
 .editor :global(.public-DraftEditor-content) {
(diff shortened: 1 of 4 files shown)


mbutsykin:
@nikgraf please take a look at this

juliankrispel:
@mbutsykin can you elaborate what problem this actually fixes?

mbutsykin:
@juliankrispel sure.

when there is a parent with relative position in DOM, sidebar positioning is wrong.
`top` property counting relatively to scroll position, which obviously will be broken with relative parents.

i added a new flag - `relative` instead of searching like in this [one](https://github.com/draft-js-plugins/draft-js-plugins/pull/629) for performance issues

mbutsykin:
you can reproduce an issue simply adding `postion: relative` to `.editor___3zB6j` element.
in https://www.draft-js-plugins.com/plugin/side-toolbar

juliankrispel:
@mbutsykin just discussed this with @nikgraf earlier this week, we'd rather not add this config option and do it automatically. Is that something you feel up to doing?

mbutsykin:
ok, i'll update PR today or on this weekend, thanks for the feedback

mbutsykin:
@juliankrispel so, will you take a look?

Discordius:
Just mentioning that I am also interested in this PR getting merged.

CentaurWarchief:
Any update?

juliankrispel:
@Discordius @CentaurWarchief @mbutsykin you can actually work around this with just css with a simple transform:

```CSS
transform: 'translate(-100px,-50px) !important;'
```

Or by replacing the Toolbar component with your own implementation.

As discussed above we'd not like to add any more configuration options to this plugin and the workaround seems rather trivial

Issue #298, opened 2016-06-08
Multiple Mention editors conflict


tonypee:
I have a component which creates a mention simple mention component. It is standalone, and works fine.

When i iterate a list, and create multiple (i would be linking them to a list of rows) they conflict. When i @mention in the top componetn, the suggestions from the bottom component are displayed. then on arrow keys it disappears, never to show again. The only component to work in the last in the list (the corrosponding input to the suggestions that always show)

When i input after it is broken, i get the error: Uncaught TypeError: Cannot read property 'getIn' of undefined 

from 'MentionSuggestions' at the line: return editorState.getBlockTree(blockKey).getIn([decoratorKey, 'leaves', leafKey]);

Have you had issues with multiple components on the same page? (i realize that the example page has 2, but i dont see any options for id'ing them etc)

nikgraf:
Make sure to get a new plugin instance for every editor. More details here: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/FAQ.md#can-i-use-the-same-plugin-for-multiple-plugin-editors

flgarrid:
I'm getting the following error when I create a new instance of the plugin:

`React.createElement: type should not be null, undefined, boolean, or number. It should be a string (for DOM elements) or a ReactClass (for composite components).`

It happens when I call the second suggestion: 
`const { EmojiSuggestions2  } = emojiPlugin2;`

Are you sure it is possible to create two instances?

flgarrid:
Ok so it seems that it's how Babel transpiles the definition of const.

If I declare the second variable as:

`const EmojiSuggestions2 = emojiPlugin2.EmojiSuggestions;`

It works fine.

nikgraf:
@flgarrid did you manage to get it work?

btw this is how JavaScript deconstruction works:

```JS
const emojiPlugin = {
  test: 2,
}

// works
const { test } = emojiPlugin;

// can't find test2
const { test2 } = emojiPlugin;
```

Does that help?

flgarrid:
@nikgraf Yup I solved it with my latest post.

The FAQ it has the following code:

```
const emojiPlugin = createEmojiPlugin();
const emojiPlugin2 = createEmojiPlugin();
const { EmojiSuggestions } = emojiPlugin;
const { EmojiSuggestions2 } = emojiPlugin2;
```

Which by what you described, won't work. 
Instead, it should be:

```
const emojiPlugin = createEmojiPlugin();
[...]

PR #869, opened 2017-09-12
Update mention documentation to assign plugin instance to property

jakegavin:
## Implementation

As discussed in https://github.com/draft-js-plugins/draft-js-plugins/issues/298, a few people have  accidentally used the same `createMentionPlugin()` plugin for multiple editor components. This happens because they expect the [examples in the docs](https://www.draft-js-plugins.com/plugin/mention) to work, even if there are multiple instances of the editor component on the same page. While this is [mentioned in the FAQ](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/FAQ.md#can-i-use-the-same-plugin-for-multiple-plugin-editors), it would be helpful to mention also this in the documentation.

What are the disadvantages to always assigning the plugin instance to a property on the component? If you don't want to suggest this for the typical use case, I can change this PR to add another example for "Multiple Editors". 

## Use-case

A common use case for draft js is comment inputs. In a news-feed-like UI, there are often multiple commentable things on the page and thus you end up with multiple draft js inputs. If you follow the current guides that define `const mentionPlugin = createMentionPlugin();` outside of the component class, reusing your mention-enabled component will result in the error described in https://github.com/draft-js-plugins/draft-js-plugins/issues/298 because every component instance will be using the same `mentionPlugin` instance.

A more robust solution would be to log a warning if multiple components are using the same `mentionPlugin` instance. But for now, I think a documentation update would help new users of the project. 

### DIFF
--- a/docs/client/components/pages/Mention/CustomComponentMentionEditor/index.js
+++ b/docs/client/components/pages/Mention/CustomComponentMentionEditor/index.js
@@ -7,6 +7,23 @@ import mentions from './mentions';
 
 export default class CustomMentionEditor extends Component {
 
+  constructor(props) {
+    super(props);
+
+    this.mentionPlugin = createMentionPlugin({
+      mentions,
+      mentionComponent: (mentionProps) => (
+        <span
+          className={mentionProps.className}
+          // eslint-disable-next-line no-alert
+          onClick={() => alert('Clicked on the Mention!')}
+        >
+          {mentionProps.children}
+        </span>
+      ),
+    });
+  }
+
   state = {
     editorState: EditorState.createEmpty(),
     suggestions: mentions,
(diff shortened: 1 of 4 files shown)


nikgraf:
Just discussed with @juliankrispel that this might confuse more than it helps. Not everyone has instance properties available in their Babel setup.

We are thinking that in a 3.0 or so createPlugin could create a function that instantiates the plugin and in the editor the plugin is instantiated. Not sure though.

An alternative could be that we add a constructor for each example. I think that makes things better. Thoughts?

jakegavin:
I opted for instance properties because these examples already use them for initial state and event handlers. But I agree that since class fields are still [stage 3](https://github.com/tc39/proposal-class-fields), constructors might be more newcomer friendly and I have updated this PR.

Regarding changes to createPlugin in a 3.0, that sounds interesting but I dont think I fully understand since I'm not familiar with the editor internals. I still think a documentation along the lines of this PR could help in the interim.


Thanks for the work you all have put into this project! It's a very helpful library.

PR #870, opened 2017-09-12
Add support for missing latin characters

thomas88:
## Implementation

- Adding support for Latin-1 Supplement and Latin Extended-A characters to the `defaultRegExp` in mentions plugin to allow searching European names.
- I've also replaced the `\w` with `a-zA-Z0-9-` (support `-` instead of `_`)
- Further I've split the regex into multiple strings and ordered them roughly by occurance in the unicode table.

Let me know if you'd like me to change anything.

## Demo

Add names like Anna Glöckner, María-Jose Quiñones, László Horvát or Linnéa Månsson to mentions.js in the simple editor example. You'll only be able to search them with the extended regex.

### DIFF
--- a/draft-js-mention-plugin/src/defaultRegExp.js
+++ b/draft-js-mention-plugin/src/defaultRegExp.js
@@ -1,8 +1,30 @@
-// common chinese symbols: \u4e00-\u9eff - http://stackoverflow.com/a/1366113/837709
-// hiragana (japanese): \u3040-\u309F - https://gist.github.com/ryanmcgrath/982242#file-japaneseregex-js
-// katakana (japanese): \u30A0-\u30FF - https://gist.github.com/ryanmcgrath/982242#file-japaneseregex-js
-// For an advanced explaination about Hangul see https://github.com/draft-js-plugins/draft-js-plugins/pull/480#issuecomment-254055437
-// Hangul Syllables (korean): \uAC00-\uD7A3 - https://en.wikipedia.org/wiki/Korean_language_and_computers#Hangul_in_Unicode
-// Hangul Jamo (korean): \u3130-\u318F - https://en.wikipedia.org/wiki/Korean_language_and_computers#Hangul_in_Unicode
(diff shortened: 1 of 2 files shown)


draft-js-mention-plugin/src/defaultRegExp.js
```diff
@@ -1,8 +1,30 @@
-// common chinese symbols: \u4e00-\u9eff - http://stackoverflow.com/a/1366113/837709
-// hiragana (japanese): \u3040-\u309F - https://gist.github.com/ryanmcgrath/982242#file-japaneseregex-js
-// katakana (japanese): \u30A0-\u30FF - https://gist.github.com/ryanmcgrath/982242#file-japaneseregex-js
-// For an advanced explaination about Hangul see https://github.com/draft-js-plugins/draft-js-plugins/pull/480#issuecomment-254055437
-// Hangul Syllables (korean): \uAC00-\uD7A3 - https://en.wikipedia.org/wiki/Korean_language_and_computers#Hangul_in_Unicode
-// Hangul Jamo (korean): \u3130-\u318F - https://en.wikipedia.org/wiki/Korean_language_and_computers#Hangul_in_Unicode
-// Cyrillic symbols: \u0410-\u044F - https://en.wikipedia.org/wiki/Cyrillic_script_in_Unicode
-export default '[\\w\u4e00-\u9eff\u3040-\u309F\u30A0-\u30FF\uAC00-\uD7A3\u3130-\u318F\u0410-\u044F]*';
+export default '[' +
+  'a-zA-Z0-9-' +
```
nikgraf: Any specific reason why to support `-` instead of `_`? Why not support both?

thomas88: My reasoning was that `-` is used in e.g. Spanish names, but `_` is not used in real names. 

No harm in supporting both though and enables usernames with underscore. I'll add the changes.

nikgraf: great thanks … makes sense for names. Nowadays this plugin is use by people for way more. I think supporting both out of the box makes sense to avoid confusion

nikgraf:
@thomas88 can you resolve the conflict in the changelog?

Issue #857, opened 2017-09-01
Mentions plugin should replace all text between @ and last character of mention


thomas88:
See this screen capture:
![mentions](https://user-images.githubusercontent.com/1851359/29961998-beccf226-8f01-11e7-88b0-70f630c66b87.gif)

Something seems to be off with the `defaultSuggestionsFilter`, as it probably shouldn't return any suggestions for " i i i i i".

And then the plugin should replace all the text, starting with the @ character to the last character of the mention.

I haven't dug into the internals of the plugin yet, but it seems to me that these two issues might be related.

dovydaskukalis:
Plugin seems to work much better if I comment out [this line](https://github.com/draft-js-plugins/draft-js-plugins/blob/e5ad5da34b6c23e0178d33edc8a1c752398aaab5/draft-js-mention-plugin/src/MentionSuggestions/index.js#L45)

Issue #825, opened 2017-08-11
Bug with mentions plugin, hitting enter at end of text


juliankrispel:
To reproduce:

1. Run docs locally
2. focus on frontpage editor
3. Hit enter a couple of times
4. See mentions being inserted

![mentions-bug](https://user-images.githubusercontent.com/1188186/29209302-bcbfd2fa-7e85-11e7-9f33-ca7611f637a0.gif)

Issue #856, opened 2017-09-01
Pressing arrow key up/down with mention/emoji plugin should move the cursor


thomas88:
First of all thank you for the amazing, high quality plugins! Great work 👍 

Currently when pressing arrow key up/down in the mention/emoji plugin examples, the cursor does not move up/down, as you can see in this screen capture:

![mentions](https://user-images.githubusercontent.com/1851359/29960622-b4e730e6-8efc-11e7-893e-c5857010d9c6.gif)

Any hints on what's happening here?

leekevinyg:
@thomas88 were you able to find a workaround for this issue?

PR #871, opened 2017-09-12
Fix incorrect opening of suggestions

thomas88:
## Implementation

We've fixed async loading of suggestions (the issue here https://github.com/draft-js-plugins/draft-js-plugins/issues/633) by re-opening the dropdown when the `MentionSuggestions` component receives non-empty suggestions (see https://github.com/draft-js-plugins/draft-js-plugins/pull/807/files). 

This incorrectly opens the dropdown also when the component is re-rendered, but suggestions have not changed.

I'm adding a check that the suggestions have actually changed.

## Demo

Run the docs locally, focus the editor on the front page and hit enter a couple of times or try to at-mention "@ i i i i". Both issues are solved.

## Use-case

Fixes https://github.com/draft-js-plugins/draft-js-plugins/issues/825
Fixes https://github.com/draft-js-plugins/draft-js-plugins/issues/857

### DIFF
--- a/draft-js-mention-plugin/CHANGELOG.md
+++ b/draft-js-mention-plugin/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Fixed incorrect opening of suggestions. Thanks to @thomas88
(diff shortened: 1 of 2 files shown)


thomas88:
This seems to also fix https://github.com/draft-js-plugins/draft-js-plugins/issues/856, though I don't really understand how.

nikgraf:
@thomas88 great work, thanks! ✨👏

PR #872, opened 2017-09-12
Add loader module name extension

thomas88:
https://webpack.js.org/guides/migrating/#automatic-loader-module-name-extension-removed


nikgraf:
thanks @thomas88 for helping out here to keep the repo up to date!

PR #876, opened 2017-09-13
Remove Sass from repo

juliankrispel:
We have sass in the repo because of one stylesheet for the emoji plugin ... unnecessary ...


nikgraf:
suuuper cool!

PR #887, opened 2017-09-23
Add ability to use custom buttons and decorator styling in alignment plugin like inline-toolbar

hyperparsa:
## Implementation

The config argument of createAlignmentPlugin now accepts two children:
- structure : an array containing buttons to be on the toolbar (from draft-js-buttons)
- decoratorRenderStyling : a function that takes in `(alignment, {style, className})` and outputs props to be used on the decorated block.

## Screenshot

image
image

## Demo

I have added a new story to the storybook to show this new feature.

## Use-case

There should be away to customize the rendering of this plugin.

### DIFF
--- a/draft-js-alignment-plugin/CHANGELOG.md
+++ b/draft-js-alignment-plugin/CHANGELOG.md
@@ -5,4 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## To Be Released
 
+### Added
+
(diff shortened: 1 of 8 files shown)


juliankrispel:
@hyperparsa sweeeeeet! WE HAVE OUR FIRST PR USING STORIES WOOOHOOOO!

Mind adding to the changelog and a test case and then we should be good 👍

hyperparsa:
@juliankrispel Haha yeah
Can you elaborate on the tests? I don't see any tests for the alignment plugin to add a test case.

juliankrispel:
all merged!

PR #894, opened 2017-09-25
Fix mention regex for characters which need escaping!!!!

### LINKED ISSUE #881: Mention Plugin - Problem with mentionTrigger
When I make the `mentionTrigger` a parenthesis `(` or bracket `[`, like this:

```js
const mentionPlugin = createMentionPlugin({
  mentionTrigger: '(',
});
```

I get this error:

```
Uncaught SyntaxError: Invalid regular expression: /(/: Unterminated group
    at RegExp (<anonymous>)
```

This only happens if the first thing that I type into the editor is `(`. If I type some other characters first, it works fine.

Is there a way to achieve the same thing with `mentionRegExp`?

juliankrispel:
## Checklist

## Use-case/Problem

fixes #881

## Implementation

- add story with custom mention trigger which needs escaping (for testing)
- update regexp escaping (use String.raw) to properly escape regex strings...

## Demo

BOOOYAAAAH

mention-regex-fix

### DIFF
--- a/draft-js-mention-plugin/CHANGELOG.md
+++ b/draft-js-mention-plugin/CHANGELOG.md
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Escape mention trigger regex properly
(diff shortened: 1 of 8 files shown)


agarrharr:
Looks good. Does it work if the parenthesis is the first character you type?

juliankrispel:
mention-regex-fix-2

@agarrharr yes indeed

PR #928, opened 2017-10-25
Move cursor after image when inserting

mxstbr:
Right now, when you insert an image the cursor is afterwards _above_ the image, rather than below it. With this fix the cursor is correctly moved after the image, which makes for a much nicer writing experience.


juliankrispel:
@mxstbr thanks I'll have a look at this now

Issue #926, opened 2017-10-21
[drag-n-drop-plugin] store.getReadOnly is not a function


mxstbr:
We have an editor built with the `draft-js-drag-n-drop-plugin` and [our error monitoring service is reporting a constant stream of this error](https://sentry.io/share/issue/6eb3f33493de4ba3a4b3bc28b9b1c9ca/):

<img width="758" alt="screen shot 2017-10-21 at 6 52 55 pm" src="https://user-images.githubusercontent.com/7525670/31853899-15434702-b691-11e7-9904-a8fcb6aaae7e.png">

Looking at the drag-n-drop-plugin, this is happening because `store.getReadOnly` is first defined as `undefined`, and only assigned a proper function in `initialize`:

https://github.com/draft-js-plugins/draft-js-plugins/blob/d6182c5a8e029d3e4acba23e69cd277db94856b6/draft-js-drag-n-drop-plugin/src/index.js#L4-L16

Looking at `createDecorator.js`, it calls `store.getReadOnly` in the `render` method but that is throwing an error for some of our users:

https://github.com/draft-js-plugins/draft-js-plugins/blob/d6182c5a8e029d3e4acba23e69cd277db94856b6/draft-js-drag-n-drop-plugin/src/createDecorator.js#L23-L24

So for some reason `render` is called here before `initialize` of the main plugin is called. How can that happen? I can submit a patch to introduce a check to `createDecorator` (`store && store.getReadOnly`), but is that the right fix or is there something bigger going on?

mxstbr:
This is consistently happening for hundreds of our users. :confused: Any hunches what the issue could be? Otherwise I'll have to hotfix this and just default to `readOnly=true` if `store` is undefined or something.

juliankrispel:
hotfixed by @mxstbr in #929

mxstbr:
I think this should stay open since it might be indicative of a larger problem, surely the drag-n-drop plugin won't be the only one affected by a render happening before `initialize` is called?

juliankrispel:
@mxstbr no, unfortunately this problem is consistent throughout the project... because we're managing state inside draft-js-plugins... and so it's working against the react render cycle...

elie222:
Yeah. The same problem just came up for me using the image plugin.

elie222:
For anyone else that runs into this problem, I was creating the plugins in the render function. When I added an image to my editor I was updating the state which caused the editor to rerender and the plugins to be recreated.

[...]

PR #929, opened 2017-10-25
Fix draft-js-drag-n-drop plugin

mxstbr:
While I still don't know why or how this can happen, this should at least make it so the breaking #926 bug doesn't happen anymore.


juliankrispel:
@mxstbr thanks :+1:

mxstbr:
Can we ship this?!

juliankrispel:
@mxstbr fix is published - try version `2.0.0-rc9` of the dnd plugin

PR #932, opened 2017-10-26
onEmojiFocus wasn't properly setting state using `setState`

hbeckeri:
## Use-case/Problem
 Same issue as https://github.com/draft-js-plugins/draft-js-plugins/pull/762 but for the emoji suggestions.


hbeckeri:
@juliankrispel have you had a chance to look at this?

juliankrispel:
@hbeckeri looks legit, thanks, I'll add a changelog

PR #936, opened 2017-10-28
Changing src to iframe embeded src

Gasheck:
## Checklist

## Use-case/Problem

When you are exporting a contentState to HTML using, for example, https://www.npmjs.com/package/draft-js-export-html , there is no source link to video iframe inside entityData. So we cannot display it while parsing a contentState
- For example what we have inside entityData - https://www.youtube.com/watch?v=c1cPBPyuJ9g
- What we are expecting to have - https://www.youtube.com/embed/c1cPBPyuJ9g

## Implementation

Moving the code, that converts video link to iframe link, from:
`components/DefaultVideoComponent.js`
to:
`modifiers/addVideo.js`
And injecting iframe link to entityData at once

### DIFF
--- a/draft-js-video-plugin/src/video/components/DefaultVideoComponent.js
+++ b/draft-js-video-plugin/src/video/components/DefaultVideoComponent.js
@@ -1,27 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import utils from '../utils';
-
-const YOUTUBE_PREFIX = 'https://www.youtube.com/embed/';
-const VIMEO_PREFIX = 'https://player.vimeo.com/video/';
-
-const getSrc = ({ src }) => {
-  const {
-    isYoutube,
-    getYoutubeSrc,
-    isVimeo,
-    getVimeoSrc,
-  } = utils;
-  if (isYoutube(src)) {
-    const { srcID } = getYoutubeSrc(src);
-    return `${YOUTUBE_PREFIX}${srcID}`;
-  }
-  if (isVimeo(src)) {
-    const { srcID } = getVimeoSrc(src);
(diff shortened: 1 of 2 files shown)


juliankrispel:
@Gasheck hi thanks for contributing.

> there is no source link to video iframe inside entityData. So we cannot display it while parsing a contentState

I'm not quite sure what you mean @Gasheck. Can you elaborate? You want to display it while parsing it? From looking at the code it looks like you're simply using another endpoint for emnbedding youtube videos...

Gasheck:
@juliankrispel I'm trying to explain.

**How it works now:**
When we we want to put a video inside the editor, we type the link like https://www.youtube.com/watch?v=c1cPBPyuJ9g and the video plugin makes the link to put it inside the iframe https://www.youtube.com/embed/c1cPBPyuJ9g
Actually, it changes the resource name:
makes `/embed/c1cPBPyuJ9g` instead `/watch?v=c1cPBPyuJ9g`
Also, the plugin saves the 'watch' link `/watch?v=c1cPBPyuJ9g` to the entityData of the contentState

**And now let's suppose we want to parse the contentState to html**:
Let's use this module for example https://www.npmjs.com/package/draft-js-export-html
The problem is that we have again to make the 'embed' link `/embed/c1cPBPyuJ9g` to put it inside the iframe because the entityData stores the 'watch' one `/watch?v=c1cPBPyuJ9g`

juliankrispel:
Ok, I've tested this thoroughly now and read the code over again.

1. The video plugin does not change the url at all. If you'e using `embed` it'll use that, if you're using `watch` it'll use that.
2. I don't see any pressing reason to change how this works and am therefore closing this pr. If you object, please comment and we can have a discussion in this thread or on slack.

Thanks so much for contributing!

Issue #925, opened 2017-10-21
[Bug] Heart emoji ":heart:"


F3LENYR:
Hey,

Thank you for this great emoji plugin.

Package : "draft-js-emoji-plugin"

I've found a bug with the emoji "heart", when I type ": heart :", when I press enter or click on it, the emoticon is not displayed :  
https://gyazo.com/72f4e6dad0da098b9a14d58f69a73cbf

(Sorry for my bad english because I'm french)

Thank you in advance.

steven-qi:
I have opened a PR #938 for this.

juliankrispel:
Hi @steven-qi. Thanks for opening the issue and contributing. However I'm struggling to reproduce this, it's working for me in the sandbox:

![heart-emoji](https://user-images.githubusercontent.com/1188186/32490693-f55bbf64-c3ab-11e7-8908-ccb6a6e74d17.gif)

@steven-qi can you reproduce this in a [codesandbox](codesandbox.io) please?

steven-qi:
@juliankrispel Please check out the comments of the updated PR. I can reproduce it directly on demo page here with Windows 10 + Chrome 62.

https://www.draft-js-plugins.com/plugin/emoji

BTW, I am not the one who opened the issue, but I do encounter the same issue few days ago when I was trying to incorporate this to my own project..

juliankrispel:
fixed

PR #938, opened 2017-10-31
Fix code of heart emoji

steven-qi:
## Checklist

## Use-case/Problem

Emoji `:heart:` in priority list won't be displayed correctly (see #925 ).

## Implementation

Fix the incorrect Unicode code of `:heart:`.

## Demo

image

### DIFF
--- a/draft-js-emoji-plugin/CHANGELOG.md
+++ b/draft-js-emoji-plugin/CHANGELOG.md
@@ -12,7 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Added type="button" to EmojiSelect button to prevent submitting forms, thanks to @alanwflood [for the pr](https://github.com/draft-js-plugins/draft-js-plugins/pull/829)
-- Fixed incorrect unicode of emoji 'heart'. By @steven-qi
+- Fixed rendering of emoji 'heart' on Windows 10 + Chrome. By @steven-qi
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


steven-qi:
@nikgraf I saw that you are the original author of this file. Could you please review this change, as I don't know whether the incorrect Unicode was made intentionally or not.

juliankrispel:
Hi @steven-qi I've addressed your issue #925 - I can't reproduce the bug I'm afraid.

There are two unicodes here, the first one means red-heart, the second is just heart (with default color). I'm not sure of the original intentions but it seems to work fine for me.

`['2764-fe0f', '2764']`

steven-qi:
@juliankrispel I guess this could be a platfom-specific issue, as I am using Windows 10 + Chrome. I can easily reproduce it on the demo page.

https://www.draft-js-plugins.com/plugin/emoji

image

I do realized that simply replace it with 2764 will cause a black heart be rendered, and I am using `:heartbeat: 1f493` instead in priority list of my own project.

steven-qi:
BTW, `2764-fe0f` doesn't seems to be one unicode but two.

Here is what I found from emojione repo regarding how `-` is handled...

https://github.com/emojione/emojione/blob/master/lib/js/emojione.js#L409

I guess that this PR can be closed now, and a true fix is needed.

steven-qi:
I might have figured out the root cause of this issue.

The svg file for [2764](https://cdn.jsdelivr.net/emojione/assets/svg/2764.svg?v=2.2.7) missing width/height, thus won't be rendered as background. Specifying a `width: 1em` would fix it.

I will update this PR soon.

juliankrispel:
merged finally thank you so much ❤️

steven-qi:
Thanks @juliankrispel for merging. It is great that we get the broken ❤️ fixed. 😄

juliankrispel:
❤️👌😂

Issue #948, opened 2017-11-04
Mention plugin. Add props to change selection by default


KuznetsovIlya:
Hello, we try to implement mention plugin in our editor and found some use-cases, which can't be realized with current `MentionSuggestions` component's interface. We use draft-js-mention-plugin without @ trigger, with enabling searche after user inputs a few symbols, so we expect to have an ability to disable default selection for suggestion list.

PR #949, opened 2017-11-04
Add props to change selection and commit by default behavior

KuznetsovIlya:
## Checklist

## Use-case/Problem
#948 

## Implementation
Adding props to change initial selection index and props for forbid commit by default. Also add mouseLeave handler for Entry to remove selection when mouse leave popover


juliankrispel:
@KuznetsovIlya thanks for opening the issue and contributing.

However the mentions plugin is already very feature full and hence I'm hesitant to add anything at this point.

To me it sounds like in your case it's probably best to fork the plugin I'm afraid.

I'm working on experimenting with changes for 3.0 which provides more flexibility overall, if you fancy adding to the discussion, here's [the doc](https://paper.dropbox.com/doc/Architecture-changes-3.0-dLkTLrMlDq53VjvesyyxH)

KuznetsovIlya:
Ok, thanks for reply!

PR #951, opened 2017-11-06
Fix wait a tick to properly set overrideContent in inline plugin

slobodanl:

## Checklist

- [X ] Fix any eslint errors that occur
- [] Add/Update tests if you add/change new functionality
- [] Add/Update docs if you add/change functionality

## Use-case/Problem

There is current bug visible in Custom Inline Toolbar Example
https://www.draft-js-plugins.com/plugin/inline-toolbar
Clicking on Headline button instead of replacing the inline toolbar content as intended it closes the toolbar as Headline button was a normal button.

## Implementation

This bugfix adds delay just enough so the Headline button, in this case, to set overrideContent so the Inline toolbar will remain visible awaiting for further user action based on overrideContent rendition.

### DIFF
--- a/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
+++ b/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
@@ -68,24 +68,28 @@ export default class Toolbar extends React.Component {
   };
 
   getStyle() {
-    const { store } = this.props;
-    const { overrideContent, position } = this.state;
-    const selection = store.getItem('getEditorState')().getSelection();
-    // overrideContent could for example contain a text input, hence we always show overrideContent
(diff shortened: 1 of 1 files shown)


juliankrispel:
Hi @slobodanl thanks for contributing!!

I'm afraid you haven't filled out the pr template. Filling it out will tell us more about what you're trying to do with this pr. What's the bug you're fixing etc

slobodanl:
Hi @juliankrispel,

I hope now the pr template is complete. 

Thanks and sorry for the inconvenience I was too lazy :S

Best Regards,
Sl

juliankrispel:
oh hey - closing this since it's not really going to work (getStyle doesn't return anything here...)

juliankrispel:
fixed in #1060

Issue #952, opened 2017-11-06
Image delete keep entity into state 


kalimantos:
When deleting an image, the entity remains binded to the block above into the entityRanges and the entity is not removed from the entityMap.
https://codesandbox.io/s/4ql8746pl7
Step to reproduce:
- delete the image
- log the state in console with the proper button

state before:
```json
{
  "entityMap": {
    "0": {
      "type": "image",
      "mutability": "IMMUTABLE",
      "data": {
        "src": "https://www.draft-js-plugins.com/images/canada-landscape-small.jpg"
      }
    }
  },
  "blocks": [
    {
      "key": "9gm3s",
      "text": "You can have images in your text field. This is a very rudimentary example, but you can enhance the image plugin with resizing, focus or alignment plugins.",
      "type": "unstyled",
      "depth": 0,
      "inlineStyleRanges": [],
      "entityRanges": [],
      "data": {}
    },
    {
      "key": "ov7r",
      "text": " ",
      "type": "atomic",
      "depth": 0,
      "inlineStyleRanges": [],
      "entityRanges": [
        {
          "offset": 0,
          "length": 1,
          "key": 0
        }
      ],
      "data": {}
    },
    {
      "key": "e23a8",
      "text": "See advanced examples further down …",
      "type": "unstyled",
      "depth": 0,
      "inlineStyleRanges": [],
      "entityRanges": [],
      "data": {}
    }
  ]
}
```
state after:
```json
{
  "entityMap": {
    "0": {
      "type": "image",
      "mutability": "IMMUTABLE",
      "data": {
        "src": "https://www.draft-js-plugins.com/images/canada-landscape-small.jpg"
      }
    }
  },
  "blocks": [
    {
      "key": "9gm3s",
      "text": "You can have images in your text field. This is a very rudimentary example, but you can enhance the image plugin with resizing, focus or alignment plugins. See advanced examples further down …",
      "type": "unstyled",
      "depth": 0,
      "inlineStyleRanges": [],
      "entityRanges": [
        {
          "offset": 155,
          "length": 1,
          "key": 0
        }
      ],
      "data": {}
    }
  ]
}
```
i noticed that in the [draft-js media example](https://github.com/facebook/draft-js/blob/master/examples/draft-0-10-0/media/media.html) this didn't happen, but for now I've not been able to solve the problem

kalimantos:
I found a solution but it doesn't solve the problem at all.
Adding
```
[...]

PR #953, opened 2017-11-07
fix atomic block deletion

kalimantos:
fixes issue #952 
initial state:
initial_state

result after deleting the block:
- before:
before

- after:
after

As this bug occurs for every atomic block i placed the principal fix in the plugins editor. Do you think it's the right place for this or it can be a better place?
This issue shows up even when pressing backspace on a focused block. So i placed the relative fix into the focus plugin

### DIFF
--- a/draft-js-focus-plugin/src/index.js
+++ b/draft-js-focus-plugin/src/index.js
@@ -1,4 +1,4 @@
-import { EditorState } from 'draft-js';
+import { EditorState, SelectionState } from 'draft-js';
 import insertNewLine from './modifiers/insertNewLine';
 import setSelection from './modifiers/setSelection';
 import setSelectionToBlock from './modifiers/setSelectionToBlock';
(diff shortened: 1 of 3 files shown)


juliankrispel:
@kalimantos thank you so much for contributing!!

I think that would be the right place for this pr. But I'm worried that we're just fixing something that should actually be fixed in within draft-js. Surely, when you remove blocks, the entity should be removed too!

jdrzejb:
Do you think it could be merged into draft-js-plugins? At least untill they fix this in draft-js?

draft-js-focus-plugin/src/index.js
```diff
@@ -33,6 +33,35 @@ export default (config = {}) => {
...
       return 'not-handled';
     },
+    handleKeyCommand: (command, editorState, { setEditorState }) => {
+      if (command === 'backspace' && focusableBlockIsSelected(editorState, blockKeyStore)) {
+        const selection = editorState.getSelection();
+        const content = editorState.getCurrentContent();
+        const startKey = selection.getStartKey();
+        const blockAfter = content.getBlockAfter(startKey);
+        const blockMap = content.getBlockMap().delete(startKey);
+        const withoutAtomicBlock = content.merge({
```
juliankrispel: hm - I'm wondering whether this should sit inside the focus plugin - or rather inside the plugins editor as a patch over draft-js

jrock17:
I just ran into this as well. Any progress or ideas on how to remove atomic blocks from entity map on backspace?

juliankrispel:
I had to change this a fair bit but this should now be fixed as part of #62e00bff3092a74a6a8aba40e6ba3a843dc6d9cf

PR #960, opened 2017-11-10
Add missing decorators to editorState when props update

### LINKED ISSUE #262: Editor Plugin doesn't react to a new editorState
when changing the editorState such as re-setting with `EditorState.createEmpty()`, a user would expect the plugin editor to react to the change in state as DraftJS's Editor does. But the plugin editor only binds the decorators in `componentWillMount` so the editor loses the plugins

### LINKED ISSUE #359: Inconsistency with decorators and EditorState
Using version `1.1.0`.

lets say we have this function:

```js
function createState(text) {
  return EditorState.createWithContent(
    ContentState.createFromText(text)
  );
}
```

each time we create a new `EditorState`, we need to manually set the decorators. Yet, the first state we send gets the decorators from the plugins.

juliankrispel:
## Checklist

- [x] Tested in storybook

## Use-case/Problem
We add decorators to the editorState upon initalisation of the Editor but not when props update. This fixes this!

fixes #359 
fixes #262

## Implementation

- [x] reuse resolveDecorators function in willReceiveProps
- [x] add missing decorators to editorState object if editorState decorators are missing
- [x] refactor decorator resolving into separate function

### DIFF
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -65,11 +65,7 @@ class PluginEditor extends Component {
   componentWillReceiveProps(next) {
     if (
       this.props.editorState.getDecorator() !== null &&
-      (
-        next.editorState.getDecorator() === null ||
-        next.editorState.getDecorator().decorators.size
-          !== this.props.editorState.getDecorator().decorators.size
-      )
+      this.props.editorState.getDecorator() !== next.editorState.getDecorator()
     ) {
       const decorator = this.props.editorState.getDecorator();
       const editorState = EditorState.set(next.editorState, { decorator });
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -1,5 +1,5 @@
 /* eslint-disable no-continue,no-restricted-syntax */
-import React, { PureComponent } from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import {
   EditorState,
@@ -15,7 +15,7 @@ import * as defaultKeyBindingPlugin from './defaultKeyBindingPlugin';
 /**
  * The main editor component
  */
-class PluginEditor extends PureComponent {
+class PluginEditor extends Component {
 
   static propTypes = {
     editorState: PropTypes.object.isRequired,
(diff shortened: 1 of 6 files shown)


juliankrispel:
@mxstbr would be nice to have this reviewed by you since it specifically fixes a problem you're encountering

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -1,22 +1,21 @@
...
 import proxies from './proxies';
+import moveSelectionToEnd from './moveSelectionToEnd';
+import resolveDecorators from './resolveDecorators';
 import * as defaultKeyBindingPlugin from './defaultKeyBindingPlugin';
 
 /**
  * The main editor component
  */
-class PluginEditor extends Component {
+class PluginEditor extends PureComponent {
```
mxstbr: Please don't make it a PureComponent: They can't have `ref`s, so you can't do `this.editor.focus()` and stuff like that.

Testing the patch now!

mxstbr:
Testing this patch locally throws this error for me: https://reactjs.org/warnings/refs-must-have-owner.html

> addComponentAsRefTo(…): Only a ReactOwner can have refs. You might be adding a ref to a component that was not created inside a component’s render method, or you have multiple copies of React loaded.

juliankrispel:
@mxstbr can you give it another test? It seems to work fine for me now :+1:

mxstbr:
Still getting that error with the latest changes :confused: All we're doing is setting a `ref` on an editor.

mxstbr:
Hmm, I think I might know why this is happening, we've actually got two copies of ReactDOM loaded on the page because for some reason linking locally duplicates the dependency...

mxstbr:
Got this running locally by publishing a fork with these changes! Unfortunately it doesn't fix our issue, decorators are still not showing up :cry:

juliankrispel:
@mxstbr WAT!!!!

I want to fix this use-case damnit! It's a really common one! Is there a chance I can see the implementation?

juliankrispel:
@mxstbr ping me on slack or spectrum plz 👍

mxstbr:
Update: This fixes the original issue perfectly, my comments were just me being dumb. This is awesome and so so helpful, ship it!

juliankrispel:
@mxstbr https://www.npmjs.com/package/draft-js-plugins-editor 2.0.2

mxstbr:
👌

PR #966, opened 2017-11-14
Resolve infinite loop

mxstbr:
The latest release of the editor introduced an infinite loop with the
componentWillReceiveProps decorator fixes.

This patch resolves the infinite loop by explicitly checking for size
changes, and making sure to not call `onChange` when the new editor
state has decorators.

### DIFF
--- a/draft-js-plugins-editor/src/Editor/index.js
+++ b/draft-js-plugins-editor/src/Editor/index.js
@@ -63,14 +63,17 @@ class PluginEditor extends Component {
   }
 
   componentWillReceiveProps(next) {
-    if (
-      this.props.editorState.getDecorator() !== null &&
-      this.props.editorState.getDecorator() !== next.editorState.getDecorator()
-    ) {
-      const decorator = this.props.editorState.getDecorator();
-      const editorState = EditorState.set(next.editorState, { decorator });
-      this.onChange(moveSelectionToEnd(editorState));
-    }
+    const curr = this.props;
+    const currDec = curr.editorState.getDecorator();
+    const nextDec = next.editorState.getDecorator();
+
+    if (currDec === nextDec) return;
+    if (currDec && nextDec && currDec.decorators.size === nextDec.decorators.size) return;
+    if (!currDec && nextDec) return;
+
+    const decorator = curr.editorState.getDecorator();
+    const editorState = EditorState.set(next.editorState, { decorator });
+    this.onChange(moveSelectionToEnd(editorState));
   }
 
   componentWillUnmount() {


juliankrispel:
@mxstbr can you show me how to reproduce the infinite loop? :)

mxstbr:
Go to Spectrum, visit your dashboard when logged in, click around some threads, then try opening the thread editor.

mxstbr:
_ping_ Can we ship this? :blush:

juliankrispel:
Will look at it in a sec 🙌

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -63,14 +63,17 @@ class PluginEditor extends Component {
...
-      const decorator = this.props.editorState.getDecorator();
-      const editorState = EditorState.set(next.editorState, { decorator });
-      this.onChange(moveSelectionToEnd(editorState));
-    }
+    const curr = this.props;
+    const currDec = curr.editorState.getDecorator();
+    const nextDec = next.editorState.getDecorator();
+
+    if (currDec === nextDec) return;
+    if (currDec && nextDec && currDec.decorators.size === nextDec.decorators.size) return;
```
mxstbr: The important change is this one (I think), previously we would check `currDec !== nextDec` but that returns true even though they're the same in our app. (don't ask me why _that_ is, but this is an easier fix than figuring that out and will keep others from running into that problem)

By checking the size explicitly we make sure we only reinject them when they've changed.

mxstbr:
_ping_ :wink: We really need this fix, it's breaking the app for a large part of our user base.

draft-js-plugins-editor/src/Editor/index.js
juliankrispel: well if it's not exactly the same object it'll fail, so my guess is you somehow recreate the decorator? meh

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -63,14 +63,17 @@ class PluginEditor extends Component {
...
+    if (!currDec && nextDec) return;
```
juliankrispel: so this says - if the current decorator is null and the next exists we return and don't change decorators? Uhm, why? Can we remove this?

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -63,14 +63,17 @@ class PluginEditor extends Component {
...
+
+    const decorator = curr.editorState.getDecorator();
```
juliankrispel: this is already declared `currDec`

juliankrispel:
@mxstbr released `2.0.3`

draft-js-plugins-editor/src/Editor/index.js
mxstbr: Because then it's already in the next editor state, lol. So if you then call `onChange` you get a loop.

mxstbr: Oh wait, this could actually be simplified. One sec

tberman:
This seems broken when you have an editor with 1 decorator and you change the decorator to another one, right?

PR #968, opened 2017-11-15
Cleanup infinite loop fix

mxstbr:
This cleans up the infinite loop fix. Now that our prod app isn't broken
anymore, let's discuss about what we actually want to do here :p

I'm also not quite sure how to test for this, any ideas?


draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -67,12 +67,14 @@ class PluginEditor extends Component {
     const currDec = curr.editorState.getDecorator();
     const nextDec = next.editorState.getDecorator();
 
+    // If there is not current decorator, there's nothing to carry over to the next editor state
+    if (!currDec) return;
```
juliankrispel: thanks for the comments, that's much better <3

mxstbr:
Actually, I'm now realizing that the initial fix wasn't perfect either and can lead to unexpected behaviour. Really, we don't want to reinject the decorators from the old editorstate, we want to reinject the decorators from the props, no?

Imagine this case: (pseudo code)

```JS
const state = EditorState;
<Editor decorators={[mentions]} editorState={state} />

// then somebody does
onChange(EditorState.set(state, { decorator: otherDecorator }));

// This is already broken, editorState shouldn't have the otherDecorator now because according to the DraftJS Plugins editor the decorators come from props

// but then you can even go a step further and then do
<Editor decorators={[otherDecorator, mentions]} editorState={state} />

// ?????????
```

I think given the way the PluginsEditor changes the DraftJS decorators API the expected behaviour would be that **the decorators are always the ones from the props**. No matter which ones are passed in, those get applied and anything set via `EditorState.set(oldState, { decorators })` is just ignored. 

Basically, the PluginsEditor should "pretend" that the DraftJS is actually that decorators are bound to the component and not the state. (which in my personal opinion makes more sense anyway, but I disgress)

So really, what we should be doing in cWRP is **reinject the decorators from the decorators props if either the decorator prop has changed OR the new editor state has different decorators**.

I don't know if I'm making any sense, but anyways. What do you think?

mxstbr:
A maybe more coherent example, this doesn't work:

```JS
class MyEditor extends Component {
  constructor() {
    this.state = {
      decorators: [mentionsDecorator],
      editorState: createEmptyEditorState(),
    };
  }

  componentDidMount() {
    setTimeout(() => {
     // Eventhough we chagne the decorators, it still only has the mentions decorator
     this.setState({
        decorators: [linksDecorator]
     })
    }, 1000)
  }

  render() {
    return (
      <PluginsEditor decorators={this.state.decorators} editorState={this.state.editorState} />
    )
  } 
}
```

I think that should be supported and that makes so much more sense than the general DraftJS API. On top of that, if that's supported there's no reason to support `EditorState.set(x, { decorators })`, since that breaks the entire mental model that you introduce with the `decorators` prop.

The `decorators` prop should be the source of truth, and those should always be the decorators that are applied.

roundrobin:
Is this version already published on NPM? 

For me it's hard figure that out, since there are no releases listed.

mxstbr:
The fix from #966 was published as `2.0.3`, this branch hasn't been published I don't think.

mxstbr:
Now that I'm awake again, let me try one last time to explain in-depth what's happening and what I believe to be the right solution. The whole issue appears due to a simple mental model mismatch:

## Problem summary

**The DraftJS API sets decorators per-editor-state. This is bad news for `draft-js-plugins` because plugins (and thusly their decorators) are per-component, not per-editor-state.**

<details>
  <summary>This leads to confusion when you render a different editor state as all the decorators go missing. (click for code example of such an issue)</summary>

```JS
class MyEditor extends Component {
  state = {
    editorState: createEmptyEditorState(),
  };

  componentDidMount() {
    // After one second empty the editor
    // ⚠️ Now both the decorator from the markdownPlugin and the mentionDecorator are no longer applied! ⚠️
    // Everything else from the markdownPlugin is still used though, including blockRenderFn, blockStyleFn, blockStyleMap etc.
    setTimeout(() => {
     this.setState({
        editorState: createEmptyEditorState(),
     })
    }, 1000)
  }

  render() {
    return (
      <PluginsEditor
        plugins={markdownPlugin}
        decorators={[mentionDecorator]}
      />
    )
  } 
}
```

</details>

<br />

This is very unexpected behavior for many users (and I guess even some maintainers 😇) who get confused because suddenly half of their plugins no longer do anything even though they just rendered different content. There isn't much you can do to work around it, either

Let me state here that this is not a "bug" per-se, neither DraftJS nor draft-js-plugins have done anything "wrong". There's nobody at fault here, it's just that the API DraftJS chose doesn't lend itself to the way `draft-js-plugins` works.

## Solution summary

**`draft-js-plugins` has to pretend that decorators are applied per-component.**

The plugins and `decorator` prop are applied _per-component_, and the user confusion stems from the fact that they go missing when a different _state_ is rendered even though the same _component_ is rendered. The correct fix is quite simply that `draft-js-plugins` pretends as if decorators were per-component rather than per-editor-state.

Both decorators from plugins and from the `decorator` prop have to be reinjected into any new editor state that doesn't have them.

### Example cases

Let me give you some contrived examples of behavior that should be supported due to the mental model of `draft-js-plugins` but isn't right now:

<details>
  <summary>Changing decorators on-the-fly</summary>

```JS
class MyEditor extends Component {
  state = {
    editorState: createEmptyEditorState(),
    decorators: [mentionsDecorator]
  };

  componentDidMount() {
    // After one second change the applied decorators
    // ⚠️ Right now this doesn't work, the editor would still only use the mentionsDecorator ⚠️
    setTimeout(() => {
     this.setState({
        decorators: [linksDecorator]
     })
    }, 1000)
  }

  render() {
    return (
      <PluginsEditor
        decorators={this.state.decorators}
      />
    )
  } 
}
```

</details>

<br />

<details>
  <summary>Changing the plugins on-the-fly</summary>

```JS
class MyEditor extends Component {
  state = {
    editorState: createEmptyEditorState(),
    plugins: [markdownPlugin] // markdownPlugin is a placeholder for any plugin that uses decorators
  };

  componentDidMount() {
    // After one second change the applied plugins
    // ⚠️ Right now this doesn't work, the editor would still only use the decorator of the markdownPlugin ⚠️
    setTimeout(() => {
     this.setState({
        plugins: [mentionsPlugin] // mentionsPlugin is a placeholder for any other plugin that uses decorators
     })
    }, 1000)
  }

  render() {
    return (
      <PluginsEditor
        plugins={this.state.plugins}
      />
    )
  } 
}
```

</details>

<br />

<details>
  <summary>Changing the editor state on-the-fly</summary>

```JS
class MyEditor extends Component {
  state = {
    editorState: createEmptyEditorState(),
  };

  componentDidMount() {
    // After one second change the editor state
    // ⚠️ The mentionsDecorator would no longer be applied to the new editor state ⚠️
    setTimeout(() => {
     this.setState({
        editorState: createEmptyEditorState(),
     })
    }, 1000)
  }

  render() {
    return (
      <PluginsEditor
        decorators={[mentionsDecorator]}
      />
    )
  } 
}
```

</details>

<br />

All of these examples _seem_ like they should work perfectly fine based on the `draft-js-plugins` API, but they don't due to the DraftJS API. The last case (swapping out the editor state) was fixed by the original patch that introduced the `componentWillReceiveProps` logic, but it still doesn't cover the other cases.

> There is another edge case here, which is what happens when users manually do `EditorState.set(editorState, { decorator: someDecorator })`? I personally think that we should discard any manually injected decorators, since we're pretending that decorators are per-component those don't make any sense at all. This would likely need some good documentation though.

## TL;DR

Let's patch the DraftJS API and pretend that decorators are applied per-component.

> Sidenote: One could argue that per-editor-state-decorators is a bad design decision on DraftJS' side (which I personally would agree with), but that's besides the point since changing that API would take much longer than patching on top of it in `draft-js-plugins`

mxstbr:
_ping_ @juliankrispel let's ship this and then discuss about the decorators thing someplace else?

roundrobin:
@mxstbr thanks for your detailed explanation. Does what you're saying at the end (`Let's patch the DraftJS API and pretend that decorators are applied per-component.`) means it's already fixed in this Pull request, or is it something to figure our for a later release?

mxstbr:
@roundrobin this is only a first step that fixes a first bug we ran into, but there is more work to be done to get there.

roundrobin:
@mxstbr does this means in conclusion, that draft.js applications that make use of decorators can't use the draft-js-plugins?

juliankrispel:
I'll merge this for now, proper decorator resolution in the works (as well as types and a bunch of other things :+1:

juliankrispel:
@roundrobin no that doesn't mean that. What it means is that decorators need to be either passed via props or plugins, decorators mounted just in EditorState will be ignored, that's how we solve this problem...

taion:
This caused a weird regression for us. We're using the mentions plugin.

After I reset my editor with:

```js
    this.setState({
      // XXX https://github.com/facebook/draft-js/issues/73#issuecomment-191394559
      editorState: EditorState.moveFocusToEnd(EditorState.createEmpty()),
    });
```

Everything looks fine, but typing is screwed up. If I type `hey @J`, I end up with `Jey @h`, the sequence of characters looking something like:
- `h`
- `eh`
- `eyh`
- `ey h`
- `ey @h`
- `Jey @h`

(Same thing happens if I don't call `EditorState.moveFocusToEnd`.)

This doesn't happen against v2.0.4 of draft-js-plugins-editor, and it's not the aria-expanded fix either, nor is it any of the dependency bumps.

I don't really understand the decorator code here, so I can't comment on cause, but it looks like this introduced some state management glitch, perhaps with the added `moveSelectionToEnd` call.

juliankrispel:
Ugh thanks a lot for reporting this, I believe the Selektion state goes out
of sync here. I’m working on a change that will end the decorator malarkey
once and for all. Mind creating an issue and mentioning me? I’m afk til
later this week. Cheers

taion:
Done: https://github.com/draft-js-plugins/draft-js-plugins/issues/1100.

juliankrispel:
thanks @taion

PR #971, opened 2017-11-16
Rendering emoji select popover only on emoji select button press

khawerrind:
First of all thank you for the great plugins. 

While using `emoji select` page was loading very slow as it was making network requests to get all the images of the emojis of the `emoji select` popover. Now i am only rendering `emoji select popover` when user wants to see it. This way we are saving **1361** network requests for the images of emojis and hence making the page load time faster. 

This change will also come in handy when user is using multiple instances of emoji select. A normal example is comment boxes.


juliankrispel:
this issue has been fixed in #1030 thanks so much for the contribution however :heart:

juliankrispel:
whoops - merged  this by accident and reverted - apologies

PR #977, opened 2017-11-23
mention-plugin: fix tapping on entries to select on mobile devices

williamboman:
This fix doesn't feel 100% because it makes it impossible to scroll a list of entries on mobile devices (this is not a problem with the default components as they do not render a scrollable list, but consumers may customize the list component do so).

Feels like `onClick` would be better suited but [this commit](https://github.com/draft-js-plugins/draft-js-plugins/commit/d07d7926a) suggests otherwise.


bj00rn:
any progress on this PR?

williamboman:
Closing due to no activity. Solved it like so instead:


```jsx
import MentionEntry from 'draft-js-mention-plugin/lib/MentionSuggestions/Entry/defaultEntryComponent'

function MentionEntryComponent(props) {
    return <MentionEntry {...props} onTouchStart={(e) => {props.onMouseDown(e); props.onMouseUp();}} />
}

<MentionSuggestions
    ...
    entryComponent={MentionEntryComponent}
/>
```

ericls:
Why is this not merged?

PR #994, opened 2017-12-13
draft-js-plugins-editor: fix immutable.js dependecy to match other plugins

TomiS:
This PR fixes a small glitch in immutable.js dependency version range that causes yarn (for example) to install multiple unnecessary versions of immutable.js when this plugin is in use.


juliankrispel:
thanks @TomiS I bumped the package version and am publishing this now

PR #1002, opened 2017-12-29
[mention] fix cursor not moving up/down on non existing mention

dem:
## Checklist

## Use-case/Problem

When typing non-existing mention, empty dropdown open and up/down arrow catch by mention plugin. Cursor stick on  @non_exists_mention line.

## Demo

Before:
before

After problem fixed:
after

### DIFF
--- a/draft-js-mention-plugin/CHANGELOG.md
+++ b/draft-js-mention-plugin/CHANGELOG.md
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fixed bug where a user typed not existing mention @xxx and cursor is not moved with up/down arrow key
 - Escape spaces before mention trigger properly
 - Escape mention trigger regex properly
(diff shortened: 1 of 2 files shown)


juliankrispel:
thanks so much @dem

monkeyDledger:
Thanks @dem , the commit helps me understand why up/down arrow not working properly. But, I still have the problem after I initialize the editor,  empty dropdown open at the very beginning.

Maybe, do not use `e.preventDefault()` if mentionSuggestion is not active:

```
onDownArrow = (keyboardEvent) => {
    if (!this.state.isActive) return;
    keyboardEvent.preventDefault();
    const newIndex = this.state.focusedOptionIndex + 1;
    this.onMentionFocus(newIndex >= this.props.suggestions.size ? 0 : newIndex);
  };

onTab = (keyboardEvent) => {
    if (!this.state.isActive) return;
    keyboardEvent.preventDefault();
    this.commitSelection();
  };

onUpArrow = (keyboardEvent) => {
    if (!this.state.isActive) return;
    keyboardEvent.preventDefault();
    if (this.props.suggestions.size > 0) {
      const newIndex = this.state.focusedOptionIndex - 1;
      this.onMentionFocus(newIndex < 0 ? this.props.suggestions.size - 1 : newIndex);
    }
  };
```

PR #1003, opened 2017-12-29
[mention] fix mention detection in styled text

dem:
## Checklist

## Use-case/Problem
When trying to detect mention in differently styled text,  detection is not work. See demo.

We have more worse case with same reason of problem, in our project, where mention plugin is used. When adding mention in series: `@first_mention @second_mention` second mention is not detected.
In depth problem looks like:
1. From the portal detection decorated text is ` @second_mention` (started from space)
2. First ` `(space) goes in first leaf and `@second_mention` goes in second leaf.
3. Mention is not detected by regexp in `@second_mention` - no space here and it is not start at the begin of line.

## Implementation

Do not check leaves separately, use full decorated text to check.

## Demo

Before:
before

After fix:
after


juliankrispel:
Hi @dem thanks for that! Good catch!

juliankrispel:
Hi @dem thanks for that! Good catch!

PR #1004, opened 2017-12-29
[mention] fix suggestions dropdown position in case of line wrap

dem:
## Checklist

## Use-case/Problem

When typing @mention in the on of line, it goes to beginning of next line, but suggestion dropdown stays on the same vertical position and overlaps @mention text.

## Implementation

Calculate position of suggestion portal in relation of bottom (not top) decorated rectangle.

## Demo

Before: 
before

After fix:
after


juliankrispel:
Awesome fix thank you @dem

PR #1018, opened 2018-01-14
Inline toolbar positioning fix

newsiberian:
## Checklist

## Use-case/Problem
Hi.
This PR tries to fix two things: outputting the toolbar beyond the relative parent (editor) and changing the size of the toolbar in some cases. These bugs will be demonstrated in demos.

## Implementation

Fix introduce new logic that not allows toolbar to go beyond the editor block. To make it looks good we have to make pointer positioning floating. The way I used to do this is an external style elements for each toolbar (there may be several editors on the screen), because pointers are working via `before`, `after` preudo classes.

## Demo
Here how it was before:
before-1
and on a small screens:
before-2

and after this implementation:
after-1
after-2

### DIFF
--- a/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
+++ b/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
@@ -1,5 +1,6 @@
 /* eslint-disable react/no-array-index-key */
 import React from 'react';
+import PropTypes from 'prop-types';
 import { getVisibleSelectionRect, genKey } from 'draft-js';
 
 const getRelativeParent = (element) => {
@@ -16,6 +17,17 @@ const getRelativeParent = (element) => {
 };
 
 export default class Toolbar extends React.Component {
+  static propTypes = {
+    /**
+     * An indents from the left / right window borders
+     */
+    toolbarMargin: PropTypes.number
+  };
+
+  static defaultProps = {
+    toolbarMargin: 5
+  };
+
   constructor(props) {
     super(props);
     this.state = {
@@ -86,20 +98,19 @@ export default class Toolbar extends React.Component {
 
       const relativeParent = getRelativeParent(this.toolbar.parentElement);
       const relativeRect = (relativeParent || document.body).getBoundingClientRect();
+      const windowWidth = document.documentElement.clientWidth;
 
       // if parent block width is wider than the toolbar, we must forcibly set its
       // height to prevent unexpected geometry changes
-      if (relativeRect.width > this.toolbar.offsetWidth) {
+      if (windowWidth > this.toolbar.offsetWidth) {
         this.toolbar.style.width = `${this.state.width}px`;
       }
 
       const toolbarHalfWidth = this.toolbar.offsetWidth / 2;
       // calculating the middle of the text selection
       const fromBeginningToMiddle = (selectionRect.left + (selectionRect.width / 2));
-      // calculating the distance from editor left side to the text selection middle
-      const fromParentBeginning = fromBeginningToMiddle - relativeRect.left;
       // the same but against editor right side
-      const beforeParentEnd = relativeRect.right - fromBeginningToMiddle;
+      const beforeWindowEnd = windowWidth - fromBeginningToMiddle;
 
       // the selection is closer to parent beginning than half of the toolbar
       // +-----------------------------------------------+
@@ -110,10 +121,14 @@ export default class Toolbar extends React.Component {
       // |  +--+                                         |
       // |   ^^ selection                                |
       // +-----------------------------------------------+
-      if (fromParentBeginning < toolbarHalfWidth) {
-        metrics.ruleValue = toolbarHalfWidth;
+      if (fromBeginningToMiddle < toolbarHalfWidth) {
+        // shift computations are different for relative editor and body
+        const leftShift = relativeParent
+          ? relativeRect.left
+          : 0;
+        metrics.ruleValue = (toolbarHalfWidth - leftShift) + this.props.toolbarMargin;
         metrics.shift = 'left';
-      } else if (beforeParentEnd < toolbarHalfWidth) {
+      } else if (beforeWindowEnd < toolbarHalfWidth) {
         // the same, but relative to the parent end
         // +-----------------------------------------------+
         // |                                 vvv toolbar   |
@@ -124,7 +139,11 @@ export default class Toolbar extends React.Component {
         // |                             selection ^^      |
         // |                                               |
         // +-----------------------------------------------+
-        metrics.ruleValue = -toolbarHalfWidth;
+        // shift computations are different for relative editor and body
+        const rightShift = relativeParent
+          ? windowWidth - relativeRect.right
+          : 0;
+        metrics.ruleValue = (-toolbarHalfWidth - rightShift) + this.props.toolbarMargin;
         metrics.rule = 'right';
         metrics.shift = 'right';
       } else {
@@ -160,11 +179,11 @@ export default class Toolbar extends React.Component {
         this.toolbar.classList.add(pointerClassName);
 
         if (metrics.shift === 'left') {
-          metrics.shiftValue = (toolbarHalfWidth - metrics.ruleValue) +
-            fromParentBeginning;
+          metrics.shiftValue = fromBeginningToMiddle - this.props.toolbarMargin;
         } else {
           metrics.shiftValue = this.toolbar.offsetWidth -
-            (relativeRect.right - (selectionRect.right - (selectionRect.width / 2)));
+            (windowWidth - (selectionRect.right - (selectionRect.width / 2)) -
+            this.props.toolbarMargin);
         }
 
         styleEl.sheet.insertRule(`.${pointerClassName}::after { left: ${metrics.shiftValue}px}`, 0);
--- a/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
+++ b/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
@@ -81,17 +81,10 @@ export default class Toolbar extends React.Component {
       };
       // boundings of the selected text
       const selectionRect = getVisibleSelectionRect(window);
-      if (!selectionRect) return;
+      const selection = this.props.store.getItem('getEditorState')().getSelection();
+      if (!selectionRect || selection.isCollapsed()) return;
 
       const relativeParent = getRelativeParent(this.toolbar.parentElement);
-      if (relativeParent) {
-        // avoid false
-        const fontSize = window.getComputedStyle(relativeParent, null).getPropertyValue('font-size');
-        // exit if no selection was made
-        if (selectionRect.left === selectionRect.right && selectionRect.height === parseFloat(fontSize)) {
-          return;
-        }
-      }
       const relativeRect = (relativeParent || document.body).getBoundingClientRect();
 
       // if parent block width is wider than the toolbar, we must forcibly set its
--- a/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
+++ b/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
@@ -73,10 +73,6 @@ export default class Toolbar extends React.Component {
 
       if (typeof this.state.width !== 'number') {
         this.setState({ width: this.toolbar.offsetWidth });
-      } else {
-        // toolbar width must forcibly overwritten to prevent unexpected geometry
-        // changes
-        this.toolbar.style.width = `${this.state.width}px`;
       }
 
       const metrics = {
@@ -176,6 +172,9 @@ export default class Toolbar extends React.Component {
       style.visibility = 'visible';
       style.transform = 'translate(-50%) scale(1)';
       style.transition = 'transform 0.15s cubic-bezier(.3,1.2,.2,1)';
+      // toolbar width must forcibly overwritten to prevent unexpected geometry
+      // changes
+      style.width = this.state.width;
     } else {
       style.transform = 'translate(-50%) scale(0)';
       style.visibility = 'hidden';
--- a/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
+++ b/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
@@ -160,8 +160,6 @@ export default class Toolbar extends React.Component {
       }
 
       if (typeof metrics.shift === 'string') {
-        this.toolbar.classList.add(this.state.pointerClassName);
-
         if (metrics.shift === 'left') {
           metrics.shiftValue = fromBeginningToMiddle - this.props.toolbarMargin;
         } else {
@@ -224,7 +222,7 @@ export default class Toolbar extends React.Component {
 
     return (
       <div
-        className={theme.toolbarStyles.toolbar}
+        className={[theme.toolbarStyles.toolbar, pointerClassName].join(' ')}
         style={this.getStyle()}
         ref={this.handleToolbarRef}
       >
--- a/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
+++ b/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
@@ -75,16 +75,13 @@ export default class Toolbar extends React.Component {
         this.setState({ width: this.toolbar.offsetWidth });
       }
 
-      const metrics = {
-        rule: 'left',
-        ruleValue: 0,
-        shift: null,
-        shiftValue: 0
-      };
+      let alignment = null;
+      let horizontalOffset = 0;
+
       // boundings of the selected text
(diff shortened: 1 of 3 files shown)


freedomlang:
@newsiberian  Thanks for the pr, it would be help if you can stay with our pr template

newsiberian:
Hi, @freedomlang,  thanks for giving this a chance. I've changed pr description.

juliankrispel:
just fyi I had a brief look but this needs a real thorough review and some testing, will get back to this soon I hope

Overall looks like very thoughtful work ♥️  thank you so much!

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -45,23 +61,126 @@ export default class Toolbar extends React.Component {
...
+      const selectionRect = getVisibleSelectionRect(window);
+      if (!selectionRect) return;
+
       const relativeParent = getRelativeParent(this.toolbar.parentElement);
-      const toolbarHeight = this.toolbar.clientHeight;
+      if (relativeParent) {
+        // avoid false
+        const fontSize = window.getComputedStyle(relativeParent, null).getPropertyValue('font-size');
+        // exit if no selection was made
+        if (selectionRect.left === selectionRect.right && selectionRect.height === parseFloat(fontSize)) {
```
juliankrispel: wouldn't `if (selectionState.isCollapsed()) {` do the same thing?

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -45,23 +61,126 @@ export default class Toolbar extends React.Component {
...
+      // the selection is closer to parent beginning than half of the toolbar
+      // +-----------------------------------------------+
+      // |          vv toolbar                           |
+      // | +------------------+                          |
+      // | +------------------+                          |
+      // |                                               |
+      // |  +--+                                         |
+      // |   ^^ selection                                |
+      // +-----------------------------------------------+
+      if (fromParentBeginning < toolbarHalfWidth) {
```
juliankrispel: rather than the parent, shouldn't it just be the viewport?

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -45,23 +61,126 @@ export default class Toolbar extends React.Component {
...
+        document.head.removeChild(
+          document.getElementById(`${styleId}-${this.state.nonce}`)
+        );
+      }
+
+      if (typeof metrics.shift === 'string') {
+        const styleEl = document.createElement('style');
+        styleEl.setAttribute('id', `${styleId}-${this.state.nonce}`);
+        document.head.appendChild(styleEl);
+        this.toolbar.classList.add(pointerClassName);
```
juliankrispel: this is a big no-no in react land. One should only use the react api to affect dom-changes, otherwise the virtual DOM and the actual DOM diverge and that almost certainly will cause exceptions or bugs.

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -45,23 +61,126 @@ export default class Toolbar extends React.Component {
...
+        if (metrics.shift === 'left') {
+          metrics.shiftValue = (toolbarHalfWidth - metrics.ruleValue) +
+            fromParentBeginning;
+        } else {
+          metrics.shiftValue = this.toolbar.offsetWidth -
+            (relativeRect.right - (selectionRect.right - (selectionRect.width / 2)));
+        }
+
+        styleEl.sheet.insertRule(`.${pointerClassName}::after { left: ${metrics.shiftValue}px}`, 0);
+        styleEl.sheet.insertRule(`.${pointerClassName}::before { left: ${metrics.shiftValue}px}`, 0);
```
juliankrispel: I'm very unclear why we need to manipulate the DOM and inject a stylesheet. Surely all the positioning logic can live inside `getStyle`

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
newsiberian: thanks for the tip

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
newsiberian: If I understood you correctly this is also done. Now it looks like this:
inlinetoolbar

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
newsiberian: Ok, this is the most unclear part to me. Do you know how we can change pointer position, read "css pseudo-class properties" with react / inline styling?

juliankrispel:
Why do you need pseudo classes exactly? The way I‘d implement this is to
adjust the object that gets returned by getStyle...

newsiberian:
@juliankrispel, because we have a CSS arrow in toolbar. It is builded via ':before', ':after'. We are changing it position each time.

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
juliankrispel: Hummm. Good point. I like what you did there, it's quite nicely done :+1:. Let me have a tinker with this. I might have time today or tomorrow morning :+1:

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
newsiberian: @juliankrispel, seems like I found a solution. We can create react tag for our styles.

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -239,6 +228,8 @@ export default class Toolbar extends React.Component {
         style={this.getStyle()}
         ref={this.handleToolbarRef}
       >
+        {shift !== 0 &&
+          <style>{`.${pointerClassName}::before, .${pointerClassName}::after${shift};`}</style>}
```
juliankrispel: yes good idea I love that!

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -153,30 +159,8 @@ export default class Toolbar extends React.Component {
...
-          document.getElementById(`${styleId}-${this.state.nonce}`)
-        );
-      }
-
       if (typeof metrics.shift === 'string') {
-        const styleEl = document.createElement('style');
-        styleEl.setAttribute('id', `${styleId}-${this.state.nonce}`);
-        document.head.appendChild(styleEl);
-        this.toolbar.classList.add(pointerClassName);
+        this.toolbar.classList.add(this.state.pointerClassName);
```
juliankrispel: I think this is the last dom manipulation in here. Rather than doing this you could simply add `this.state.pointerClassName` to the `className` prop in the render method

juliankrispel: that's on line 227 - instead of `className={theme.toolbarStyles.toolbar}` you could do `className={[theme.toolbarStyles.toolbar, this.state.pointerClassName].join(' ')}

newsiberian: Thanks for the tip. Done)

juliankrispel:
ok that looks really nice now, one more thing could you put the code responsible for positioning the pointer in a separate function, just so it's a little bit clearer what's going on :), can be a class method or a function outside the class, up to you.

Thank you so much this is really neat 😍

newsiberian:
@juliankrispel, I moved code you mentioned to separate method. Is that looks suitable?

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -16,21 +17,53 @@ const getRelativeParent = (element) => {
...
+      pointerClassName: '',
+      /**
+       * pointerClassName internals here. It could look like: "{ left: 1px; }"
+       * @type {number|string}
+       */
+      pointerShift: 0
+    };
+
+    // we need to flush toolbar styles on each appearance (especially its width)
+    this.toolbar = null;
```
juliankrispel: this.toolbar is undefined when this line is executed since the constructor is run before the component is mounted so I'm not sure what this line does?

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -16,21 +17,53 @@ const getRelativeParent = (element) => {
...
   }
 
   componentWillMount() {
     this.props.store.subscribeToItem('selection', this.onSelectionChanged);
+    this.setState({ pointerClassName: `draft-js-inline-toolbar-pointer-${genKey()}` });
```
juliankrispel: unnecessary, see above

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -45,25 +78,99 @@ export default class Toolbar extends React.Component {
...
   onSelectionChanged = () => {
     // need to wait a tick for window.getSelection() to be accurate
     // when focusing editor with already present selection
     setTimeout(() => {
       if (!this.toolbar) return;
-      const relativeParent = getRelativeParent(this.toolbar.parentElement);
-      const toolbarHeight = this.toolbar.clientHeight;
-      const relativeRect = (relativeParent || document.body).getBoundingClientRect();
+
+      if (typeof this.state.width !== 'number') {
```
juliankrispel: Yea I don't see why this is necessary at all, see other comments about width...

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -45,25 +78,99 @@ export default class Toolbar extends React.Component {
...
+      if (!selectionRect || selection.isCollapsed()) return;
 
-      if (!selectionRect) return;
+      const relativeParent = getRelativeParent(this.toolbar.parentElement);
+      const relativeRect = (relativeParent || document.body).getBoundingClientRect();
+      const windowWidth = document.documentElement.clientWidth;
+
+      // if parent block width is wider than the toolbar, we must forcibly set its
+      // height to prevent unexpected geometry changes
+      if (windowWidth > this.toolbar.offsetWidth) {
```
juliankrispel: I don't understand this problem. What do you mean by `parent block width`? If you're talking about draft js blocks then those are expected to be wider.

juliankrispel: Also, `height` is a typo here I guess?

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -16,21 +17,53 @@ const getRelativeParent = (element) => {
...
-
+  static propTypes = {
     /**
-     * If this is set, the toolbar will render this instead of the regular
-     * structure and will also be shown when the editor loses focus.
-     * @type {Component}
+     * An indents from the left / right window borders
      */
-    overrideContent: undefined
+    toolbarMargin: PropTypes.number
```
juliankrispel: why can we not just use css for this?

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -88,13 +195,38 @@ export default class Toolbar extends React.Component {
...
+        return `{ left: ${fromBeginningToMiddle - this.props.toolbarMargin}px; }`;
+      }
+
+      return `{ left: ${this.toolbar.offsetWidth -
+      (windowWidth - (selectionRect.right - (selectionRect.width / 2)) -
+        this.props.toolbarMargin)}px; }`;
+    }
+
+    // explicitly set to zero, because it may already be mutated
+    return 0;
```
juliankrispel: we don't use 0 as a falsey value, `null` is more explicit

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
newsiberian: Hi, I did some research today. This prop is not really related with css, it is used as some abstract argument while positioning. So, if we will use css margin instead of this prop, that will not give us any advantages, except for increasing calculation complexity, because will then will have to take into account these css margins.

newsiberian: So, if you insist, we can implement css margins

juliankrispel:
Hello 👋 thanks for replying so soon!

What I’m saying is we can and should leave something like that in userland
via css. We don’t want to cater to too many usecases imo

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -16,21 +17,53 @@ const getRelativeParent = (element) => {
...
+       * If this is set, the toolbar will render this instead of the regular
+       * structure and will also be shown when the editor loses focus.
+       * @type {Component}
+       */
+      overrideContent: undefined,
+      /**
+       * We are holding default toolbar width to prevent geometry changing, that
+       * happens very often
+       */
+      width: null,
```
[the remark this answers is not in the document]
newsiberian: Hi, @juliankrispel.
This is important. Browser sometimes makes an attempt to change element's geomentry to fit it in a 'better' way to window. But for us this is not suitable, because our calculations strictly depends on width size. You can see what's going on in the first gif. So we are recording toolbar width after first rendering and then override it sometimes

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
newsiberian: I'm agree, this is very odd comment. Honestly I do not remember why I wrote this :smile: and what this check should do. We just need to override `width` everytime.

newsiberian:
Please, review

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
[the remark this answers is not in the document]
juliankrispel: @newsiberian this can also be solved via css - like with `white-space: nowrap` - just needs to be added to the container of the buttons. Give it a shot please, I'd rather not add state if we don't need it

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -45,25 +63,103 @@ export default class Toolbar extends React.Component {
...
+        // changes
+        this.toolbar.style.width = `${this.state.width}px`;
+      }
+
+      const metrics = {
+        rule: 'left',
+        ruleValue: 0,
+        shift: null,
+        shiftValue: 0
+      };
```
juliankrispel: this object is fairly confusing imo.

Make them variables instead of an object and define the following:

```
let alignment = 'left';
let horizontalOffset = 0;
```

You shouldn't really need more than these two values. Since shift and rule are both set to either 'left' or 'right'

shiftValue is unused...

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -45,25 +63,103 @@ export default class Toolbar extends React.Component {
...
+      const selection = this.props.store.getItem('getEditorState')().getSelection();
+      if (!selectionRect || selection.isCollapsed()) return;
 
-      if (!selectionRect) return;
+      const relativeParent = getRelativeParent(this.toolbar.parentElement);
+      const relativeRect = (relativeParent || document.body).getBoundingClientRect();
+      const windowWidth = document.documentElement.clientWidth;
+      // we should take into account a case when we don't have relative parent,
+      // but our body has a margin
+      const bodyMargin = relativeParent ? 0 : getMargin(document.body);
```
juliankrispel: what I meant here was that we should leave this completely in user land.  I don't think we should support this usecase.

juliankrispel:
@newsiberian there's still a fair bunch of unaddressed comments, added some new comments as well :+1:

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
newsiberian: Hi, this is not related with previous conversation. This is new bugfix. It is related with cases like we have in storybook, when body has a margin, and parent not relative. We should take into account this body margin in that case.

newsiberian:
Hi, @juliankrispel , anything else I can do here?

devkev95:
Awesome stuff, I've been trying to make the editor to work the way it's working here for days and couldn't find any way to do it until I stumbled upon this PR.

Any idea of when this will be merged?

Issue #1026, opened 2018-01-30
draft-js-mention-plugins: mentionSuggestions is handling up/down arrow keys even when it's inactive


gnestor:
## Use-case/Problem

![](https://user-images.githubusercontent.com/1851359/29960622-b4e730e6-8efc-11e7-893e-c5857010d9c6.gif)

This was initially reported in https://github.com/draft-js-plugins/draft-js-plugins/issues/856. I'm running the current master of this repo and still running into this issue. I have looked at the source and I will submit a PR that I think should fix this. I can't reproduce this in [the example](https://codesandbox.io/s/MjG72jX2R) because the mentions don't change in that example, therefore the [mentionSuggestions.componentWillReceiveProps](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/src/MentionSuggestions/index.js#L52-L58) is never called, but when it is, it will either open the dropdown or close it based on whether the suggestions have changed or not, which doesn't make sense to me (if the dropdown is open when props change, keep it open, if closed, then keep it closed).

davidmfoley:
I ran into this because I was filtering the suggestions upon render, which makes the equality check on `suggestions` always fail here: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/src/MentionSuggestions/index.js#L36

(since it is two different references to arrays what happen to have the same items)

Switched over to filtering the suggestions in `onSearchChange`, and storing the result in `state` and got my up and down arrows back.

ilya-lopukhin:
> I ran into this because I was filtering the suggestions upon render, which makes the equality check on `suggestions` always fail here: https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/src/MentionSuggestions/index.js#L36
> 
> (since it is two different references to arrays that happen to have the same items)
> 
> Switched over to filtering the suggestions in `onSearchChange`, and storing the result in `state` and got my up and down arrows back.

I had a similar issue and and based on your workaround i wrapped my suggestions into `useMemo` hook. The problem is gone 🔥

PR #1027, opened 2018-01-30
draft-js-mention-plugins: Only handle up/down arrow keys when mentionSuggestions is active

gnestor:
Closes https://github.com/draft-js-plugins/draft-js-plugins/issues/1026

This also excludes `isFocused` from `defaultEntryComponent` because otherwise it throws this React error:

```
Warning: React does not recognize the `isFocused` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `isfocused` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/index.js
@@ -49,14 +49,6 @@ export class MentionSuggestions extends Component {
     this.props.callbacks.onChange = this.onEditorStateChange;
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.suggestions.size === 0 && this.state.isActive) {
-      this.closeDropdown();
-    } else if (nextProps.suggestions.size > 0 && !nextProps.suggestions.equals(this.props.suggestions) && !this.state.isActive) {
-      this.openDropdown();
-    }
-  }
-
   componentDidUpdate = (prevProps, prevState) => {
     if (this.popover) {
       // In case the list shrinks there should be still an option focused.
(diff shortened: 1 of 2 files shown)


draft-js-mention-plugin/src/MentionSuggestions/index.js
```diff
@@ -49,14 +49,6 @@ export class MentionSuggestions extends Component {
...
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.suggestions.size === 0 && this.state.isActive) {
-      this.closeDropdown();
-    } else if (nextProps.suggestions.size > 0 && !nextProps.suggestions.equals(this.props.suggestions) && !this.state.isActive) {
-      this.openDropdown();
-    }
-  }
-
```
juliankrispel: this can't be removed I'm afraid - and it's out of scope for this pr really

gnestor: This is how I resolved the arrow key issue for myself, but I understand that this probably serves a purpose. The issue I was running into was `this.state.isActive` was `true` even though the suggestions wasn't open, therefore `keyboardEvent.preventDefault()` below was being called and the up/down arrow key events were being intercepted by this. What is the purpose of opening the suggestions when `it's not active and the suggestions is not empty and the suggestions have changed`?

juliankrispel: All fair points. I'm afraid that removing this causes other things to break. I can't remember off the top of my head but it's there for a reason, even though it's awkward. A bunch of logic will need to be rewritten to fix these issues :/

PR #1028, opened 2018-02-01
Warnings are displayed only in the not production

sugarshin:
Hi,

I changed to warnings of `Immutable.List` are displayed only in the `process.env.NODE_ENV !== production` . Because not necessary in production.


juliankrispel:
thanks a lot for the pr :heart:

Issue #976, opened 2017-11-23
Side toolbar not placed correctly when parent(s) is positioned relative


robgietema:
When the editor is inside a parent (or multiple parents) with position relative the side toolbar is not placed correctly. It will be place absolute inside the relative parent but should be absolute to the document. Could be the same issue as the inline toolbar #531

lednhatkhanh:
+1 for this problem, I am facing the same thing....

juliankrispel:
this was fixed in #1029

PR #1029, opened 2018-02-02
Fix side toolbar positioning when there's a positioned parent.

amannn:
## Checklist

## Use-case/Problem

This PR improves the positioning of the side toolbar. The current implementation didn't work when there's a positioned parent (see #976).

## Implementation

The class `DraftEditor-root` is applied to the very root node of the editor by draft.js itself – this class allows to find that node. The side toolbar is a sibling of the editor root node, therefore we can position the toolbar by considering the offset of the editor node within it's next positioned parent. Note that if there's no explicitly positioned parent, then `body` will be taken into account (this is the case in the docs).

Tested in Chrome, Firefox, Safari and IE11.

### DIFF
--- a/draft-js-side-toolbar-plugin/src/components/Toolbar/index.js
+++ b/draft-js-side-toolbar-plugin/src/components/Toolbar/index.js
@@ -36,13 +36,19 @@ export default class Toolbar extends React.Component {
     // Note: need to wait on tick to make sure the DOM node has been create by Draft.js
     setTimeout(() => {
       const node = document.querySelectorAll(`[data-offset-key="${offsetKey}"]`)[0];
-      const top = node.getBoundingClientRect().top;
-      const editor = this.props.store.getItem('getEditorRef')().refs.editor;
-      const scrollY = window.scrollY == null ? window.pageYOffset : window.scrollY;
+
+      // The editor root should be two levels above the node from
+      // `getEditorRef`. In case this changes in the future, we
+      // attempt to find the node dynamically by traversing upwards.
+      let editorRoot = this.props.store.getItem('getEditorRef')().refs.editor;
+      while (editorRoot.className.indexOf('DraftEditor-root') === -1) {
+        editorRoot = editorRoot.parentNode;
+      }
+
       this.setState({
         position: {
-          top: (top + scrollY),
-          left: editor.getBoundingClientRect().left - 80,
+          top: node.offsetTop + editorRoot.offsetTop,
+          left: editorRoot.offsetLeft - 80,
           transform: 'scale(1)',
           transition: 'transform 0.15s cubic-bezier(.3,1.2,.2,1)',
         },


amannn:
@juliankrispel I just tested this PR and #1031 in a project of mine and it looks like they work fine. Same for the docs 🙂

amannn:
@juliankrispel I've just fixed a small possible null pointer that I've encountered in an app. I guess it's related to the wrapping timeout which can result into reading a ref at a point where it happens to not be available. It's a very rare case though and usually doesn't happen.

I made the same change for #1031. Do you think we can merge and release these PRs soon? Currently the side- and inline-toolbar plugins aren't usable in IE11 if there's a positioned parent. Thanks!

juliankrispel:
Ok I’m doing a bunch of maintenance this weekend.

amannn:
Great, thanks a lot!

draft-js-side-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -36,13 +36,21 @@ export default class Toolbar extends React.Component {
...
+      while (editorRoot.className.indexOf('DraftEditor-root') === -1) {
+        editorRoot = editorRoot.parentNode;
+      }
+
       this.setState({
         position: {
-          top: (top + scrollY),
-          left: editor.getBoundingClientRect().left - 80,
+          top: node.offsetTop + editorRoot.offsetTop,
+          left: editorRoot.offsetLeft - 80,
```
juliankrispel: ok I just tested this and this now seems to break when the root node isn't relatively positioned.

Mind fixing that (I guess you just need to treat the positioning differently if a position is set)

Also - if you could add a story for relatively positioned ones that'd be fab.

amannn: I think you're referring to the existing stories for the side toolbar – is that correct? It was already working in the docs, so I was a bit confused. I've seen that the Draft.css stylesheet wasn't loaded in the Storybook. I've added that import and a story and now everything seems fine.

juliankrispel: it was however working before (non positioned and without the draft.css stylesheet)?

juliankrispel:
@amannn can you also resolve merge conflicts :+1:

draft-js-side-toolbar-plugin/src/components/Toolbar/index.js
amannn: Yeah, that's true. It's because the positioning strategy is different now. I think it can be considered a bug though if Draft.css is missing in an app. From the Draft.js docs:
> The Draft framework includes a handful of CSS resources intended for use with the editor, available in a single file via the build, Draft.css.

> This CSS should be included when rendering the editor, as these styles set defaults for text alignment, spacing, and other important features. Without it, you may encounter issues with block positioning, alignment, and cursor behavior.

> If you choose to write your own CSS independent of Draft.css, you will most likely need to replicate much of the default styling.

([source](https://draftjs.org/docs/advanced-topics-issues-and-pitfalls.html#missing-draft-css))

We're also [mentioning this in our FAQ](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/FAQ.md#my-default-styles-are-broken).

juliankrispel: ok I have it working now o_O . this is awesome ♥️ !!!!

the only thing now is that when we don't include Draft.css it looks like this:

image

Bear in mind that draft.css actually adds position:relative to the root...

draft-js-side-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -42,7 +42,10 @@ export default class Toolbar extends React.Component {
       // attempt to find the node dynamically by traversing upwards.
       const editorRef = this.props.store.getItem('getEditorRef')();
       if (!editorRef) return;
-      let editorRoot = editorRef.refs.editor;
+
+      // this keeps backwards-compatibility with react 15
+      let editorRoot = editorRef.refs && editorRef.refs.editor
+        ? editorRef.refs.editor : editorRef.editor;
```
juliankrispel: :+1:

draft-js-side-toolbar-plugin/src/components/Toolbar/index.js
amannn: Yeah, right – you really have to include the stylesheet 🙂. It's that `position: relative` on the editor root that is necessary for the positioning to work correctly (also every other kind of position will do, as long as it isn't `static`). The PR fixes the use case where theres an enclosing parent that is positioned relatively.

juliankrispel: hmhmhm ok! We just need to make that extra clear imo!

juliankrispel: I'll open an issue to add this to the docs

amannn:
Great, thanks! 👍

PR #1030, opened 2018-02-02
Prevent all EmojiOne emoji loading when EmojiSelect is rendered, instead load them progressively as needed

joshfarrant:
## Checklist

## Use-case/Problem

When using EmojiOne emoji (rather than native) all of the required emoji (1000+) are fetched as soon as the `<EmojiSelect />` component is rendered, regardless of whether or not the selector is open. This can cause long initial load times, especially in Firefox.

## Implementation

This solution prevents the loading of the `<Groups />` component when the `<EmojiSelect />` isn't open. This stops all of the emoji being fetched as soon as the `<EmojiSelect />` is rendered and instead waits until it has been opened by the user.

Also, this adds progressive loading to the `<EmojiSelect />`, so emoji aren't fetched until that specific `<Group />` is reached.


juliankrispel:
this awesome stuff - testing and reviewing now

draft-js-emoji-plugin/src/components/EmojiSelect/Popover/Groups/Group/index.js
```diff
@@ -14,9 +14,32 @@ export default class Group extends Component {
...
+    if (nextProps.isActiveGroup) {
+      this.setState({
+        hasRenderedEmoji: true,
+      });
+    }
+  }
+
+  shouldComponentUpdate = (nextProps) => {
+    // If the emoji have been rendered once, never update again
+    if (this.state.hasRenderedEmoji) {
```
juliankrispel: why can't we just check `nextProps.isActiveGroup` here?

joshfarrant: If I remember correctly, the component was being re-rendered at some points when `isActiveGroup` was `false`. There didn't appear to be any good reasons for these re-renders, so I used the `hasRenderedEmoji` state to prevent the (I think) unnecessary re-renders.

juliankrispel:
Right - my question though is - why set state at all and not just use
nexProps ?

joshfarrant:
`hasRenderedEmoji` is essentially tracking whether or not `isActiveGroup` has _ever_ been `true`. 

If it has been `true` at least once, then we know that we've rendered (and, more importantly, fetched) the emoji before, so there's no need to perform any additional renders.

As the component is tracking whether or not `isActiveGroup` has ever been true across multiple renders (not just the next/last render), then we need to use the state to track it.

If we got rid of `hasRenderedEmoji` and just used `nextProps.isActiveGroup` then behaviour will be slightly different when the component re-renders. Imagine a case where the component is initially rendered with `isActiveGroup: false`. `isActiveGroup` is then updated to `true`, so the component fetches the emoji and renders, which is all good.

If `isActiveGroup` is then set to `false` again, then we don't re-render (because of [this](https://github.com/joshfarrant/draft-js-plugins/blob/17706d3effbb5352ee8e4eb11da8f0d06fa2e0f0/draft-js-emoji-plugin/src/components/EmojiSelect/Popover/Groups/Group/index.js#L41)). However, when it is next updated to `true`, it _will_ trigger a re-render as `nextProps.isActiveGroup = true`, so `shouldComponentUpdate` will return `true`. This appears to be an avoidable re-render, which is what `hasRenderedEmoji` is implemented to prevent. 

There may be a different way to implement this (avoiding unnecessary re-renders), but this method requires using the state, as far as I can see.

Hopefully that makes sense, and I've not misunderstood the question! If you've got a preference on a different way to implement this I'd be happy to take a look and update the PR 👍

juliankrispel:
No that makes sense. Here’s a snag: What if there’s no connectivity - I
haven’t tested this but if the request won’t be retried as you say - then
those emojis will just stay blank - so what you really want is to set the
flag to true when the request has succeded - with the onLoad prop.

Mind implementing and testing this? If you need help lmk!

Thank you soooooooo much this is a really great pr! 👌👌👌

joshfarrant:
Good thinking, makes sense!

I'll take a look at switching this over to using `onLoad` shortly.

juliankrispel:
hey @joshfarrant how's it going - did you have a chance to play around with onLoad? ❤️

joshfarrant:
I did have a brief play around with it just after my last message, but it wasn't as quick a change as I initially thought.

I'm planning on taking another look at it at some point, but if anyone else wants to jump in and take a look before then feel free.

stevewillard:
I hope this PR gets merged soon. Thanks for all your work!

sibelius:
can you rebase?

joshfarrant:
Just wanted to post to say that this isn't something I'm likely going to get to (if you hadn't already guessed)!

If the PR is still valid and the codebase hasn't changed so much that it requires fully reimplementing then please someone else feel free to complete this PR, or modify my branch and make your own up-to-date PR.

PR #1031, opened 2018-02-02
Fix inline toolbar positioning when there's a positioned parent.

amannn:
## Checklist

## Use-case/Problem

This PR improves the positioning of the **inline toolbar**. The current implementation didn't work when there's a positioned parent (see #531).

## Implementation

Very similar to #1029.

The class `DraftEditor-root` is applied to the very root node of the editor by draft.js itself – this class allows to find that node. The side toolbar is a sibling of the editor root node, therefore we can position the toolbar by considering the offset of the editor node within it's next positioned parent. Note that if there's no explicitly positioned parent, then `body` will be taken into account (this is the case in the docs).

Note that we don't need the `getRelativeParent` helper as the browser offers a built in function [`offsetParent`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent).

Tested in Chrome, Firefox, Safari and IE11 (both with and without relative positioning).


draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -52,16 +39,32 @@ export default class Toolbar extends React.Component {
...
-      const relativeRect = (relativeParent || document.body).getBoundingClientRect();
-      const selectionRect = getVisibleSelectionRect(window);
 
+      // The editor root should be two levels above the node from
+      // `getEditorRef`. In case this changes in the future, we
+      // attempt to find the node dynamically by traversing upwards.
+      const editorRef = this.props.store.getItem('getEditorRef')();
+      if (!editorRef) return;
+      let editorRoot = editorRef.refs.editor;
+      while (editorRoot.className.indexOf('DraftEditor-root') === -1) {
```
juliankrispel: hm not sure about this tbh. Isn't there a way to get a parent that is absolutely positioned? I'd investigate that - because this is brittle

juliankrispel:
@amannn I'm a little unsure about the approach - but convinced of the need for the pr.

Can you please - add a story to test this problem. Unit testing this would be hard - but a story should do.

Cheers

amannn:
@juliankrispel Thanks for the review!

This is a screenshot from the editor that's used in the docs:

screen shot 2018-03-21 at 12 11 26

The highlighted node is the one returned from `getEditorRef()`. There's the [HTMLElement.offsetParent](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent) API which returns the closest parent that isn't positioned statically. However `DraftEditor-editorContainer` is positioned relatively as well. So we'd need `editorRoot.offsetParent.offsetParent` which I thought isn't that nice in case there will be additional elements in between in the future.

Draft.js always [sets the DraftEditor-root class name](https://github.com/facebook/draft-js/blob/master/src/component/base/DraftEditor.react.js#L321) on the container (not configurable by the user). So even if you don't use the supplied stylesheet, the class name should always be there. My code would break as soon as they change the name of that class though. I'm not sure if there's a better way? If the options are hardcoding the levels of (positioned) DOM nodes that need to be traversed upwards or the class name I thought I'd go with the latter.

What do you think?

Should I add the story as comment in the component code on how this can be tested? I could also change one of the examples in the docs to have a positioned parent.

amannn:
@juliankrispel Similar to the #1029 PR I've added a story here and the missing draft.js stylesheet.

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
```diff
@@ -52,16 +39,32 @@ export default class Toolbar extends React.Component {
...
       if (!this.toolbar) return;
-      const relativeParent = getRelativeParent(this.toolbar.parentElement);
-      const toolbarHeight = this.toolbar.clientHeight;
-      const relativeRect = (relativeParent || document.body).getBoundingClientRect();
-      const selectionRect = getVisibleSelectionRect(window);
 
+      // The editor root should be two levels above the node from
+      // `getEditorRef`. In case this changes in the future, we
+      // attempt to find the node dynamically by traversing upwards.
+      const editorRef = this.props.store.getItem('getEditorRef')();
```
juliankrispel: we already a method (above) for finding the relative parent. will that not do the job?

juliankrispel:
I'll give this another proper look tomorrow, it looks a lot better than other attempts so far!

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
amannn: We're not looking for the next enclosing positioned parent, but rather for the DOM node with the class `DraftEditor-root` – so I think we don't need it anymore. Btw. that method does the same as simply calling `node.offsetParent` 🙂

amannn:
Thanks! I just synced with master and added that React 15 compatibility check.

draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
juliankrispel: before doing any more work on this, mind having a look at https://github.com/draft-js-plugins/draft-js-plugins/pull/1018, there's some good stuff in there. We have two prs now that fix this problem 😨

amannn: That's an interesting PR! I think it's doing a different thing though. That PR adds support for a smart positioning of the inline toolbar and its caret. As far as I can tell the positioning within a relative parent isn't addressed there. Maybe we can merge this PR and rebase the other PR on top of this to use the relevant parts?

juliankrispel:
Ok phew 😅 - that’s good stuff

draft-js-inline-toolbar-plugin/src/index.js
```diff
@@ -35,9 +35,10 @@ export default (config = {}) => {
   };
 
   return {
-    initialize: ({ getEditorState, setEditorState }) => {
+    initialize: ({ getEditorState, setEditorState, getEditorRef }) => {
       store.updateItem('getEditorState', getEditorState);
       store.updateItem('setEditorState', setEditorState);
+      store.updateItem('getEditorRef', getEditorRef);
```
juliankrispel: ugh it makes me sad that we have our own abstraction for state management in plugins...

juliankrispel:
ok sweet let's merge

gnestor:
I don't know how, but this broke toolbar positioning for me! I needed to downgrade to 2.0.1 to resolve.

My DOM structure looks like:

```
div.content (absolute)
  div.doc (relative)
    Editor
  InlineToolbar
```

What do I need to change in order for things to work with this PR?

amannn:
@gnestor I'm sorry for that. The inline toolbar assumes that the editor and the inline toolbar have the same positioned parent.

So I think this would work for you:

```
div.content (absolute)
  div.doc (relative)
    Editor
    InlineToolbar
```

gnestor:
That worked! Thanks @amannn!

Issue #1032, opened 2018-02-05
blockStyleFn warning triggered by Image with Draft-js 0.10.5


HenrikBechmann:
Adding this to the Image/index.js transpiled code suppresses the warning
```
          blockStyleFn = otherProps.blockStyleFn, // changed by Henrik
          elementProps = _objectWithoutProperties(otherProps, ['blockProps', 'customStyleMap', 'customStyleFn', 'decorator', 'forceSelection', 'offsetKey', 'selection', 'tree', 'contentState','blockStyleFn']); // blockStyleFn added by Henrik
```
iow `blockStyleFn` needs to be added to the list of suppressed props at Image/index.cs at L13 of https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-image-plugin/src/Image/index.js

Thanks to `@JohnM` in the draft-js-plugins slack channel for the solution

freedomlang:
fixed !

PR #1033, opened 2018-02-06
Omit blockStyleFn prop from list of elementProps in image plugin

JohnMaguir:
## Checklist

## Use-case/Problem

https://github.com/draft-js-plugins/draft-js-plugins/issues/1032

When using draft-js version 0.10.5 an extra prop `blockStyleFn` is passed and needs to be omitted like the others otherwise it is assigned as an elementProp and causes the error
`Unknown prop 'blockStyleFn' on  tag`

## Implementation
Solution is to add this `blockStyleFn` prop to the omitted props list so it is not assigned as an elementProp


juliankrispel:
thanks @JohnMaguir for the fix!

PR #1049, opened 2018-03-14
fix: import draft-js-static-toolbar-plugin style

Hongbo-Miao:
I think it is wrong. It should import static toolbar style instead of inline toolbar style.

Please double confirm. Thanks


Hongbo-Miao:
Duplicated with https://github.com/draft-js-plugins/draft-js-plugins/pull/1046

Hongbo-Miao:
@freedomlang what is the point.? I already closed this..
If you can help, please review https://github.com/draft-js-plugins/draft-js-plugins/pull/1046

PR #1060, opened 2018-04-05
Fixed overrideContent feature in InlineToolbar

### LINKED ISSUE #967: Headline buttons are broken in the Custom Inline Toolbar Example
These buttons "HeadlineOneButton", "HeadlineTwoButton", "HeadlineThreeButton" are broken in the [Custom Inline Toolbar Example](https://www.draft-js-plugins.com/plugin/inline-toolbar).

Tested on Chrome and Firefox.

juliankrispel:
## Checklist

## Use-case/Problem

fixes #967

### DIFF
--- a/docs/client/components/pages/InlineToolbar/CustomInlineToolbarEditor/index.js
+++ b/docs/client/components/pages/InlineToolbar/CustomInlineToolbarEditor/index.js
@@ -47,8 +47,9 @@ class HeadlinesPicker extends Component {
 }
 
 class HeadlinesButton extends Component {
-  // When using a click event inside overriden content we have
-  // to cancel click events from the toolbar itself unfortunately
+  // When using a click event inside overridden content, mouse down
+  // events needs to be prevented so the focus stays in the editor
+  // and the toolbar remains visible  onMouseDown = (event) => event.preventDefault()
   onMouseDown = (event) => event.preventDefault()
 
   onClick = () =>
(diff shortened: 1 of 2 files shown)


juliankrispel:
thanks for the review @amannn :heart: I can't remember exactly - but it was another fix (via setTimeout)...

docs/client/components/pages/InlineToolbar/CustomInlineToolbarEditor/index.js
```diff
@@ -47,6 +47,10 @@ class HeadlinesPicker extends Component {
 }
 
 class HeadlinesButton extends Component {
+  // When using a click event inside overriden content we have
+  // to cancel click events from the toolbar itself unfortunately
```
amannn: There's a small typo: "overridden"

Since this comment is in the docs and is supposed to educate users, what about something like: "When using a click event inside overridden content, mouse down events needs to be prevented so the focus stays in the editor and the toolbar remains visible."?

amannn: @juliankrispel Don't you think this should be updated?

juliankrispel: @amann I did :) - you're looking at the outdated version ;)

juliankrispel: @amannn are you sure you don't want to join our fortnightly meetings?  ;)

amannn: Ah – sorry, my bad! Haha, I'll think it over ;-)

Issue #1073, opened 2018-04-23
Mentions plugin: the "focused" option should have an `aria-selected="true"` attribute


afercia:
The combobox ARIA widget is a bit complicated. There are now two different valid patterns, the "traditional" one from ARIA 1.0 and the new one from ARIA 1.1. The differences between the two patterns are very well described in the [ARIA Authoring Practices 1.1](https://www.w3.org/TR/wai-aria-practices-1.1/#combobox).

Regardless, in both patterns the "highlighted" option should get an `aria-selected="true"` attribute which is missing in the mentions plugin implementation.

References:

ARIA 1.0
https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-selected
> Used in Roles: ... option 

ARIA 1.1
https://www.w3.org/TR/wai-aria-1.1/#aria-selected
> Used in Roles: ... option 

ARIA Authoring Practices 1.1
See under the combobox "WAI-ARIA Roles, States, and Properties"
https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties-5
> In a combobox with a listbox, grid, or tree popup, when a suggested value is visually indicated as the currently selected value, the option, gridcell, row, or treeitem containing that value has aria-selected set to true.

I've often tested combobox widgets and helped building a few ones. Based on that, I've noticed that Windows screen readers, typically NVDA (on Firefox) and JAWS (on IE11) correctly read out the highlighted option even without `aria-selected="true"`.

The problem is with Safari + VoiceOver. Even with the current versions:
- Safari 11.1 (13605.1.33.1.2)
- VoiceOver utility 8 (562.18.4)

the highlighted option is not read out at all. Absolutely nothing: resounding silence 🙂 

In all the combobox widgets I've tested, I've noticed Safari and VoiceOver really need `aria-selected="true"`. You can see it in action also in the 5 examples in the ARIA Authoring Practices 1.1: https://www.w3.org/TR/wai-aria-practices-1.1/#examples-1

Please notice the highlighted option has `aria-selected="true"` while this attribute is omitted on the other options.

However, to actually test this there's one more problem to solve: there's also the need for an `aria-owns` attribute. I've read https://github.com/facebook/draft-js/commit/7f0cab28386ec4bde8ec6289377bff9e53cd019b and I'll open a separate issue for that. In a few words, to test this I've used Draft.js v0.10.1 to bring back support for `aria-owns`.

[...]

PR #1075, opened 2018-04-24
Mentions and Emoji: Use aria-selected for the focused listbox option

afercia:
First time submitting a PR to your repo, please let me know if I'm doing anything wrong 🙂 

Seems to me there's no changelog for the Mentions plugin, so just thought to summarize the changelog entry here:
- Add `aria-selected="true"` for the Mentions and Emojis listbox focused option.

## Checklist

- [ x] Fix any eslint errors that occur

## Use-case/Problem

Based on the ARIA spec and the ARIA Authoring Practices, `aria-selected="true"` is a required ARIA state for the suggested option that is visually indicated as the currently selected option. For more details please refer to the related issue https://github.com/draft-js-plugins/draft-js-plugins/issues/1073

## Implementation

Just using the existing `isFocused` prop to render an `aria-selected="true"` attribute. Avoiding to render the attribute when is not focused.

To test, please check the Mentions or Emojis plugins and inspect the rendered DOM in your browser's console. When highlighting one of the options in the listbox (whether with mouse or keyboard) the highlighted option should get an `aria-selected="true"`.

Fixes https://github.com/draft-js-plugins/draft-js-plugins/issues/1073

### DIFF
--- a/draft-js-emoji-plugin/CHANGELOG.md
+++ b/draft-js-emoji-plugin/CHANGELOG.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - render props.children instead, fixes editing behaviour
 
 ## To Be Released
+- Added `aria-selected="true"` for the suggestions listbox focused option.
 
 ### Fixed
 - Added type="button" to EmojiSelect button to prevent submitting forms, thanks to @alanwflood [for the pr](https://github.com/draft-js-plugins/draft-js-plugins/pull/829)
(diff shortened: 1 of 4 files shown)


juliankrispel:
@afercia the changelog is here btw https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/CHANGELOG.md

afercia:
Uh, my editor didn't find that file with a file search for `changelog`, while it did find the Emoji one for example 🤔

Issue #1072, opened 2018-04-23
aria-expanded is always true: its value should be passed as boolean not as string


afercia:
At the company I work for we've implemented our first draft.js thingy with the mentions plugin, and it's awesome to see it in place in our apps.

However, as an accessibility specialist, I was wondering why the `aria-expanded` HTML attribute is always set to `true` regardless if the "dropdown" is open or closed.

Turns out `draft.js` has changed something since version `0.10.2` and now uses a pattern to make sure `aria-expanded` is only set when there's a `role="combobox"`. That's perfectly fine, but the actual pattern is the following one:

`var ariaExpanded = ariaRole === 'combobox' ? !!this.props.ariaExpanded : null;`

Please notice the double `!!`. It happens that the mentions plugin correctly switches the value on `openDropdown` and `closeDropdown` but then it passes the values as strings `'false'` and `'true'`...

Now, the double `!!` in draft.js casts any non-empty string to boolean `true`... that's the reason why the HTML actual set value is always `true`.

I guess this applies to all the plugins that use `aria-expanded` and pass strings, as far as I see also the Emoji plugin and maybe other ones?

juliankrispel:
awesome! I'll get on it if you don't first.

Also if you feel like opening a pr yourself but don't know where to start, feel free to get in touch on the draft-js slack channel

afercia:
@juliankrispel I have a local branch ready but I'm a bit busy in the next few hours, please do feel free to go head if you have a chance.

juliankrispel:
@afercia I wouldn't be ready until later this week - have loads of other stuff on :D

PR #1077, opened 2018-04-24
Mentions and Emojis: Pass the aria-expanded values as boolean.

afercia:
New to contributing to your repo, please let me know if I'm doing anything wrong 🙂

Seems to me there's no changelog for the Mentions plugin, so just thought to summarize the changelog entry here:

- Mentions and Emojis plugin: Pass the combobox `aria-expanded` values as boolean.

## Checklist

- [ x] Fix any eslint errors that occur

## Use-case/Problem

`draft.js` has changed something since version 0.10.2 and now uses a pattern to make sure `aria-expanded` is only set when there's a `role="combobox"`. That's perfectly fine, but the actual pattern is the following one:

`var ariaExpanded = ariaRole === 'combobox' ? !!this.props.ariaExpanded : null;`

Thus, if a value is passed as string, it's always converted to boolean `true`. Actually, the `aria-expanded` value is always true, even when the listbox dropdown is open. The values need now to be passed as boolean true/false.

## Implementation

Just changed all the passed string `'true'` / `'false'` to boolean.

To test, please check the mentions and emojis plugins: the `aria-expanded` value should be initially false, when the listbox dropdown opens, should change to true and so on.

Fixes https://github.com/draft-js-plugins/draft-js-plugins/issues/1072

### DIFF
--- a/draft-js-emoji-plugin/src/components/EmojiSuggestions/index.js
+++ b/draft-js-emoji-plugin/src/components/EmojiSuggestions/index.js
@@ -237,7 +237,7 @@ export default class EmojiSuggestions extends Component {
     this.props.ariaProps.ariaActiveDescendantID = descendant;
     this.props.ariaProps.ariaOwneeID = `emojis-list-${this.key}`;
     this.props.ariaProps.ariaHasPopup = 'true';
-    this.props.ariaProps.ariaExpanded = 'true';
+    this.props.ariaProps.ariaExpanded = true;
     this.setState({
       isActive: true,
     });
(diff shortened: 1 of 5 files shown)


juliankrispel:
although if you'd rather add flowtypes than tests that'd probably make even more sense :D

juliankrispel:
but if you don't have time for either, I'd be ok with merging this as is and doing it myself later

afercia:
@juliankrispel I really don't know anything about flow/flowtypes (well I've read something but that's not enough) I'm just an accessibility person 🙂 I'd greatly appreciate if you could do that, when you have a chance. Thanks!

juliankrispel:
Ok no probs! In any case your contribution is highly appreciated ☺️ ! Do
you have a twitter handle?

afercia:
> Do you have a twitter handle?

Same as the GitHub one ;)

PR #1078, opened 2018-04-24
Editor 3.0

juliankrispel:
Changes:

- [x] flow types everywhere (almost done)
- [x] updated tests
- [x] refactored hook resolution (no more magic matching patterns)
- [x] export flow types in package
- [x] decorator resolution (deprecate editorState.decorator)
- [x] add build step for flow types

## Checklist


mxstbr:
Let me know when an alpha is out and I'll test it!

juliankrispel:
@mxstbr I've added a story `DynamicPluginEditor` if you can test it in there and try break the thing that'd be awesome!

draft-js-plugins-editor/src/Editor/__test__/index.js
```diff
@@ -671,20 +684,5 @@ describe('Editor', () => {
       expect(customPluginDecorator).has.been.called();
       expect(decoratorStrategy).has.been.called();
     });
-
-    it('reassigns decorators to editorState when props are updated with naked editorState', (done) => {
```
juliankrispel: this is now deprecated

draft-js-plugins-editor/src/Editor/MultiDecorator.js
```diff
@@ -46,11 +46,8 @@ class MultiDecorator {
 
   /**
    * Return props to render a decoration
-   *
-   * @param {String} key
-   * @return {Object}
    */
-  getPropsForKey(key) {
+  getPropsForKey(key: string): ?Object {
```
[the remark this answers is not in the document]
juliankrispel: I'm actually copying draft-js internal types here - to be consistent

draft-js-plugins-editor/package.json
```diff
@@ -23,15 +23,19 @@
...
     "prepublish": "npm run build"
   },
   "license": "MIT",
   "dependencies": {
     "decorate-component-with-props": "^1.0.2",
     "find-with-regex": "^1.1.3",
+    "flow-copy-source": "^1.3.0",
     "immutable": "~3.7.4",
     "prop-types": "^15.5.8",
+    "shallowequal": "^1.0.2",
```
nikgraf: I'm wondering if this dependency is necessary. I'm not sure if React or Draft has a dependency on fbjs (https://www.npmjs.com/package/fbjs). If that's the case we can use the shallowequal from  that without introducing duplicated code.

draft-js-plugins-editor/src/Editor/createCompositeDecorator.js
```diff
@@ -1,18 +1,23 @@
...
+  decorators: List<DraftDecorator>,
+  getEditorState: () => EditorState,
+  setEditorState: (editorState: EditorState) => void) => {
+  const convertedDecorators = decorators
     .map((decorator) => ({
       ...decorator,
       component: decorateComponentWithProps(decorator.component, { getEditorState, setEditorState }),
-    }))
-    .toJS();
+    })).toArray();
```
nikgraf: I'm not 100% sure, but afaik toJS does a deep conversion and toArray only a shallow one. Is this desired here?

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -74,62 +92,88 @@ class PluginEditor extends Component {
...
   // Cycle through the plugins, changing the editor state with what the plugins
   // changed (or didn't)
-  onChange = (editorState) => {
+  onChange = (editorState: EditorState) => {
     let newEditorState = editorState;
+    const pluginMethods = this.getPluginMethods();
     this.resolvePlugins().forEach((plugin) => {
-      if (plugin.onChange) {
-        newEditorState = plugin.onChange(newEditorState, this.getPluginMethods());
+      if (plugin.onChange != null && newEditorState != null) {
```
nikgraf: so if I call onChange with `null` it will not trigger the onChange callbacks for plugins. Is this desired behaviour?

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -74,62 +92,88 @@ class PluginEditor extends Component {
...
+        newEditorState = plugin.onChange(newEditorState, pluginMethods);
       }
     });
 
     if (this.props.onChange) {
```
nikgraf: Especially compared to above we have different behavior to plugins since this onChange would still fire even though the new editor state is `null`

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -139,165 +183,197 @@ class PluginEditor extends Component {
...
+  createHandleHooks = (methodName: string, plugins: Array<Plugin|EditorProps>) => (firstArg: any, ...args:any):string => {
+    const newArgs = [firstArg, ...args, this.getPluginMethods()];
 
-    return plugins.some((plugin) =>
-      typeof plugin[methodName] === 'function'
-        && plugin[methodName](...newArgs) === 'handled'
-    ) ? 'handled' : 'not-handled';
+    return plugins.some((plugin) => {
+      const method = plugin[methodName];
+      return method != null && method(...newArgs) === 'handled';
```
nikgraf: why not use typeof here? Not null could still mean it's not a function.

stories/DynamicPluginsEditor/index.js
```diff
@@ -0,0 +1,77 @@
+import React, { Component } from 'react';
```
nikgraf: cool feature the DynamicPlugins!

package.json
```diff
@@ -103,7 +104,7 @@
     "emojione": "^2.2.7",
     "eslint-plugin-import": "^2.2.0",
     "find-with-regex": "^1.1.3",
-    "immutable": "^3.8.1",
+    "immutable": "4.0.0-rc.9",
```
nikgraf: does it make sense to wait until immutable actually released the 4.0?

draft-js-plugins-editor/src/Editor/resolveDecorators.js
```diff
@@ -1,27 +1,47 @@
...
 
 // Return true if decorator implements the DraftDecoratorType interface
 // @see https://github.com/facebook/draft-js/blob/master/src/model/decorators/DraftDecoratorType.js
-const decoratorIsCustom = (decorator) => typeof decorator.getDecorations === 'function' &&
-  typeof decorator.getComponentForKey === 'function' &&
-  typeof decorator.getPropsForKey === 'function';
-
+// $FlowFixMe
+const decoratorIsCustom = (decorator: DraftDecorator|DraftDecoratorType): boolean => (
+  decorator.getDecorations != null &&
```
nikgraf: is this flow related? checking if it's a function was on purpose back then :)

draft-js-plugins-editor/src/Editor/index.js
```diff
@@ -139,165 +183,197 @@ class PluginEditor extends Component {
...
-    const plugins = [this.props, ...this.resolvePlugins()];
+    const plugins = this.resolvePlugins();
+
     plugins.forEach((plugin) => {
       if (typeof plugin.getAccessibilityProps !== 'function') return;
       const props = plugin.getAccessibilityProps();
       const popupProps = {};
 
-      if (accessibilityProps.ariaHasPopup === undefined) {
+      if (accessibilityProps.ariaHasPopup == null) {
```
nikgraf: Is this every true? I guess most of the times it's undefined and that's why we had the check for it?!

draft-js-plugins-editor/package.json
mxstbr: I'm pretty sure React depends on fbjs, or at least it did a while ago

draft-js-plugins-editor/src/Editor/index.js
mxstbr: `value == null` is a shorthand for `value === null || value === undefined` :+1: #javascript

draft-js-plugins-editor/package.json
juliankrispel: it's not actually, I had used it and then forgot to take it out ....

juliankrispel:
@nikgraf thx for the review - yes a bunch of the changes were to satisfy flow checks to be fair. I'll see if I can fix that to some degree. It should really give us more security rather than less...

draft-js-plugins-editor/src/Editor/index.js
nikgraf: ah missed the `==` … can you add that as a comment on top of it to make it beginner friendly and people like me constantly forgetting this :)

nikgraf:
I'm approving in the meantime. Ping me if you want to have another review after more changes :)

sibelius:
This is great

I'm gonna close this until we refactor most of the code to bring new changes

PR #1109, opened 2018-05-11
draft-js-inline-toolbar-plugin: do not show inline toolbar when selection is in code-block

andrewfan:

## Checklist

## Use-case/Problem

When the user selects text inside code block inline toolbar should not appear (`Slack post` redactor has the same functionality)

## Implementation

Just check current block type


markolofsen:
Could you implement Inline Toolbar for nextjs? I did it, but toolbar isn't working.
pro

sibelius:
can you rebase and add some tests

PR #1121, opened 2018-06-01
remove google analytics in order to avoid a cookie notice and mailchimp signup form which we deprecated (gdpr)

nikgraf:
As discussed with Julian we are dropping Google Analytics from the side. In addition we dropped the mailchimp mailing list. 

## Checklist


juliankrispel:
thx @nikgraf <3

PR #1132, opened 2018-06-25
Using sprites

creativelikeadog:
inspired by: https://github.com/sevikon/draft-js-plugins

## Use-case/Problem

Loading images makes the component slow

## Implementation

Enable use of sprites


udanpe:
@creativelikeadog what about pr to implementation emoji-mart?
https://github.com/missive/emoji-mart

pmehta18:
@juliankrispel Can you please merge this PR so that this feature can be used by us as well instead of cloning the plugin and then adding the code?
Thanks in advance!

sibelius:
I think using https://github.com/missive/emoji-mart would be simpler

PR #1133, opened 2018-06-26
Add isOpen prop

udanpe:
Pls add isOpen prop to EmojiSelect component


sibelius:
what is the use case?

PR #1135, opened 2018-07-03
Added support for custom method for creating custom entities

dotansimha:
## Checklist

## Use-case/Problem

My use case requires to use mentions and it's logic, but after a mention has been picked - I need to add it as custom entity I have in my editor, and apply some custom logic to handle it's creation in DraftJS.

## Implementation

Added new prop `addMentionToEditor` to `MentionSuggestions` that allow the developer to provide an alternative method for `addMention`.


sibelius:
can you rebase?

PR #1142, opened 2018-07-06
Fix popover positioning when it spills out of screen

phanindra48:
## Checklist

## Use-case/Problem
Suggests popover is spilling out of the viewport

## Implementation
Using device width to push suggestions list popover just little left so as to make it completely visible

## Demo
fixed-popover


sibelius:
can you rebase?

PR #1145, opened 2018-07-13
Fix offset top for Toolbar

Archakov06:
If remove the editorRoot offset, then everything is okay.

## Demo

jul-13-2018 20-44-08


yhjor:
can anyone help merge and publish this PR?

freedomlang:
@Archakov06 can not reproduce the issue you mentioned above, can you provide your dom structure so that we can check and reproduce it. BTW, following the pr template would be nice  and don't forget the changelog

sibelius:
can you rebase and provide a codesandbox?

PR #1153, opened 2018-07-24
[emoji-plugin] lazy load emojis

vybu:
This is an update of https://github.com/draft-js-plugins/draft-js-plugins/pull/1030 PR which never got merged.

## Checklist

## Use-case/Problem

So currently emoji plugin loads 1.3k emojis initially, which is just insane.

With this change it initially loads only 225 (first emoji group), then after the emoji popup is opened - second group of emojis loads. When user navigates to other groups it lazy loads current viewed group + next group to give best user experience. So if user doesn't scroll very fast his currently viewed group should be always loaded.

I know there was discussion in  https://github.com/draft-js-plugins/draft-js-plugins/pull/1030 about checking if image loaded with `onLoad` and only then preventing re-rendering of a group. But to me this seems not related to issue of lazy loading. Current implementation (as in production code) also doesn't handle case when image fails to load. This can be solved with separate PR. One solution would be to add `onError` handler to `` and try to reload.

## Demo
emoji-lazy-load

### DIFF
--- a/draft-js-emoji-plugin/src/components/EmojiSelect/Popover/Groups/Group/index.js
+++ b/draft-js-emoji-plugin/src/components/EmojiSelect/Popover/Groups/Group/index.js
@@ -22,7 +22,7 @@ export default class Group extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (nextProps.isActiveGroup) {
+    if (nextProps.isActive) {
       this.setState({ hasRenderedEmoji: true });
     }
   }
--- a/draft-js-emoji-plugin/src/components/EmojiSelect/Popover/Groups/Group/index.js
+++ b/draft-js-emoji-plugin/src/components/EmojiSelect/Popover/Groups/Group/index.js
@@ -14,9 +14,26 @@ export default class Group extends Component {
     onEmojiSelect: PropTypes.func.isRequired,
     onEmojiMouseDown: PropTypes.func.isRequired,
     useNativeArt: PropTypes.bool,
+    isActive: PropTypes.bool,
   };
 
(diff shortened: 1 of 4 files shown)


juliankrispel:
I'm aware that this is something people have been asking for repeatedly, will do my best to help you get this in!

draft-js-emoji-plugin/src/components/EmojiSelect/Popover/Groups/Group/index.js
```diff
@@ -14,9 +14,26 @@ export default class Group extends Component {
...
 
-  shouldComponentUpdate = () => false;
+  state = {
+    hasRenderedEmoji: false,
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.isActive) {
+      this.setState({ hasRenderedEmoji: true });
+    }
```
juliankrispel: `willReceiveProps` is marked as `UNSAFE` in react 16 and will be deprecated

@vybu would you mind updating this to use [`getDerivedStateFromProps`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) instead.

vybu:
I used `componentDidUpdate` instead, it achieves same goal and maybe makes even more sense.
Because we want to prevent further render of component after it has been rendered with emojis. Also it works for older and most recent react versions

Sidenote - if I were to use ` getDerivedStateFromProps` it wouldn't work because current react version for this project is 15.5 and this method in that version doesn't exist. Also what about backwards compatibility? I guess if you wanted to use this new method and be backwards compatible to older react versions you would have to use both `componetWillReceiveProps` and `getDerivedStateFromProps`.

PR #1165, opened 2018-08-13
[mention plugin] Fix getWordAt failure when mentionTrigger is set as empty

danyu:
## Checklist

- Fixed getWordAt failure when mentrionTrigger is empty
- Added unit test.

## Use-case/Problem

I am using mention plugin and set **‘mentionTrigger’ as empty**. By doing that, the mention pops up automatically as typing without keying in trigger letter, which works pretty good.
BUT some issues came up lately when I was adding mention **before existing mentions**:
1. First, the suggestions filtering did not work; the suggestions was not narrowed down as typing
2. Second, when I selected a suggestion, wrong portion of text got replaced as mention.
Note: Above scenario works when mentionTrigger is not empty.

## Implementation

Check if cursor is already pointing to end of word being typed, which is whitespace(' '). If Yes, do not move cursor right. If NO, move 1 position right (pos+1) as it is now.

## Demo of issue
This is example of custom mention. Set mentionTrigger as ''.
To Add mention '**julian**' before mention '**max**'

**1. Filtering not working when add mention before existing mention**
mention-filtering-not-working

**2. Replace text not working when mention is selected.**
mention-replace-not-working

### DIFF
--- a/draft-js-mention-plugin/src/utils/getWordAt.js
+++ b/draft-js-mention-plugin/src/utils/getWordAt.js
@@ -5,7 +5,8 @@ const getWordAt = (string, position) => {
   const pos = Number(position) >>> 0;
 
   // Search for the word's beginning and end.
-  const left = str.slice(0, pos + 1).search(/\S+$/);
+  // Do NOT adjust pos if it already at the end of word.(Happens when mentiontrigger is empty(''))
+  const left = str.slice(0, str.charAt(pos) === ' ' ? pos : pos + 1).search(/\S+$/);
   const right = str.slice(pos).search(/\s/);
 
   // The last word in the string is a special case.
--- a/draft-js-mention-plugin/src/utils/getWordAt.js
+++ b/draft-js-mention-plugin/src/utils/getWordAt.js
@@ -2,11 +2,16 @@ const getWordAt = (string, position) => {
   // Perform type conversions.
   const str = String(string);
   // eslint-disable-next-line no-bitwise
-  const pos = Number(position) >>> 0;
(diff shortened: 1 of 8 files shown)


draft-js-mention-plugin/src/utils/getWordAt.js
```diff
@@ -5,7 +5,8 @@ const getWordAt = (string, position) => {
   const pos = Number(position) >>> 0;
 
   // Search for the word's beginning and end.
-  const left = str.slice(0, pos + 1).search(/\S+$/);
+  // Do NOT adjust pos if it already at the end of word.(Happens when mentiontrigger is empty(''))
+  const left = str.slice(0, str.charAt(pos) === ' ' ? pos : pos + 1).search(/\S+$/);
```
juliankrispel: hi @danyu thx for this pr! Could you make this an if statement rather a ternary expression, this is more compact but I think it should be a little easier to read (for the benefit of us all ❤️ )

danyu: Sure, does following look better: :sparkles:

```javascript
  // Pos goes to end of word already when mentiontrigger is empty(''),
  // so make sure adjusting pos ONLY when cursor pos is NOT at the end of word.
  if (str.charAt(pos) !== ' ') {
    pos += 1;
  }

  // Search for the word's beginning and end.
  const left = str.slice(0, pos).search(/\S+$/);
```

juliankrispel:
this issue seems to have been fixed in another pr, thanks so much still!

PR #1168, opened 2018-08-15
Change unsafe componentWillMount to componentDidMount lifecycle hook.

### LINKED ISSUE #1167: [draftjs-focus-plugin]: bug with lifecycle hooks
## Problem

In `createDecorator`, please rename the `componentWillMount` call to `componentDidMount`.

Reason:

`componentWillMount` is [unsafe and will be deprecated](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount).

What happens in my case is that an old instance is calling `componentWillUnmount` just before the new instances `componentDidMount` is called. This removes the keys from `blockKeyStore` just after rendering.. See debugger gif below.

createdecorator

The sequence for instances 1, 2 is:
1 willMount
1 render
1 didMount
2 willMount
2 render
1 willUnmount
2 didMount

@juliankrispel

bebbi:
This closes #1167

### DIFF
--- a/draft-js-focus-plugin/CHANGELOG.md
+++ b/draft-js-focus-plugin/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 2.2.0
+- Fix blockKeyStore behaviour so it is not accidentally deleted before the component has finished mounting.
+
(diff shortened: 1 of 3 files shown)


juliankrispel:
@bebbi awesome! Please add to changelog and update package version 👍

bebbi:
@juliankrispel done 👍

Issue #1171, opened 2018-08-22
[draft-js-image-plugin] Entity type not backwards compatible


dobesv:
## Use-case/Problem

In https://github.com/draft-js-plugins/draft-js-plugins/commit/0c98f97537dc937e1dc4bce15d26386f52e9b008#diff-7c6e0b8e32c6156fadff3f0170a0a420 the plugin was changed to emit and expect an all-caps entity type `IMAGE` instead of `image`.  However, we were using this plugin before and storing the text in the old format.  This change is not backwards-compatible so those old image no longer show up.

freedomlang:
fixed!

PR #1172, opened 2018-08-22
Support old lower-case entity type `image` for backwards compatibility

dobesv:
See https://github.com/draft-js-plugins/draft-js-plugins/issues/1171

## Implementation

This will match on both the old `image` and new `IMAGE` entity types.


dobesv:
@juliankrispel OK I did that.

juliankrispel:
@dobesv thanks! Hey btw, if you'd like to be involved on a continuous basis, we run a meeting every two weeks. Currently the team is small but I'm looking to get more people involved. Just hit me up on the draft js slack channel if you're interested!

Issue #1173, opened 2018-08-22
Custom component for CustomCounter


cedricmessiant:
## Use-case/Problem

I would like to be able to use my own component for rendering when I use the `CustomCounter`. For example, I need to render an internationalized string that uses `count` as a param:

```
<div className="annotation margin-xs">
  <Translate value="debate.remaining_x_characters" nbCharacters={count} />
</div>
```

(the count may be anywhere in the string depending on the language.

freedomlang:
@cedricmessiant the `countFunction` for `CustomCounter` shoud do, you can return a string whatever a language!

PR #1174, opened 2018-08-22
Allow user to use a custom component for CustomCounter

cedricmessiant:
## Use-case/Problem

I would like to be able to use my own component for rendering when I use the CustomCounter. For example, I need to render an internationalized string that uses count as a param (Refs #1173 )

## Implementation

I added an optional `component` prop to the `CustomCounter` component. When it is defined, this component is rendered instead of the usual `<span>`.


cedricmessiant:
(well I guess that we could do it for all counters in fact)

freedomlang:
@cedricmessiant I'll close this pr! Cause it can be done by `countFunction `(as we've been discuss in #1173 ) ❤️

PR #1186, opened 2018-08-29
Fix anchor plugin getEditorState undefined issue

iskWang:
Ref. https://codesandbox.io/s/93p6o47wlp

### DIFF
--- a/draft-js-anchor-plugin/src/components/LinkButton/index.js
+++ b/draft-js-anchor-plugin/src/components/LinkButton/index.js
@@ -27,10 +27,6 @@ export default class LinkButton extends Component {
 
   render() {
     const { theme, onRemoveLinkAtSelection } = this.props;
-    const className = hasLinkSelected
-      ? unionClassNames(theme.button, theme.active)
-      : theme.button;
-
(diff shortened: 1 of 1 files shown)


freedomlang:
@iskWang Not sure if anchor plugin support static-toolbar. I just found that description in online [docs](https://www.draft-js-plugins.com/plugin/anchor)


@juliankrispel Is archor plugin supporsed to support static-toolbar? Or just inline-toolbar?

juliankrispel:
@iskWang first of all thanks for your contrubution! Unfortunately the codepen doesn't work for me I'm afraid

image

If you've made the anchor plugin work with the static toolbar, could you please add a story to the pr for this particular case (or modify an existing one)

Cheers

iskWang:
Thanks @freedomlang  @juliankrispel reply , It's my error screenshot.
And I am not sure how to release "LinkButton" this plugin.
default

P.s: If you comment out [Codesandbox](https://codesandbox.io/s/93p6o47wlp) `linkPlugin.LinkButton` (via line 37) will working.

sibelius:
can you explain more about it?

PR #1191, opened 2018-09-02
[Emoji plugn] Custom Emoji Feature

icarusysuper:
This is a PR that enables us to develop custom Emoji Component.

## Checklist

## Use-case/Problem

The browser and server will carry a heavy burden when  emoji plugin use 1.3k http requests to load the emoji images. 

Developers can implement a custom Emoji Component to solve the above problem. For example,  the custom Emoji Component can accept base64 emoji image data, then display it.


juliankrispel:
I like this, can you please update the docs for this?

PR #1197, opened 2018-09-05
SideToolbar: Add scroll support

vkolova:
## Checklist

## Use-case/Problem
Plugin is ignoring scroll value when editor is in scroll container.

## Implementation
Sum the scrollTop values of all nodes in editor and remove it from original value.

### DIFF
--- a/docs/client/components/pages/SideToolbar/SimpleSideToolbarEditorWithScroll/editorStyles.css
+++ b/docs/client/components/pages/SideToolbar/SimpleSideToolbarEditorWithScroll/editorStyles.css
@@ -1,0 +1,19 @@
+.editor {
+  box-sizing: border-box;
+  border: 1px solid #ddd;
+  cursor: text;
+  padding: 16px;
+  border-radius: 2px;
+  margin-bottom: 2em;
+  box-shadow: inset 0px 1px 8px -3px #ABABAB;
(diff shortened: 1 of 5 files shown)


juliankrispel:
@vkolova sorry, not adding a prop here, the API is frozen I'm afraid. Can you think of another way to fix it? Bear in mind that the sidebar plugin also has a render prop so you can do stuff like this:

```
        <SideToolbar>
          {
            // may be use React.Fragment instead of div to improve perfomance after React 16
            (externalProps) => (
              <div>
                <HeadlineOneButton {...externalProps} />
                <HeadlineTwoButton {...externalProps} />
                <BlockquoteButton {...externalProps} />
                <CodeBlockButton {...externalProps} />
              </div>
            )
          }
        </SideToolbar>
```

vkolova:
@juliankrispel, unfortunately that's the only way I can think of....

G-Rath:
@juliankrispel Then I guess you can close the issue I made a couple of days ago (#1211), as I'm fairly confident the only reasonable way to solve this is with a new prop.

The only alternative I can think of would be to calculate the positioning of the toolbar in our own components, which would require basically doing everything the toolbar is already doing in regards to positioning calculations, defeating the point of using a package.

PR #1200, opened 2018-09-12
feat(resizeable): options to define initial sizes

### LINKED ISSUE #1166: [resizeable-plugin] When setting the options horizontal and vertical to "absolute", the initial size of the image is not respected
## Use-case/Problem

I'm using draft-js-plugins in an e-mail editor, so the and i'm using the following packages
* `"draft-js": "0.10.4"`
* `"draft-js-focus-plugin": "^2.0.1"`
* `"draft-js-image-plugin": "^2.0.1"`
* `"draft-js-plugins-editor": "^2.0.4"`
* `"draft-js-resizeable-plugin": "^2.0.6""`

I need the image to always have the same size defined by the user (when he resizes), so I have to configure the plugin like the following:
```js
const resizeablePlugin = createResizeablePlugin({
  horizontal: "absolute",
  vertical: "absolute",
});
```

The thing is, when you do this, the image always get the default size of 40x40, which in my opinion makes no sense, because the image gets too small and you can't use the default size of the image.
Investigating the code, looks like the line causing this is [this one](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-resizeable-plugin/src/createDecorator.js#L161)

For now, I made a fork which fixes this for my needs, but I'm not sure if it addresses all use cases of the plugin, as I'm not familiar with the codebase. At least all tests passed :) You can check it [here](https://github.com/exactsales/draft-js-plugins/blob/draft-js-resizeable-plugin-v2.0.6/src/createDecorator.js#L182). Unfortunately, I commited some things not related to the fix without knowing, like prettier's formatting and some fixes to be able to run the npm scripts on windows (there are some unix specific stuff on it), so to make a PR (given that you guys agree with my solution) I would have to make re-fork it.

To demonstrate the issue, here's a [codesandbox example](https://codesandbox.io/s/yjx096rw1x). You can see that by default the image gets unreadable, with 40x40, when the original is 300x125

This may be related to #1094

GabrielDuarteM:
Implements an option to set the initial height and initial width of the resized content.

Closes #1166.

BREAKING CHANGE: Removed the default initial size of 40. You can use the new options `initialHeight`
and `initialWidth` to configure it.

## Checklist

## Use-case/Problem

When using the option "horizontal" or "vertical" === 'absolute', the default value of the content to be resized will be 40px, which is too small.

Better described on #1166

## Implementation

This PR creates two new options to the resizeable plugin:

* initialWidth: sets the initial width of the resizeable content
* initialHeight: sets the initial height of the resizeable content

Also, on the issue #1166 , I raised the following question:
```
When working on the PR, one thing got me thinking, would it be a problem if we removed the default value, and made it an option, instead of always defaulting to 40? That way, the image size initially will always be the image's actual size, and then the user would resize it to be the size he want's it to be. If the default option is supplied, then the image size will initially use that, instead of the images default size.

What do you think?
```
This PR implements the behaviour above, but I can change, if that's not desireable.

Also, I couldn't find anywhere describing the options which the `createResizeablePlugin` function receives. If there is a place, I can add the new options there too.

Also², I was not able to locally run the tests, so I don't really know if that change breaks anything. Apparently this plugin doesn't have any tests, so it shouldn't (hopefully) break anything. The error I got is the following: 

```
[...]/draft-js-plugins/node_modules/jsdom/lib/jsdom/browser/Window.js:257
        throw new DOMException("localStorage is not available for opaque origins", "SecurityError");
        ^
SecurityError: localStorage is not available for opaque origins
    at get localStorage ([...]/draft-js-plugins/node_modules/jsdom/lib/jsdom/browser/Window.js:257:15)
    at [...]/draft-js-plugins/testHelper.js:36:14
```

### DIFF
--- a/draft-js-resizeable-plugin/CHANGELOG.md
+++ b/draft-js-resizeable-plugin/CHANGELOG.md
@@ -3,6 +3,9 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 3.0.0
+- BREAKING CHANGE: Removed the default initial size. Now it will by default use the default size of the image. You can use the new options `initialHeight` and `initialWidth` to set a fixed initial size to it. [#1166](https://github.com/draft-js-plugins/draft-js-plugins/issues/1166)
+
 ## 2.0.8
 - Fixed creator method.
 
--- a/draft-js-resizeable-plugin/src/createDecorator.js
+++ b/draft-js-resizeable-plugin/src/createDecorator.js
@@ -144,6 +144,8 @@ export default ({ config, store }) => (WrappedComponent) => class BlockResizeabl
       vertical,
       horizontal,
       style,
+      initialHeight,
+      initialWidth,
       // using destructuring to make sure unused props are not passed down to the block
       resizeSteps, // eslint-disable-line no-unused-vars
       ...elementProps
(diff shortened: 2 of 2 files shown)


nsklyarov:
@GabrielDuarteM i create patch with your code, and catch this warning:

```
React does not recognize the `initialWidth` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `initialwidth` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

Issue #1111, opened 2018-05-15
[draft-js-inline-toolbar-plugin] toolbar out of window viewport


karaushu:
steps to reproduce:

1. Select text near the edge of the window
2. Toolbar get out of window and user can't click some buttons.

Also get an issue with closing toolbar.
steps to reproduce:
1. Select text in editor
2. Click anywhere excluding the editor
3. Click anywhere in editor space (Toolbar will "blink")

Video bellow contain both issue (first 10 seconds issue with viewport, next with blink effect)
https://vimeo.com/269838587

PR #1209, opened 2018-10-04
keep inline toolbar inside viewport

heydiplo:
## Checklist

## Use-case/Problem

Fixed draft-js-inline-toolbar positioning to make sure it always fits inside viewport (#1111).

There's one downside. When toolbox stopped at the window's edge, arrow implemented via ::before/after remains at the center of the toolbox pointing at the wrong place. I don't see a way to fix that without adding another <div /> and splitting styles between them. And it will require a change in theming and will break current users styles. I'd like to keep this fix small and easily mergeable so I figured slightly mispositioned arrow is way better than non-usable toolbar anyway.

## Implementation

Calculate window bounds and don't let `left` go beyond them in `onSelectionChanged`.

### DIFF
--- a/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
+++ b/draft-js-inline-toolbar-plugin/src/components/Toolbar/index.js
@@ -79,13 +79,25 @@ export default class Toolbar extends React.Component {
       // but rather with a small offset so the caret doesn't overlap with the text.
       const extraTopOffset = -5;
 
+      let left = (
+        editorRoot.offsetLeft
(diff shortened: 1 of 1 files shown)


heydiplo:
I believe [tests fail](https://travis-ci.org/draft-js-plugins/draft-js-plugins/builds/437375481) due to reasons unrelated to my changes.

vdesdoigts:
Hello 👋 Did you find a solution? Thanks

heydiplo:
Hi @vdesdoigts, this is a solution, just ignored for whatever reasons.

PR #1210, opened 2018-10-06
Upgrade to babel 7

TrySound:
In this PR I'm trying to simplify build step.

`babel-plugin-webpack-loaders` is not supported anymore (I'm working
with its author) so I implemented small version of css-modules with only
private classes generation. I think later we could introduce a better
solution for styles without requiring user to inject css files.
`autoprefixer` is bultin in babel plugin.

Babel config for plugins is hoisted to the root.

The prefix in css classes is fixed in a lot of packages.


TrySound:
Tests are broken in master too. Probably because of some packages has `latest` tag.

/cc @juliankrispel

TrySound:
Friendly ping @juliankrispel

juliankrispel:
hey @TrySound thank you so much, that seem like a good change. Let me fix master, and I'll be in touch shortly

TrySound:
Hey @juliankrispel 
Would you like to continue working on this?

PR #1251, opened 2019-03-29
Add mention postfix.

stephenkey:

## Checklist

## Use-case/Problem

Add option to pass a mention postfix. This will allow wrapping the mention when used in conjunction with the prefix. 

## Implementation

Add a config option `mentionPostfix`, which will add trailing characters to the mention replaced text.


juliankrispel:
@stephenkey can you give me a more concrete example of a usecase? :)

stephenkey:
@juliankrispel Yes, I was using this as a simple way to support templating languages where you need to wrap a variable with tags (e.g, {{ firstName }} ), but I think this would be useful when wrapping the target with other markup as well. Thanks

juliankrispel:
sry but this isn't something that we're considering changing at this point

PR #1273, opened 2019-06-22
Update index.js

sirily11:
This styling fixing is about when the user input some image which over the width of the editor width.
Screen Shot 2019-06-22 at 8 48 10 AM
As you can see, when the image is oversized, the editor cannot fully display the image.
Screen Shot 2019-06-22 at 8 48 24 AM
After the fixes, the image now has the maxwitdth which equals to the size of the container.


juliankrispel:
@sirily11 you can add a css class to the config of the image plugin and avoid hard coding a style

PR #1280, opened 2019-07-11
Use modular lodash

TrySound:
In this diff I replaced per method packages which with modular lodash. Per method packages will not be supported in v5.

Details: https://github.com/lodash/lodash/wiki/Roadmap#v500-2019

This PR will reduce amount of dependencies as previously two ways were
used in mention plugin (lodash/escapeRegExp and lodash.escaperegexp).

Closes https://github.com/draft-js-plugins/draft-js-plugins/pull/1223


TrySound:
Looks like the last error is not related.

PR #1282, opened 2019-07-11
Remove decorate-component-with-props

TrySound:
With function components such HOC is not relevant. One less dependency.

/cc @juliankrispel


TrySound:
I will rebase

TrySound:
Ok, you are faster :)

TrySound:
Oops, will fix this

PR #1284, opened 2019-07-11
Linaria proof-of-concept

TrySound:
- added babel plugin to remove linaria imports (not touched by linaria plugin)
- match all css files inside of lib-css in concatCssFiles.js
- moved defaultTheme into own module with all styles


draft-js-mention-plugin/lib/plugin.css
```diff
@@ -0,0 +1,6 @@
+.m11k5wy5,.m11k5wy5:visited{color:#575f67;cursor:pointer;display:inline-block;background:#e6f3ff;padding-left:2px;padding-right:2px;border-radius:2px;-webkit-text-decoration:none;text-decoration:none;}.m11k5wy5:hover,.m11k5wy5:focus{color:#677584;background:#edf5fd;outline:0;}.m11k5wy5:active{color:#222;background:#455261;}
```
TrySound: linaria generates only hashes without prefixes like package name or inferred from const class names. There is a PR for prefix but they looks like stuck on migration to TS for now.
https://github.com/callstack/linaria/pull/369

PR #1293, opened 2019-07-24
Replace .npmignore with files filter

TrySound:
`files` fields is simpler and covers more cases.


TrySound:
Ping @juliankrispel

TrySound:
Friendly ping @juliankrispel

PR #1306, opened 2019-08-21
Format source code and tests with prettier

TrySound:
This code should be formatted at once to reduce diff of upcoming PRs and
simplify review. Docs and stories are less relevant and can be formatted
later.


juliankrispel:
maybe skip eslint rules that touch prettier, or use a prettier friendly config, https://prettier.io/docs/en/integrating-with-linters.html

PR #1307, opened 2019-08-21
Remove deprecated draft-js hooks usage

TrySound:
All onDownArrow, onUpArrow, onLeftArrow, onRightArrow, onEscape, onTab
usages are replaced with keyBindingFn.

Their mentions are dropped from HOW_TO_CREATE_A_PLUGIN tut.

### DIFF
--- a/HOW_TO_CREATE_A_PLUGIN.md
+++ b/HOW_TO_CREATE_A_PLUGIN.md
@@ -58,12 +58,6 @@ A plugin accepts all standard props the Draft.js Editor Component uses. [See det
 - [handlePastedFiles](https://facebook.github.io/draft-js/docs/api-reference-editor.html#handlepastedfiles)
 - [handleDroppedFiles](https://facebook.github.io/draft-js/docs/api-reference-editor.html#handledroppedfiles)
 - [handleDrop](https://facebook.github.io/draft-js/docs/api-reference-editor.html#handledrop)
-- [onEscape](https://facebook.github.io/draft-js/docs/api-reference-editor.html#onescape)
-- [onTab](https://facebook.github.io/draft-js/docs/api-reference-editor.html#ontab)
-- [onUpArrow](https://facebook.github.io/draft-js/docs/api-reference-editor.html#onuparrow)
-- [onRightArrow](https://draftjs.org/docs/api-reference-editor.html#onrightarrow)
-- [onDownArrow](https://facebook.github.io/draft-js/docs/api-reference-editor.html#ondownarrow)
-- [onLeftArrow](https://draftjs.org/docs/api-reference-editor.html#onleftarrow)
 - [onFocus](https://draftjs.org/docs/api-reference-editor.html#onfocus)
 - [onBlur](https://draftjs.org/docs/api-reference-editor.html#onblur)
 
--- a/draft-js-emoji-plugin/CHANGELOG.md
+++ b/draft-js-emoji-plugin/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## to be released
+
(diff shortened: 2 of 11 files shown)


HOW_TO_CREATE_A_PLUGIN.md
```diff
@@ -58,12 +58,6 @@ A plugin accepts all standard props the Draft.js Editor Component uses. [See det
 - [handlePastedFiles](https://facebook.github.io/draft-js/docs/api-reference-editor.html#handlepastedfiles)
 - [handleDroppedFiles](https://facebook.github.io/draft-js/docs/api-reference-editor.html#handledroppedfiles)
 - [handleDrop](https://facebook.github.io/draft-js/docs/api-reference-editor.html#handledrop)
-- [onEscape](https://facebook.github.io/draft-js/docs/api-reference-editor.html#onescape)
-- [onTab](https://facebook.github.io/draft-js/docs/api-reference-editor.html#ontab)
-- [onUpArrow](https://facebook.github.io/draft-js/docs/api-reference-editor.html#onuparrow)
-- [onRightArrow](https://draftjs.org/docs/api-reference-editor.html#onrightarrow)
-- [onDownArrow](https://facebook.github.io/draft-js/docs/api-reference-editor.html#ondownarrow)
-- [onLeftArrow](https://draftjs.org/docs/api-reference-editor.html#onleftarrow)
```
juliankrispel: :+1:

draft-js-emoji-plugin/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## to be released
+
+- removed deprecated draft-js hooks (onUpArrow, onDownArrow, onEscape, onTab) usage
+
```
juliankrispel: yaasss :+1:

pederjohnsen:
@TrySound think this is broken? Using latest version, arrow keys, tab and esc no longer do anything in suggestions lists :thinking:

juliankrispel:
@pederjohnsen can you open an issue and give us like a codesandbox example?

pederjohnsen:
@juliankrispel I'll double check it's not just caused by something in our forks.. :ok_hand:

TrySound:
@pederjohnsen Make sure you don't have keyBIndingFn with getDefaultKeyBinding(event)

pederjohnsen:
@TrySound ah, thanks, that seems to be the reason :ok_hand: 

Does this mean such function can no longer be used? :thinking: 
```
  keyBindingFunction(event) {
    const {submitOnEnterKey, singleLine} = this.props;
    const {hasCommandModifier} = KeyBindingUtil;

    if (submitOnEnterKey && (event.keyCode === 13 || event.keyCode === 10) && (singleLine || !event.shiftKey)) {
      return 'submit';
    }

    if (hasCommandModifier(event) && (event.keyCode === 13 || event.keyCode === 10)) {
      return 'submit';
    }

    return getDefaultKeyBinding(event);
  }
```

TrySound:
The problem is only `return getDefaultKeyBinding(event)` in the end. It should be removed.

pederjohnsen:
@TrySound great :ok_hand: thanks for the clarification :tada:

coldshower:
@TrySound
Hey Bogdan, thanks for the guidance. What do you think about updating the documentation, which recommends adding a `return getDefaultKeyBinding(event)`? 

https://draftjs.org/docs/advanced-topics-key-bindings#example

I can make the PR unless I'm missing something.

Stalinko:
Gosh! I've spent whole day today trying to understand why key bindings for mentions popup broke at some moment. Even re-read the docs and source code, but the issue was in just 1 line! Guys, that should really be addressed in the docs! @coldshower is absolutely right

PR #1333, opened 2019-09-05
percentage of width and height disappear

popmanhe:
In editor, width and height are working fine. The percentage of width and height are kept, even though the width and height returned at line 152 are just number.  for instance, the width is returned as 75, but in editor, it is 75%. I have not figured out the reason.
however, if use draftToHtml component to convert draft to html,  the percentage of width and height are lost, which is correct because entitydata does not have percetage. so the changes just make sure the percentage is also set  with numbers.

## Checklist

## Use-case/Problem

## Implementation

## Demo


sibelius:
can you provide screenshots of before and after behavior?

PR #1336, opened 2019-09-05
[mention-plugin] make open state controlled

TrySound:
See changelog and stories/mentions/src/App.js

- dropped UNSAFE_ hooks
- made suggestions open state controlled

### DIFF
--- a/draft-js-mention-plugin/CHANGELOG.md
+++ b/draft-js-mention-plugin/CHANGELOG.md
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add esm support
 - Use lodash-es in esm bundle
 - Added open and onOpenChange required props to make state controlled outside and prevent reacting on suggestions list
+- onOpen and onClose callbacks are removed in favour of onOpenChange
 - Made suggestions prop required (pass empty array for async suggestions)
 - Remove prevState and state from positionSuggestions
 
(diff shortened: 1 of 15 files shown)


draft-js-mention-plugin/CHANGELOG.md
```diff
@@ -5,11 +5,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
-- Replace legacy lifecycle hooks with UNSAFE aliases; the required react version is 16.3
+- Remove legacy lifecycle hooks
+- Require react 16.3 and above version
 - Migrate styles to linaria
 - Hide internals in single bundle
 - Add esm support
 - Use lodash-es in esm bundle
+- Added open and onOpenChange required props to make state controlled outside and prevent reacting on suggestions list
+- Made suggestions prop required (pass empty array for async suggestions)
+- Remove prevState and state from positionSuggestions
```
juliankrispel: this needs to be added to the docs

draft-js-mention-plugin/src/MentionSuggestions/index.js
```diff
@@ -10,35 +10,25 @@ import defaultEntryComponent from './Entry/defaultEntryComponent';
...
-    if (nextProps.suggestions.length === 0 && this.state.isActive) {
-      this.closeDropdown();
-    } else if (
-      nextProps.suggestions.length > 0 &&
-      nextProps.suggestions !== this.props.suggestions &&
-      !this.state.isActive
-    ) {
-      this.openDropdown();
-    }
-  }
```
juliankrispel: what do we do now to replace this behaviour?

TrySound: Allow user to handle it. Most of the time it's not desired behaviour. It forces to have suggestions in local state instead of being inferred which leads to complications when we need to update this state from props.

PR #1339, opened 2019-09-16
Handle video and add parameter to prevent user resize

RexProg:
Update resizable plugin by adding a parameter "isResizable " to allow the size of the media to be controlled and unchangeable
Update video plugin to handle video by video element instead of invalid div


sibelius:
can you rebase?

RexProg:
> can you rebase?

Done :)

sibelius (approved):

PR #1345, opened 2019-09-28
Fix decorator resolve

### LINKED ISSUE #1343: [Mention plugin] - Resolving decorators on componentWillReceiveProps
Since https://github.com/draft-js-plugins/draft-js-plugins/pull/960,
we resolve plugin decorators on mount (componentDidMount).

But our component receives often a new editorState, and loses its decorators.
Is it an option to also resolve it in componentWillReceiveProps? 

I would think you always want to resolve decorators again? Specially for plugins?

thaoms:
Even when the current and next decorators from props are empty.

Can someone please review, I'm not used working in this codebase.

closes #1343

### DIFF
--- a/draft-js-plugins-editor/CHANGELOG.md
+++ b/draft-js-plugins-editor/CHANGELOG.md
@@ -5,7 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## To be released
 
-- Replace legacy lifecycle hooks with UNSAFE aliases; the required react version is 16.3
+- Remove deprecated lifecycle hooks
 - Hide internals in single bundle
 - Add esm support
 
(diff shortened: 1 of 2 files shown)


TrySound:
Hi, thanks for this. Do you know can we safely get rid from UNSAFE_ lifecycles here?

thaoms:
> Hi, thanks for this. Do you know can we safely get rid from UNSAFE_ lifecycles here?

I should really look into that, not sure.

TrySound:
Awesome! Could you also replace this note in changelog
> Replace legacy lifecycle hooks with UNSAFE aliases; the required react version is 16.3

with this one?
> Remove deprecated lifecycle hooks

thaoms:
Could you or someone else give this a good testing? Thanks!

juliankrispel:
I will give this a test today or tmw :+1:

juliankrispel:
Ok so in all honesty I have very little time currently, need a test case for this because I currently don't have one so I can't say 100% whether this actually works or not. Can someone add a story that switches decorators just so we can test this?

sibelius:
we wanna move to hooks instead of classes

can you tackle this?

PR #1346, opened 2019-09-30
Added typescript definitions for SideToolbar plugin

CFerndp:

## Checklist

## Use-case/Problem
Not able to use this plugin within a TypeScript project

## Implementation
Added `index.d.ts` file with a definition of each part of the plugin

### DIFF
--- a/draft-js-side-toolbar-plugin/CHANGELOG.md
+++ b/draft-js-side-toolbar-plugin/CHANGELOG.md
@@ -8,8 +8,6 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Migrate styles to linaria
 - Hide internals in single bundle
 - Add esm support
-
-## 3.0.3
 - Added TS definitions
 
 ## 3.0.2
--- a/draft-js-side-toolbar-plugin/CHANGELOG.md
+++ b/draft-js-side-toolbar-plugin/CHANGELOG.md
@@ -9,6 +9,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Hide internals in single bundle
 - Add esm support
 
+## 3.0.3
+- Added TS definitions
+
 ## 3.0.2
 
 - Allow draft-js v0.11
--- a/draft-js-side-toolbar-plugin/package.json
+++ b/draft-js-side-toolbar-plugin/package.json
@@ -12,6 +12,7 @@
   },
   "main": "lib/index.cjs.js",
   "module": "lib/index.esm.js",
+  "types": "lib/index.d.ts",
   "files": [
     "lib"
   ],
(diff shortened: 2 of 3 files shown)


draft-js-side-toolbar-plugin/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Hide internals in single bundle
 - Add esm support
 
+## 3.0.3
+- Added TS definitions
```
TrySound: Put it in to be released section please

draft-js-side-toolbar-plugin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "draft-js-side-toolbar-plugin",
-  "version": "3.0.2",
+  "version": "3.0.3",
```
TrySound: Do not increase version in this PR. We will do that separately.

CFerndp: Sure, sorry about it

draft-js-side-toolbar-plugin/src/index.d.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export interface SideToolbarChildrenProps {
+    theme: SideToolbarPluginTheme["buttonStyles"];
+    getEditorState: () => EditorState;
+    setEditorState: (editorState: EditorState) => void;
+    onOverrideContent: (content: ComponentType<SideToolbarChildrenProps>) => void;
+}
+
+export interface SideToolbarProps {
+    children?: ComponentType<SideToolbarChildrenProps>;
```
TrySound: This should be a function, not a component.

CFerndp: it's the same, a function who receive props and return a ReactNode is the same that if I defined it as a ComponentType who receive props typed by SideToolbarChildrenProps

TrySound: No, it's not.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L71

This function is just called. Not instantiated with React.createElement. If you use ComponentType user can pass class components. Classes cannot be called without `new` keyword.
https://github.com/draft-js-plugins/draft-js-plugins/blob/040436205cdf59236267c4dca40c80d41b8b8f9c/draft-js-side-toolbar-plugin/src/components/BlockTypeSelect/index.js#L62

CFerndp: Totally right, fixed. Thank you so much for the information and the effort :)

PR #1347, opened 2019-09-30
Fixed ts definition in order to be able to include the component with…

CFerndp:
…out childs

## Checklist

## Use-case/Problem

When you try to reproduce the example fo this plugin in a TS project, the definition does not allow you to include the component without chid

## Implementation

Fixed `index.d.ts` definition in order to be able to include the component without childs

### DIFF
--- a/draft-js-inline-toolbar-plugin/src/index.d.ts
+++ b/draft-js-inline-toolbar-plugin/src/index.d.ts
@@ -28,7 +28,7 @@ export interface ToolbarChildrenProps {
 }
 
 export interface ToolbarProps {
-  children(externalProps: ToolbarChildrenProps): ReactNode;
+  children?: ComponentType<ToolbarChildrenProps>;
 }
 
 export type InlineToolBarPlugin = EditorPlugin & {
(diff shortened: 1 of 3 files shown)


draft-js-inline-toolbar-plugin/src/index.d.ts
```diff
@@ -28,7 +28,7 @@ export interface ToolbarChildrenProps {
 }
 
 export interface ToolbarProps {
-  children(externalProps: ToolbarChildrenProps): ReactNode;
+  children?: ComponentType<ToolbarChildrenProps>;
```
TrySound: I'm sure it's not component meant here. It's render prop. It should be just a function.

CFerndp: It's the same but whit this syntax I'm able to include it as optional.

TrySound: It's not the same. ComponentType can be class too.
```js
children: (externalProps: ToolbarChildrenProps) => ReactNode
```

CFerndp: Oh, thanks them, fixed and thanks for the info :)

PR #1357, opened 2019-10-25
Fix Types.

ghostrick:
The current code does not work properly when with using TypeScript.


adamdharrington:
Looks good to me for what it's worth.

crusoexia:
Met the same issue. When will this pr get merged? It's about half of a year  since the pr created.

sibelius:
can you rebase?

PR #1374, opened 2020-01-14
Trigger rule

hyy1115:
When the input is 'trigger: @' anywhere, open should be triggered.
Because the user doesn't always know that a leading space is needed.

### DIFF
--- a/draft-js-mention-plugin/src/MentionSuggestions/index.js
+++ b/draft-js-mention-plugin/src/MentionSuggestions/index.js
@@ -122,7 +122,7 @@ export class MentionSuggestions extends Component {
               'g'
             ).test(plainText) &&
             anchorOffset <= end) || // @ is the first character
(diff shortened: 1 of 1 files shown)


HenryCharlesAnderson:
+1 to this PR. Opening the suggestions immediately after typing the trigger is a pretty common usecase. That's how it works right here on github, for example. 

Changing this could break other people's expectations though. Twitter, for example, only opens the popover on the first character after the trigger. Perhaps it could be a configuration option?

sibelius:
can you rebase?

sibelius:
on master this is opening for me

image

maciejlaufer:
> on master this is opening for me
> 
> image

It's opening only without support for whitespace.

sibelius:
I can merge if the fixes are rebase or get into another PR

sibelius:
fixed here https://github.com/draft-js-plugins/draft-js-plugins/pull/1507

PR #1396, opened 2020-04-03
Update index.d.ts

edparsons:
Adding undefined as an acceptable return from a keyBindingFn to signify it has not been handled

## Checklist

## Use-case/Problem

It appears in line 321 of Editor/index.js that undefined is an acceptable return from a plugin[methodName]. As the keyBindingFn is automatically added to the plugins list, at the moment there is a ts lint error when passing in a function that can return undefined

## Implementation

Overriding the type def from draft-js for keyBindingFn

### DIFF
--- a/draft-js-plugins-editor/src/index.d.ts
+++ b/draft-js-plugins-editor/src/index.d.ts
@@ -47,7 +47,7 @@ export interface EditorPlugin {
   keyBindingFn?(
     e: KeyboardEvent,
     pluginFunctions: PluginFunctions
-  ): DraftEditorCommand | null;
+  ): DraftEditorCommand | null | undefined;
   handleReturn?(
     e: KeyboardEvent,
     editorState: EditorState,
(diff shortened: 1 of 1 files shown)


draft-js-plugins-editor/src/index.d.ts
```diff
@@ -105,6 +105,8 @@ export interface PluginEditorProps extends EditorProps {
   defaultKeyBindings?: boolean;
   defaultKeyCommands?: boolean;
   defaultBlockRenderMap?: boolean;
+  
+  keyBindingFn?(e: SyntheticKeyboardEvent): EditorCommand | null | undefined;
```
adamdharrington, line 109: It should be altered up on line 50 IMO, aside from that +1

 https://github.com/draft-js-plugins/draft-js-plugins/pull/1396/commits/14ebc8b4a777bfe53433acf2bf74cc9675283c20#diff-b2ce8263c95dfe11ec41b895dc569ac6R50
```suggestion
  keyBindingFn?(
    e: KeyboardEvent,	
    pluginFunctions: PluginFunctions
  ): DraftEditorCommand | null | void;
```

edparsons: Hi @adamdharrington, I think you are right it should be added to line 50, yet it also needs to remain here as PluginEditorProps, does not inherit from EditorPlugin

PR #1403, opened 2020-04-23
Fixed emoji select popover close issue on multiple instance case

bagratinho:

## Checklist

## Use-case/Problem

When using multiple emoji picker plugins on one page (imagine a list of posts like fb timeline that each have its own comment input and emoji picker button) there is an issue that already open emoji picker popover want close if you click on another instances popover trigger button. 
The issue can be seen on demo page of emoji plugin https://www.draft-js-plugins.com/plugin/emoji

## Implementation

As can be seen, in the change I have updated the event listener on the document to call a specific function `closeIfClickedOutside` which is simply checking if the click target element is something inside emoji popover component or no. And if no then it calls a function to close the popover and doing nothing otherwise.

```
  closeIfClickedOutside = e => {
    if (this.emojiSelectRef && !this.emojiSelectRef.contains(e.target)) {
      this.closePopover();
    }
  }
```

## Demo

After fix the one popover instance will always close when opening another popover.

### DIFF
--- a/draft-js-emoji-plugin/src/components/EmojiSelect/index.js
+++ b/draft-js-emoji-plugin/src/components/EmojiSelect/index.js
@@ -37,6 +37,8 @@ export default class EmojiSelect extends Component {
     toneSelectOpenDelay: 500,
   };
 
+  emojiSelectRef = React.createRef();
+
   // Start the selector closed
   state = {
     isOpen: false,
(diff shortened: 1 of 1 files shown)


draft-js-emoji-plugin/src/components/EmojiSelect/index.js
```diff
@@ -45,17 +47,18 @@ export default class EmojiSelect extends Component {
...
   }
 
   componentWillUnmount() {
-    document.removeEventListener('click', this.closePopover);
+    document.removeEventListener('click', this.closeIfClickedOutside);
   }
 
-  onClick = e => {
-    e.stopPropagation();
-    e.nativeEvent.stopImmediatePropagation();
```
juliankrispel, line 57: I'd prefer it if those are kept, to make sure the behaviour doesn't change

bagratinho: thanks for reviewing it @juliankrispel, the unique purpose of the portion of the code you have mentioned is to prevent EmojiSelect from closing when clicking on something thats inside it, but with this modifications its no longer needed cause now it will be handled by this portion.

```
  componentDidMount() {
    document.addEventListener('click', this.closeIfClickedOutside);
  }

&

  closeIfClickedOutside = e => {
    if (this.emojiSelectRef && !this.emojiSelectRef.current.contains(e.target)) {
      this.closePopover();
    }
  }
```

What it actually does it checks if the random click target is outside of the EmojySelect component and closes it. 

If I leave that it will prevent other open instances of EmojySelect from being closed when you open another one.
So imagine I open an instance of EmojySelect, it opens up, then I open another one without closing the first one, the stop propogation will prevent the click bubbling so it want reach the document and already opened instance of EmojySelect will simply not close.

But still as I have already highlighted that portion is simply not needed because of use of `closeIfClickedOutside` function.

let me know what you think ;)

bagratinho:
@juliankrispel please review my comment on reviewed changes

PR #1439, opened 2020-07-07
inline-toolbar: Try getting the relative parent window object first, …

nadelalon:
The toolbar isn't positioned according to the selected text because it should be using the iframe window object and not the main window.


sibelius:
can you provide screenshot before and after this change?

nadelalon:
As mentioned, this applies only to draft editors rendered inside an iframe. The green dashed border marks the boundaries of the draft editor. These are two examples for before and after this change.

Before:

Screenshot from 2020-09-09 13-40-05
Screenshot from 2020-09-09 13-40-19

After:

Screenshot from 2020-09-09 13-44-55
Screenshot from 2020-09-09 13-45-13

Issue #1393, opened 2020-03-16
Editor mount causing react warning "Warning: Cannot update a component from inside the function body of a different component."


sspread:
Likely due to more recent version of react throwing the warning. As discussed here, (https://github.com/facebook/react/issues/18178) the warning is new and will eventually be a hard exception. I believe the problem comes from this line https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-plugins-editor/src/Editor/index.js#L74 when `onChange` callback results in a `setState` call, resulting in attempt to render the Editor component within its own onMount hook. I suspect this line didn't work as intended and now the warning brings it to light.

abheist:
Same issue;

lciii:
I've updated to React 16.13.1 as suggested in [this thread](https://github.com/facebook/react/issues/18178) and I'm still getting the same warning.

sspread:
@lciii Yes its not a react issue. It just surfaced more recently since react started warning about the issue that already existed. I've outlined the specific problem in my original post.

abheist:
Instead of passing `setState` and updating the component's state from the parent, I'm updating the state from the same component, and it solved the problem for me.

But you might be calling `setState` on every react rendering, try to put `console.log` below `setState` and check how many times it's updating on any change. 

You can put the `setState` under any condition or under `useEffect`, this might solve your issue.
It's not from the Draft but is a restriction from `react`.

This might help:
https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render

sspread:
@abheist Putting `setState` in `useEffect` will solve it, the same way using `setState` in a `setTimeout` will solve it. But the issue is still in this plugin, as following the simple implementation example on this github home page will cause the warning.

abheist:
@sspread Yeah, totally, I was facing many issues while using this plugin library, so kinda move to the pure draft.js. Still, this need a major upgrade I think.

gnoesiboe:
The problem I think is that `componentWillMount` is considered part of the rendering process. However if [the same line](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-plugins-editor/src/Editor/index.js#L74) is to be moved to the `componentDidMount`, it appears not to be. I've been playing around with this locally and no longer got a warning on mount. Is this something that might be considered a solution? If it is, let me know and I will file a PR.

sspread:
[...]

PR #1445, opened 2020-07-15
#1393 Apply decorators and set the selection after rendering

gnoesiboe:
Apply decorators and set selection after initial rendering to prevent `onChange` to trigger during the render cycle, as this now results in ["Warning: Cannot update a component from inside the function body of a different component."](https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render). 

**The problem**

Many people will have the `onChange` callback hooked up to some state setter on the editor, i.e.

```javascript
const [editorState, setEditorState] = useState(..);

return <Editor onChange={(newEditorState) => setEditorState(newEditorState)} editorState={editorState} />
```

As `onChange` is called during rendering [in the Editor](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-plugins-editor/src/Editor/index.js#L74), when it is hooked up to a state setter, this will result in the console warning:

> ["Warning: Cannot update a component from inside the function body of a different component."](https://reactjs.org/blog/2020/02/26/react-v16.13.0.html#warnings-for-some-updates-during-render)

To prevent this, I moved these lines to `componentDidMount`. 

Todo:

Help is wanted, as at this point I don't even get `pre-commit` hooks and testing working locally..

### DIFF
--- a/draft-js-plugins-editor/CHANGELOG.md
+++ b/draft-js-plugins-editor/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Replace legacy lifecycle hooks with UNSAFE aliases; the required react version is 16.3
 - Hide internals in single bundle
 - Add esm support
+- Apply decorators and set selection after initial rendering to prevent `onChange` to trigger during the render cycle
 
 ## 3.0.0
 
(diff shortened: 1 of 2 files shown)


gnoesiboe:
Anyone?!

sibelius:
We want to move this component to use hooks instead of class components

Do you wanna tackle this?

gnoesiboe:
@sibelius I don't see what that has to do with this pull request. Why make it bigger than it has to be? Shouldn't that be the next step, after this is merged?

gnoesiboe:
Thanks!

gnoesiboe:
> We want to move this component to use hooks instead of class components
> Do you wanna tackle this?

I'll see if I can find some time for it, but I'm quite busy at the moment. No garantees..

PR #1473, opened 2020-09-02
feat(mention): hooks on mention suggestions portal

sibelius:


draft-js-mention-plugin/src/index.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { Map } from 'immutable';
 import Mention from './Mention/Mention';
 import MentionSuggestions from './MentionSuggestions'; // eslint-disable-line import/no-named-as-default
-import MentionSuggestionsPortal from './MentionSuggestionsPortal';
+import MentionSuggestionsPortal from './MentionSuggestionsPortal/MentionSuggestionsPortal';
```
TrySound, line 5: I'm not sure why this nesting is need? I would flatten files.

sibelius: it makes sense

it was a lot of index.js, but I prefer to have the name in the file to make discovery easier on IDE

I or you can move files around

PR #1476, opened 2020-09-03
migrating counter-plugin

gabrielreisn:

## Checklist

## Use-case/Problem

## Implementation

## Demo


sibelius (approved):
It is looking good

Feel free to merge if all tests are pssing

sibelius:
image

some tests are failing

gabrielreisn:
Yeah, i'm aware of that, it was caused by some confusion on mocha/chai trying to run the new tests that now uses jest, so probably I'll need to remove the pattern .spec. and make it as a standard only for jest (I'm open to suggestions)

sibelius:
We can have .spec for mocha and .test for jest or the other way around until we finish the migration

gabrielreisn:
Also, i'd like to use this one to start the typescript migration, you'd mentioned the babel-typescript-plugin. Any other tip for the walkthrough? should I place it on the project root or configure it on each plugin? should I export some specific type for each plugin?

sibelius:
for typescript

we can convert one plugin at a time as well

we can have typescript in the root

and use project references to make sure one plugin can compose from another

https://www.typescriptlang.org/docs/handbook/project-references.html

we can also let typescript check errors on .js files until we finish the migration

we can start being less strict and improve later on

keep pull requests shorter so more people can help with other plugins at the same time

sibelius:
do you need any help with this?

gabrielreisn:
Since i'd take a bit more time on this than expected (couldn't do much on the last week) I'll leave the TS migration on this plugin for later. Can you take a look again before merge @sibelius ?

sibelius (approved):
We can take a more incremental approach

Do one thing per PR

Tks

PR #1490, opened 2020-09-09
Set up stale bot

julioxavierr:
## Use-case

Mark as stale and close inactive issues.

## Implementation

Integration of [stale bot](https://github.com/probot/stale)


sibelius (approved):
tks

PR #1594, opened 2020-11-20
Improve issue templates

hcomPullReq:
## Checklist

## Use-case/Problem

Currently there is only one template for questions, feature requests and bugs. Also it would be better if questions are asked in the slack channel as it may be hard to find the link.

## Implementation

This PR add new templates which should also add tags. A link to Slack should be provided.


sibelius (approved):
nice

PR #1630, opened 2020-12-08
fix(anchor): url validation was wrong & naming of the state variable was inconsistent

Nasicus:
## Checklist

### DIFF
--- a/packages/anchor/src/components/AddLinkForm.tsx
+++ b/packages/anchor/src/components/AddLinkForm.tsx
@@ -27,7 +27,7 @@ interface AddLinkFormParams extends AddLinkFormPubParams {
 
 const AddLinkForm = (props: AddLinkFormParams): ReactElement => {
   const [value, setValue] = useState('');
-  const [isInvalid, setIsvalid] = useState(false);
+  const [isValid, setIsValid] = useState(false);
 
   const input = useRef<HTMLInputElement>(null);
 
@@ -45,13 +45,7 @@ const AddLinkForm = (props: AddLinkFormParams): ReactElement => {
 
   const onChange = (event: ChangeEvent<HTMLInputElement>): void => {
     const newValue = event.target.value;
-
-    if (isInvalid && isUrl(URLUtils.normalizeUrl(newValue))) {
-      setIsvalid(false);
-    } else {
-      setIsvalid(true);
-    }
-
+     setIsValid(isUrl(URLUtils.normalizeUrl(newValue)));
     setValue(newValue);
   };
 
@@ -65,7 +59,7 @@ const AddLinkForm = (props: AddLinkFormParams): ReactElement => {
     if (!URLUtils.isMail(URLUtils.normaliseMail(url))) {
       url = URLUtils.normalizeUrl(url);
       if (!isUrl(url)) {
-        setIsvalid(true);
+        setIsValid(false);
         return;
       }
     } else {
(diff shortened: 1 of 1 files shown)


packages/anchor/src/components/AddLinkForm.tsx
```diff
@@ -45,13 +45,7 @@ const AddLinkForm = (props: AddLinkFormParams): ReactElement => {
 
   const onChange = (event: ChangeEvent<HTMLInputElement>): void => {
     const newValue = event.target.value;
-
-    if (isInvalid && isUrl(URLUtils.normalizeUrl(newValue))) {
```
Nasicus, line 49: I don't understand why this condition was like this: in the case if the current state `isInvalid` then we for sure want to recheck if the newly url is now valid, in the case when the url was `valid` we also need to change (on each change) if the new value is valid.... or did I miss something?

fxOne (changes requested):
It would be awesome if you could find some time to add jest test with react-testing-lib for your changes!

packages/anchor/src/components/AddLinkForm.tsx
Nasicus: looks like it was already wrong before though

```
    if (this.state.isInvalid && URLUtils.isUrl(URLUtils.normalizeUrl(value))) {	
      nextState.isInvalid = false;	
    }	
    this.setState(nextState);
```

let's assume the value was `https://google.ch` (valid) so `this.state.isInvalid` was `false` now the next change is `https:/google.ch` (removed a slash). It will now still be seen as valid, since the check is never executed.

But anyway I think it's correct now :)

Nasicus:
> It would be awesome if you could find some time to add jest test with react-testing-lib for your changes!

are there any tests yet? if there are tests which fail I'd be fine to fix them.

Also I cannot check out this whole repo on windows since there appears to be a slash in a file name:

```
POST git-upload-pack (gzip 2675 to 1349 bytes)
remote: Enumerating objects: 164, done.
remote: Counting objects: 100% (164/164), done.
remote: Compressing objects: 100% (146/146), done.
remote: Total 25145 (delta 62), reused 37 (delta 12), pack-reused 24981
Receiving objects: 100% (25145/25145), 11.74 MiB | 7.55 MiB/s, done.
Resolving deltas: 100% (15825/15825), done.
error: Invalid path 'packages/mention/src/Mention/__test__/Mention.test.tsx\'
```

fxOne:
> > It would be awesome if you could find some time to add jest test with react-testing-lib for your changes!
> 
> are there any tests yet? if there are tests which fail I'd be fine to fix them.

All test are working, no panic! The issue is that there are currently no test for this file, but I'll merge this PR as soon as Travis CI is done.
 
> Also I cannot check out this whole repo on windows since there appears to be a slash in a file name:
> 
> ```
> POST git-upload-pack (gzip 2675 to 1349 bytes)
> remote: Enumerating objects: 164, done.
> remote: Counting objects: 100% (164/164), done.
> remote: Compressing objects: 100% (146/146), done.
> remote: Total 25145 (delta 62), reused 37 (delta 12), pack-reused 24981
> Receiving objects: 100% (25145/25145), 11.74 MiB | 7.55 MiB/s, done.
> Resolving deltas: 100% (15825/15825), done.
> error: Invalid path 'packages/mention/src/Mention/__test__/Mention.test.tsx\'
> ```
> 
> I actually did all my changes directly here on the site so far ;)

You should try out [WSL2](https://docs.microsoft.com/en-us/windows/wsl/install-win10) it's works really amazing! I have no Idea where the error is from :(

fxOne (approved):

Issue #1365, opened 2019-11-26
Hashtag plugin not mapping to entity map


t0sic:
Don't know if this is a bug or not but hashtag does not map to entity map meanwhile the mention plugin does. This is a issue for me as my parser doesn't understand that there is a entity type.

gauravsoti1:
Are you sure that the entity doesn't get mapped? Or is that just the styles are not visible, hence you are not able to see the difference?

vinceprofeta:
@gauravsoti1 I am able to reproduce this.

webdeb:
also facing this, and like OP I would like to be able to have it as an entity so I don't need to parse it separately, see the output from `convertToRaw` 

draft-js-plugins/editor@4.0.0-beta3
draft-js-plugins/hashtag@4.0.0-beta2

```
{
  "blocks": [
    {
      "key": "e546j",
      "text": "No #tag entity",
      "type": "unstyled",
      "depth": 0,
      "inlineStyleRanges": [],
      "entityRanges": [],
      "data": {}
    }
  ],
  "entityMap": {}
}
```

hanihusam:
> also facing this, and like OP I would like to be able to have it as an entity so I don't need to parse it separately, see the output from `convertToRaw`
> 
> draft-js-plugins/editor@4.0.0-beta3 draft-js-plugins/hashtag@4.0.0-beta2
> 
> ```
> {
>   "blocks": [
>     {
>       "key": "e546j",
>       "text": "No #tag entity",
>       "type": "unstyled",
>       "depth": 0,
>       "inlineStyleRanges": [],
>       "entityRanges": [],
>       "data": {}
>     }
>   ],
>   "entityMap": {}
> }
> ```

Do you already have the solution for this? I have the same issue. @webdeb

PR #1633, opened 2020-12-09
Export links and hashtag functions

hcomPullReq:
## Checklist

## Use-case/Problem

Sometimes it is useful to get the links and hashtag which are handled by the plugins, see also: https://github.com/draft-js-plugins/draft-js-plugins/issues/1365. Currently there are no entities exported.

## Implementation

As far as I saw the only possible way to generate the entities while typing is in the draft-js `onChange` method and parse always the full text. Another issue would be that the entities would be to detect removed entities. 

This PR exports the `extractHashtagsWithIndices` for `@draft-js-plugins/hashtag` and `extractLinks` for `@draft-js-plugins/linkify` to implement own solutions base on the same algorithm that the plugins are using for highlighting.


irvingv8:
@sibelius this is a really nice feature, and I have been implementing workarounds for this for a while, is there anything I can help with to get this merged ?

badazzdev:
Hey there,

are there any implementation demos to use `extractHashtagsWithIndices` and` extractLinks` to create entities from it? wrap my head around since days and can't find a working solution.

Thanks =)

iraybi:
@CrazyDevCake Have you found any solution?

PR #1701, opened 2021-01-13
chore: expose createBlockAlignmentButton in buttons package

lwhiteley:
## Use-case/Problem

in v3, I use to be able to access the `createBlockAlignmentButton` util even though it wasn't publicly exposed, but now it is written in typescript thus this is now necessary

```js
// before
import createBlockAlignmentButton from 'draft-js-buttons/lib/utils/createBlockAlignmentButton';
```

## Implementation

exposing util to make it public

```js
// now
import { createBlockAlignmentButton } from '@draft-js-plugins/buttons'
```

### DIFF
--- a/packages/buttons/CHANGELOG.md
+++ b/packages/buttons/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## To be released
 
+- Expose createBlockAlignmentButton util
+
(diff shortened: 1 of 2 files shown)


fxOne:
Thanks for your contribution!

Can you please add your change to the changelog?

lwhiteley:
> Can you please add your change to the changelog?

updated

fxOne (approved):

PR #1710, opened 2021-01-16
Revert "chore(deps): bump immutable from 3.7.6 to 3.8.2"

TrySound:
Reverts draft-js-plugins/draft-js-plugins#1531

Draft-js still depends on "~3.7.4" which makes production bundle quite bloated with immutable.js instances. Only yarn resolution would help.


sibelius (approved):
What about make it a peer dep?

TrySound:
Not really necessary. Better to just stick with the same version as draft-js.

PR #1727, opened 2021-01-22
Fix regex to ignore trigger in text with supportWhitespace

### LINKED ISSUE #1723: [Mention Plugin] Whitespace is ignored before trigger
## Environment

- @draft-js-plugins/editor version: 4.0.0
- @draft-js-plugins/mention version: 4.0.2

## Description

With the option `supportWhitespace` active whitespaces before the trigger are ignored and the mention suggestion overlay is opening. This is not the case if you use `supportWhitespace: false`.

## Reproducible Demo

https://codesandbox.io/s/supportwhitespace-9yjzy
Enter Text with a following @ without a whitespace like `test@`

hcomPullReq:
## Checklist

## Use-case/Problem

Closes #1723


sibelius (approved):
cool tests

PR #1419, opened 2020-06-03
Add support for multiple mentionTriggers

kaustubh-karkare:
## Checklist

- [✓] Fix any eslint errors that occur
- [✓] Update change-log for every plugin you touch
- [✓] Add/Update tests if you add/change new functionality
- [?] Add/Update docs if you add/change functionality
- [✓] Enable "Allow edits from maintainers" for this PR

^ There aren't any docs for this plugin, but it is fairly straightforward (see the implementation section below).

## Use-case

Issue = #1418 

I'm working on a personal app that requires a rich text editor, for which I'm using DraftJS and plugins. In addition to be able to @mention people (which provides that nice typeahead experience), I also want the same for #hashtags and additional entities like $documents or something.

## Implementation

* MentionSuggestionsPortal (`draft-js-mention-plugin/src/mentionStrategy.js`) is a decorator that marks already-created or potential mention entities. I modified the regex it uses to handle a list of triggers, instead of a single one.
* The MentionSuggestions component (`draft-js-mention-plugin/src/MentionSuggestions/index.js`) goes through each "marked" item, checks to see if the current selection (aka text cursor position) is inside the appropriate range, and if yes, shows the dropdown. Since it already stores the active "marked" item, I modified it to also store the active trigger. This is provided to the `onSearchChange` (surfaces to the application) and the `addMention` method, which uses it to figure out the entity type to use in the editorState.
* I also modified the other decorator (`draft-js-mention-plugin/src/mentionSuggestionsStrategy.js`), which is used for styling mention entities, to recognize the addition triggers.
* Those were the major changes, rest of the updates were mostly signature changes to support a `Array<string>` instead of a plain `string`. I also updated the storybook example that I was using for testing.

## Demo

Demo

PR #1755, opened 2021-02-01
add multi support mentions

fancykingstar:
- Revise mention package to support multi mentions.


vas85:
@sibelius this is the same PR as the earlier one here: https://github.com/draft-js-plugins/draft-js-plugins/pull/1419

@fancykingstar and I work together. It would be great if we can get this merged

.eslintrc.js
```diff
@@ -86,7 +86,7 @@ module.exports = {
     'jsx-a11y/img-has-alt': 0,
     'react/require-default-props': 0,
     'react/display-name': 1,
-
+    'linebreak-style': 0,
```
fxOne, line 89: Why did you disable this rule?

fancykingstar:
without that I got the error like above in the windows.
Expected linebreaks to be 'LF' but found 'CRLF' linebreak-style

And that rule helps me to avoid it.

fancykingstar:
Hi @sibelius 
Would you review the PR and merge?

sibelius:
can you add this to storybook and docs as well?

fancykingstar:
I think I've already added.

fxOne:
> without that I got the error like above in the windows. Expected linebreaks to be 'LF' but found 'CRLF' linebreak-style And that rule helps me to avoid it.

The idea of the rule is that the code has only one type linebreaks and not mixed ones. Can you please enable the rule and fix potential issues.

packages/mention/src/index.tsx
```diff
@@ -58,8 +59,10 @@ export interface MentionPluginConfig {
   mentionSuggestionsComponent?: ComponentType;
   entityMutability?: 'SEGMENTED' | 'IMMUTABLE' | 'MUTABLE';
   mentionTrigger?: string;
+  mentionTriggers?: string[];
```
fxOne, line 62: If we have now `mentionTrigger` and `mentionTriggers` what happens if you set both?

fxOne (changes requested):

fxOne:
> I think I've already added.

You missed at least https://github.com/draft-js-plugins/draft-js-plugins/blob/master/packages/docs/pages/plugin/mention/index.tsx and the changelog.

packages/mention/src/index.tsx
fancykingstar: - If you'd like to use only one trigger, then you should use mentionTrigger like above
`const mentionPlugin = createMentionPlugin({ mentionTrigger: '(' });`

- if you'd like to use multiple triggers, then you should use mentionTriggers like
`const mentionPlugin = createMentionPlugin({ mentionTriggers: ['@', '#'] });`

I think this makes more sense.

fancykingstar:
@sibelius 
Would you take a look at PR and do me a favor to handle the error.

fancykingstar:
@sibelius Actually, I don't have this kind of error in the local.
I run 'npm run lint' to test lint in the local.

fancykingstar:
image
@sibelius @fxOne 
I just did strict clone and run "yarn lint" in the local machine.
But I got the error like above.

fxOne:
Run a `yarn build` to build the dependencies.

fancykingstar:
image
I got this error when I run "yarn build" @fxOne

fancykingstar:
image

fxOne:
You need to run it under linux or go into each folder and run yarn build except the docs folder

fancykingstar:
I run it under mac. doesn't it work?

fancykingstar:
I use yarn 1.22.4

fancykingstar:
image
I run "yarn build" in each subfolder and I got the error.

fxOne:
I just tried it under Windows with WSL2 and ubuntu 20.4 where it worked:
build
I have never used a mac so I cannot help you

fancykingstar:
Hi @sibelius @fxOne Would you review the PR and give me feedback?

fancykingstar:
@sibelius When could you review the PR.

sibelius:
I'm gonna have some time next week

sibelius:
I'm gonna have some time next week

fxOne:
I just tried the PR an notice that most of the mention storybook do not work (Mention Suggestions are not showing). Can you please check this?

fancykingstar:
image
Storybook works fine. Please checkout this image.

fxOne:
I tried it again and still not working:
Unbenannt

fancykingstar:
image
Sorry for confusion. I revised it and will push the updates.

fancykingstar:
@fxOne Would you review again?

fxOne:
The storybook examples `Editor with Mention Plugin and asynchronously loaded suggestions` and `Mentions with whitespace` are still not working for me.

fancykingstar:
Hi @fxOne.
I tried the storybook of master branch of draft-js-plugin and it was not working either. That means there isn't issue in my PR.
And I tested all the other storybook, docs, and examples, and they works as same as the original draft-js-plugin.

FYI, storybook for `Editor with Mention Plugin and asynchronously loaded suggestions` is not working because of base PATH.

fxOne:
I see the issue with `Editor with Mention Plugin and asynchronously loaded suggestions`. I'll try to fix it.

vas85:
@fxOne / @sibelius can we get a review of this please?

This feature has been asked for repeatedly for 6+ months by many people. It would be great to get this merged so that all users of this package can benefit.

sibelius:
We can't merge code that is not ready

vas85:
1. The issue you brought up already exists in this project and has nothing to do with this PR.
2. Other issues you brought up have been addressed

Please review and add comments specific to this PR. Existing issues are outside the scope of this PR

fxOne:
I still there an issue with the `Mentions with whitespace` example which works for on master but not in the PR

packages/mention/src/index.tsx
fxOne: can you change it to `mentionTrigger?: string | string[];` please?

PR #1807, opened 2021-02-19
Mention multi support

fancykingstar:


fancykingstar:
Hi @sibelius @fxOne This is new PR for multi mention support funtions.

I fixed the issue for `supportWhiteSpace mention editor` and pass `mentionTrigger` for both `string` and `string[]`
Please review it and give me a feedback.
Thanks.

packages/docs/components/Examples/mention/CustomComponentMentionEditor/CustomComponentMentionEditor.tsx
```diff
@@ -45,9 +45,12 @@ export default function CustomComponentMentionEditor(): ReactElement {
   const onOpenChange = useCallback((_open: boolean) => {
     setOpen(_open);
   }, []);
-  const onSearchChange = useCallback(({ value }: { value: string }) => {
```
fxOne, line 48: Maybe we should leave the original function without the `trigger` as in the non multi mention trigger use cases the option is irrelevant and for `defaultSuggestionsFilter` optional? This comment is for all samples which are not using multi mentions.

packages/docs/pages/plugin/mention/index.tsx
```diff
@@ -197,6 +204,14 @@ export default function Mention(): ReactElement {
             not support a multi-character mentionTrigger.
           </span>
         </div>
+        <div className={styles.param}>
```
fxOne, line 207: There are now 2 `mentionTrigger` paragraphs. I guess you can delete the old one. You also need to add the new parameter for the `defaultSuggestionsFilter` function.

stories/mentions-multi-triggers/src/mentions.js
```diff
@@ -0,0 +1,48 @@
+const mentions = [
```
fxOne, line 1: Can you please change this file to a TypeScript file?

fancykingstar: @fxOne all `mentions` files are js file, as you can see.
image

Do you want me to change file to a typescript file for only multiTriggers?
Or do it for all?

packages/docs/components/Examples/mention/CustomComponentMentionEditor/CustomComponentMentionEditor.tsx
fancykingstar: Leave original function without trigger for non multi mention trigger

packages/docs/pages/plugin/mention/index.tsx
fancykingstar: I removed old one.

stories/mentions-multi-triggers/src/mentions.js
fxOne: Would be nice if you have the time to migrate them all :)

packages/docs/pages/plugin/mention/index.tsx
fxOne: Theres also the paragraph `In addition to the plugin the module exports defaultSuggestionsFilter. As first argument it takes the search term as a String. The second argument is an array of mentions. The function returns the filter list based on substring matches.`. Would be nice if you could add the new `trigger` parameter.

fancykingstar:
I fix issues.

fxOne (approved):

PR #1894, opened 2021-03-13
[Emoji Plugin] Add unicode to emoji image component

### LINKED ISSUE #1877: Provide Unicode for each emoji to enable easier integration with other Emoji libraries
## Describe the feature

Provide the Unicode in the `emojiImage` configuration parameter to make it easier to integrate other Emoji libraries. 

## Motivation

Currently, the Emoji plugin supports EmojiOne and native art. EmojiOne's [free license](https://www.joypixels.com/licenses/free) makes it tricky to use in commercial products. However, [Twemoji](https://twemoji.twitter.com/) is under MIT license which makes it easier to use when building products.

In #1503, @fxOne added support for custom images to incorporate other Emoji libraries. 

Currently, the `config` prop passed into the `emojiImage` configuration has the following type declaration:

```typescript
export interface EmojiImageProps {
    emoji: string; // the decorated name like ":thumbs_up:"
    theme: EmojiPluginTheme; // all the class names
}
```

If the Unicode is available, then we could use it to lookup [assets in Twemoji by Unicode](https://github.com/twitter/twemoji/tree/master/assets/72x72) for example. So the proposed change would be to add something along the lines of:

```typescript
export interface EmojiImageProps {
    unicode: string // new prop
    emoji: string;
    theme: EmojiPluginTheme;
}
```

> It's possible I overlooked something and this is already possible in the current implementation.
> If that's the case, how should I do so?

## Possible implementations

I believe the Unicodes are in scope when the [Emoji list is created](https://github.com/draft-js-plugins/draft-js-plugins/blob/8f36e0c5c7ae9b9fe39e0b35af64dcb5ca09c298/packages/emoji/src/utils/emojiList.ts#L23). Maybe here is where the Unicode could be added?

## Related Issues

#1503

fxOne:
## Checklist

## Use-case/Problem

closes #1877

### DIFF
--- a/packages/docs/pages/plugin/emoji/index.js
+++ b/packages/docs/pages/plugin/emoji/index.js
@@ -395,7 +395,7 @@ export default class App extends Component {
               image for the suggestions. If this parameter is set the{' '}
               <InlineCode code={'useNativeArt'} /> is ignored. You can use this
               component to implement your own emoji library. It receives the
-              following props: emoji and theme
+              following props: emoji, unicode and theme
             </span>
           </div>
           <div className={styles.paramBig}>
--- a/packages/emoji/CHANGELOG.md
+++ b/packages/emoji/CHANGELOG.md
@@ -7,6 +7,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+## 4.3.0
+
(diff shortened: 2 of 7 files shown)


swkeever:
Thank you @fxOne! 🚀

PR #1978, opened 2021-04-11
[Image Plugin] adding missing forwardRef

### LINKED ISSUE #1977: Ref warning when using image and resizeable plugins
## Environment

"@draft-js-plugins/editor": "4.1.0",
"@draft-js-plugins/focus": "4.1.1",
"@draft-js-plugins/image": "4.1.0",
"@draft-js-plugins/resizeable": "4.1.0",
"draft-js": "0.11.7",
"react": "17.0.2",
"react-dom": "17.0.2",
"react-scripts": "4.0.0",

## Description
Hello, 
I am using the image and resizeable plugin to recreate the second example of the [image docs](https://www.draft-js-plugins.com/plugin/image), but I always get following warning message:

`Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

I've seen that there were some ref issues in the past, but maybe I am also using the library in a wrong way.
Another potential explanation could be, that I am currently using a function component with Hooks instead of a class and the `useRef `Hook might behave differently.

## Reproducible Demo

Here is an example that reproduces the error: https://codesandbox.io/s/bold-thunder-v2pnv?fontsize=14&hidenavigation=1&theme=dark

fxOne:
## Checklist

## Use-case/Problem

closes #1977
Adding missing forwardRef which is used by `@draft-js-plugins/resizable`

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,15 +15,15 @@
     "react-dom": "^16.9.0"
   },
   "devDependencies": {
-    "@babel/core": "7.13.14",
+    "@babel/core": "7.13.15",
     "@babel/plugin-proposal-class-properties": "7.13.0",
-    "@babel/preset-env": "7.13.12",
+    "@babel/preset-env": "7.13.15",
     "@babel/preset-react": "7.13.13",
     "@babel/preset-typescript": "7.13.0",
-    "@changesets/cli": "^2.15.0",
+    "@changesets/cli": "^2.16.0",
     "@rollup/plugin-babel": "^5.3.0",
     "@rollup/plugin-node-resolve": "^11.2.1",
-    "@storybook/react": "^6.2.4",
+    "@storybook/react": "^6.2.7",
     "@testing-library/jest-dom": "^5.11.10",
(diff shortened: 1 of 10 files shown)


Apporix:
@fxOne

The same warning occurs when combining `@draft-js-plugins/resizeable` and  `@draft-js-plugins/drag-n-drop`. So I guess that `forwardRef` also must be added in `@draft-js-plugins/drag-n-drop`.

It is possible that `forwardRef` is also missing in `@draft-js-plugins/drag-n-drop-upload` and/or `@draft-js-plugins/alignment`, although I have not tested it yet.

I guess that the fix is as simple as it was for  `@draft-js-plugins/image` .

fxOne:
@Apporix as the `@draft-js-plugins/drag-n-drop` works a little bit different I would like to ask you to create a new issue with a https://codesandbox.io example so that I can test a fix.

Apporix:
@fxOne 
I've created a new issue with a codesandbox example here; #1992

PR #2039, opened 2021-05-04
[Mention Plugin] Fixing multi-character trigger

### LINKED ISSUE #2017: Multi-character triggers no longer work
## Environment

- @draft-js-plugins/editor version: v4.1.0
- plugin name and version: @draft-js-plugins/mention v4.4.1

## Description

Multi-character triggers appear to be broken for `@draft-js-plugins/mention`.

With [this PR](https://github.com/draft-js-plugins/draft-js-plugins/pull/1807/files#diff-f5be53e8ae08de0a2489882b1824da7087229abf7a92b29c8d5f3a9522e43690R45), support for multi-character triggers appears to have gone away.  I've verified that a change along the following lines to `@draft-js-plugins/mention` v4.4.1 restores the support for multi-character triggers, although I'm not sure whether there are other unwanted side effects, e.g., related to the new group capture:

```patch
diff --git a/packages/mention/src/mentionSuggestionsStrategy.ts b/packages/mention/src/mentionSuggestionsStrategy.ts
index 3107214a..f7d726a8 100644
--- a/packages/mention/src/mentionSuggestionsStrategy.ts
+++ b/packages/mention/src/mentionSuggestionsStrategy.ts
@@ -67,9 +67,9 @@ export default (
   regExp: string
 ): ((contentBlock: ContentBlock, callback: FindWithRegexCb) => void) => {
   //eslint-disable-line
-  const triggerPattern = `[${triggers
+  const triggerPattern = `(${triggers
     .map((trigger) => escapeRegExp(trigger))
-    .join('')}]`;
+    .join('|')})`;
   const MENTION_REGEX = supportWhiteSpace
     ? new RegExp(`${triggerPattern}(${regExp}|\\s)*`, 'g')
     : new RegExp(`(\\s|^)${triggerPattern}${regExp}*`, 'g');
```

Note that the earlier code in the diff is constructing a regular expression along the lines of `/[trigger](...)/`; i.e., it's collapsing the multi-character trigger into a character class search.

In version v3.1.5 of `draft-js-mention-plugin`, a trigger of "in:" worked for me:

fxOne:
## Checklist

## Use-case/Problem

closes #2017

### DIFF
--- a/packages/mention/CHANGELOG.md
+++ b/packages/mention/CHANGELOG.md
@@ -5,6 +5,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## To Be Released
 
+## 4.5.0
+
+- fixing multi-character trigger [#2017](https://github.com/draft-js-plugins/draft-js-plugins/issues/2017)
 - change MentionSuggestions to popper.js with option `popperOptions`, `popoverContainer` and deprecate `popoverComponent` and `positionSuggestions` [#1933](https://github.com/draft-js-plugins/draft-js-plugins/issues/1933)
 - Fix @-mentions to work when the trigger character appears within the search string, for instance an email address.
 
--- a/packages/mention/package.json
+++ b/packages/mention/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@draft-js-plugins/mention",
-  "version": "4.4.1",
+  "version": "4.5.0",
   "sideEffects": false,
   "description": "Mention Plugin for DraftJS",
   "author": {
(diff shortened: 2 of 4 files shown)


emwalker:
Thank you for fixing this.  With the change, I'm able to update to the latest version of @draft-js-plugins.

Just a note that the capture seems to be off in the update.  Instead of returning the bare value after the prefix, I'm seeing a part of the prefix returned as well:
```javascript
"n:automobiles"
```
instead of
```javascript
"automobiles"
```

This was easy enough to provide a workaround for.  If I had to guess, downstream code somewhere is assuming a single-character trigger and so not removing the full prefix.

fxOne:
I would guess that the issue might be in this regex: https://github.com/draft-js-plugins/draft-js-plugins/blob/3e6f4b2b51530596ff605e0784139767b03e763e/packages/mention/src/utils/getSearchTextAt.ts#L20

Can you create a new issue for it, please?

Issue #1696, opened 2021-01-12
Mention keybindings not attached if initial suggestions is empty


felippepuhle:
## Environment

- @draft-js-plugins/editor version: 4.0.0
- @draft-js-plugins/mention version: 4.0.0

## Description

I found a bug where the workflow kinda makes sense for our project:

- Suggestion list starts as an empty array
- As soon as the user types `@` we open the suggestion list (we want to show a loading state, for example)
- Key bindings are not working (navigating using the keyboard), unless you move the cursor through the editor and open the suggestions again

Probably related to this line (I think we should remove it): https://github.com/draft-js-plugins/draft-js-plugins/blob/master/packages/mention/src/MentionSuggestions/MentionSuggestions.tsx#L216

## Reproducible Demo

https://codesandbox.io/s/draft-js-pluginsmention-dynamic-suggestion-keybindings-e6rhs

fxOne:
I don't think that removing [`this.props.suggestions.length > 0`](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/packages/mention/src/MentionSuggestions/MentionSuggestions.tsx#L216) is the right solution. Why should the dropdown be in an open state even if there are no suggestions to show? I think a better solution would be to reevaluate the dropdown open state when `this.props.suggestions` is changing.

felippepuhle:
Yeah, but in our case, we're opening the dropdown before having the suggestions (show a loading spinner, or an empty message if no results found - that lead us into another ticket where we'd need to support a custom renderer or something like that)... 

The case is, props are kinda inconsistent... why do we need to control an open state if we're evaluating the suggestions length (it doesn't make sense too)? 

I could solve it on our side for now by doing something like `suggestions={[{ id: 'loading' }]}` or `suggestions={[{ id: 'empty' }]}`, but it's not the ideal IMO.

fxOne:
I understand your issues. I think most of the application uses `open` and `onOpenChange` as a simple state without any logic. 

I would guess that changing suggestion to an `MentionData[] | Promise<MentionData[]>` would be to much work or didn't work?

felippepuhle:
Yeah I agree that `open` and `onOpenChange` is better and simple, that's the reason I suggested to remove the `suggestions.length > 0` verification.

I think `Promise<MentionData[]>` wouldn't work well too, we're using Apollo and it doesn't returns a promise for us (as well as several other data fetching libs).

PR #2051, opened 2021-05-08
[Mention Plugin] Do not render popover if there are no mentions

### LINKED ISSUE #2049: Bug in mention plugin - popover/popper.
## Environment

It happens everywhere, even on official website. 

- @draft-js-plugins/editor version: current
- plugin name and version: current

## Description

The issue appears to trigger when have more than one mention. For example of there is Matthew Russel and type @ and mouse is focused on at least one suggestion from suggestions list, deleting @ with backspace leaves empty popper artifact that can be discarded only after the page reload.

fxOne:
## Checklist

## Use-case/Problem

closes #2049

## Implementation

Do not render component if there are no mentions to show. @felippepuhle can you please have a look at it if it might break anything in relation with #1696


felippepuhle:
@fxOne we have an use case where we want to show a message if we have empty results... how we'd handle this case? Maybe if we'd have a new 'emptyComponent' prop or something like that?

fxOne:
@felippepuhle As you would like to show the `emptyComponent` inside the popup it wont change anything. I guess have to look for an other solution :)

fxOne:
@felippepuhle would it be OK if I add a `renderEmptyPopup` prop which is by default `false`? In your case you can set it to true and show the empty message component.

felippepuhle:
@fxOne I just checked and we're using `popoverComponent`, so I think we wouldn't be affected by this change

PR #2087, opened 2021-05-22
[Side Toolbar] Change to popper.js

### LINKED ISSUE #2054: max-height and overflow auto breaks the position of the side-tool bar plugin
Hi i noticed that if you set following css:
`.editor :global(.public-DraftEditor-content) {
  max-height: 450px;
  overflow: auto;
}
`
and press enough enter so it will start to overflow the position of the side-toolbar breaks and it looks like it should be if there was not the overflow: "auto" property.
here is a sandbox - https://codesandbox.io/s/xenodochial-banzai-vmg3b

fxOne:
## Checklist

## Use-case/Problem

closes #2054

## Implementation

`popper.js` can be controlled by the new `popperOptions` config parameter which is used for initialization.

Before:
image

After:
image

### DIFF
--- a/packages/docs/components/Examples/side-toolbar/CustomSideToolbarEditor/index.js
+++ b/packages/docs/components/Examples/side-toolbar/CustomSideToolbarEditor/index.js
@@ -35,6 +35,14 @@ export default class CustomSideToolbarEditor extends Component {
     });
   };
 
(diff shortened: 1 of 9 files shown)


ilanskogan:
@fxOne thanks man!

PR #1301, opened 2019-08-15
Modify RegEx to match trigger without leading space

jkhaui:
Simple fix for https://github.com/draft-js-plugins/draft-js-plugins/issues/973

Ideally, there should be a boolean prop for "requiresLeadingSpace"

PR #2165, opened 2021-06-28
[Mention Plugin] Modify RegEx to match trigger without leading space

### LINKED ISSUE #973: Trigger mention suggestions without leading space
## Use-case/Problem
The existing regex for initiating the suggestion dropdown requires that the mentionTrigger is the first character or is preceded by a space.  

See: `(\\s|^)${escapeRegExp(trigger)}${regExp}`
https://github.com/draft-js-plugins/draft-js-plugins/blob/master/draft-js-mention-plugin/src/mentionSuggestionsStrategy.js#L8

In my use case I wanted to the user to be able to wrap the mention in characters that provide other meaning for our application such as:
`Hi @Matt [@John]`
The mention for @Matt will work because it is preceded by a space, but the @John will not work because it is preceded by a `[`

I thought I could edit the `mentionRegExp` to overcome this, but found out that it only affects the portion after the trigger.

I feel like this variable should allow over riding the whole regex, but for backwards compatibility we could add new option such as `mentionTriggerRegExp` or something that would replace the entire thing and could default to the current implementation. (Although this name is less generic even though it has more responsibility)

Some questions:
**1. Is there anything in the plugin that requires the leading space?**
It seems odd to not allow this specific part of the regex to be over written. Perhaps this is by design and there was a requirement for the space in order to reliably get search text or selection of entity since it was known to be bounded by spaces?  Is there any reason this plugin couldn't support this scenario?

**2.  Is there anyway to configure the plugin to support this?**
Based on information I found, the answer is _no_.  I would either have to fork the plugin or write my own custom implementation to get around the hard coded regex in mentionSuggestionStrategy.  Although maybe I am wrong and some expert has already developed work-around.

**3. If the answer to the above 2 is true, what would you suggest the best approach is?**
I feel like I am starting to hit the limits of what the plugin is designed for and perhaps custom implementation would be better in the long run, but there is a whole bunch of cursor / content selection logic and work arounds in here that seem like the average person would not know unless they hit the issue and I thought re-using this work would be faster. 

For example I also would like to be able to customize behavior of adding space after entity is added. (Disabling it) and don't think this is possible.

TL;DR;
I would like to be able to have the mentionTrigger not required to be preceded by a space to allow my editor to support optional entities wrapped in brackets like: `[@John]`

ikhattab:
## Checklist

## Use-case/Problem

fixes #973

## Implementation

Modified trigger regex to allow triggering suggestions in the middle of text without leading space

## Demo

image

### DIFF
--- a/packages/mention/CHANGELOG.md
+++ b/packages/mention/CHANGELOG.md
@@ -5,6 +5,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## To Be Released
 
+- modify trigger regex to allow triggering menu suggestions without leading space [#973]
+
 ## 4.5.2
 
 - add `sideEffects` for css files [#1833](https://github.com/draft-js-plugins/draft-js-plugins/issues/1833)
--- a/packages/mention/src/__test__/mentionSuggestionsStrategy.test.ts
+++ b/packages/mention/src/__test__/mentionSuggestionsStrategy.test.ts
@@ -38,10 +38,10 @@ describe('test strategy with whitespace support disabled', () => {
     ['should match not match spaces', '@the walking dead', ['@'], [[0, 4]]],
     ['match within text', 'a lof @of text', ['@'], [[6, 9]]],
     [
-      'should not match if no whitespace before trigger',
+      'should match if no whitespace before trigger',
       'a lof@of text',
       ['@'],
-      [],
+      [[5, 8]],
     ],
     [
       'should match multiple mentions',
@@ -94,10 +94,10 @@ describe('test strategy with whitespace support enabled', () => {
     ['should match a word with special characters', '@ęĻŌ', ['@'], [[0, 4]]],
     ['match within text', 'a lof @of text', ['@'], [[6, 14]]],
     [
-      'should not match if no whitespace before trigger',
+      'should match if no whitespace before trigger',
       'a lof@of text',
       ['@'],
-      [],
+      [[5, 13]],
     ],
     [
       'should match multiple mentions with spaces',
@@ -156,7 +156,7 @@ describe('test strategy with whitespace support enabled', () => {
         '@the walking dead tv the white house',
         [
           {
-            key: (key as unknown) as number,
+            key: key as unknown as number,
             offset: 20,
             length: 15,
           },
@@ -192,7 +192,10 @@ describe('multi mentions test strategy with whitespace support disabled', () => 
       'should match a word with special characters',
       '@ęĻŌ ęĻŌ(',
       ['@', '('],
-      [[0, 4]],
+      [
+        [0, 4],
+        [8, 9],
+      ],
     ],
     [
       'should match not match spaces',
@@ -213,10 +216,13 @@ describe('multi mentions test strategy with whitespace support disabled', () => 
       ],
     ],
     [
-      'should not match if no whitespace before trigger',
+      'should match if no whitespace before trigger',
       'a lof@of text(',
       ['@', '('],
-      [],
+      [
+        [5, 8],
+        [13, 14],
+      ],
     ],
     [
       'should match multiple mentions',
@@ -259,7 +265,10 @@ describe('multi mentions test strategy with whitespace support enabled', () => {
       'should match a word with special characters',
       '@ęĻŌ ęĻŌ(',
       ['@', '('],
-      [[0, 8]],
+      [
+        [0, 8],
+        [8, 9],
+      ],
     ],
     [
       'should match not match spaces',
@@ -280,10 +289,13 @@ describe('multi mentions test strategy with whitespace support enabled', () => {
       ],
     ],
     [
-      'should not match if no whitespace before trigger',
+      'should match if no whitespace before trigger',
       'a lof@of text(',
       ['@', '('],
-      [],
+      [
+        [5, 13],
+        [13, 14],
+      ],
     ],
     [
       'should match multiple mentions',
@@ -322,7 +334,7 @@ describe('multi mentions test strategy with whitespace support enabled', () => {
         '@the walking dead tv the white house',
         [
           {
-            key: (key as unknown) as number,
+            key: key as unknown as number,
             offset: 20,
             length: 15,
           },
--- a/packages/mention/src/mentionSuggestionsStrategy.ts
+++ b/packages/mention/src/mentionSuggestionsStrategy.ts
@@ -5,15 +5,6 @@ interface FindWithRegexCb {
   (start: number, end: number): void;
 }
 
-const whitespaceRegEx = /\s/;
-
-function checkForWhiteSpaceBeforeTrigger(text: string, index: number): boolean {
-  if (index === 0) {
-    return true;
-  }
-  return whitespaceRegEx.test(text[index - 1]);
-}
-
 function findInContentBlock(
   regex: RegExp,
   text: string,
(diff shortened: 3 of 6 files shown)


ikhattab:
checking faild tests...

fxOne:
Did I get it right that you are looking at some issues with failing tests?
Can you please also add some new test cases for missing behavior?

ikhattab:
yes, you are right. 

definitely, this is a good idea, I currently don't have much time these days, will continue working on it when I have some time.

ikhattab:
I pushed what I believe could be a fix for `getSearchTextAt` tests. not totally sure if I got it right.

tbh I feel a bit lost here 😅

ikhattab:
I updated mentionSuggestionsStrategy tests now all tests are passed

p.s: I have to exclude `packages/docs` from the build because it was failing to build

fxOne:
@ikhattab Where had you issues with `packages/docs`? Id did yesterday a quick test on master and couldn't find any issues.

ikhattab:
I did this too in master and it works but on my branch it gave this
```
✖ bash -c tsc --noEmit:
packages/docs/components/Examples/inline-toolbar/SimpleInlineToolbarEditor/index.tsx(10,39): error TS2307: Cannot find module '@draft-js-plugins/inline-toolbar' or its corresponding type declarations.
packages/docs/components/Examples/mention/CustomComponentMentionEditor/CustomComponentMentionEditor.tsx(12,8): error TS2307: Cannot find module '@draft-js-plugins/mention' or its corresponding type declarations.
packages/docs/components/Examples/mention/CustomComponentMentionEditor/CustomComponentMentionEditor.tsx(26,24): error TS7006: Parameter 'mentionProps' implicitly has an 'any' type.
packages/docs/components/Examples/mention/CustomComponentMentionEditor/Mentions.ts(1,29): error TS2307: Cannot find module '@draft-js-plugins/mention' or its corresponding type declarations.
packages/docs/components/Examples/mention/CustomMentionEditor/CustomMentionEditor.tsx(15,8): error TS2307: Cannot find module '@draft-js-plugins/mention' or its corresponding type declarations.
packages/docs/components/Examples/mention/CustomMentionEditor/CustomMentionEditor.tsx(123,30): error TS7031: Binding element 'children' implicitly has an 'any' type.
packages/docs/components/Examples/mention/CustomMentionEditor/Mentions.ts(1,29): error TS2307: Cannot find module '@draft-js-plugins/mention' or its corresponding type declarations.
packages/docs/components/Examples/mention/MultiMentionTriggers/MultiMentionTriggers.tsx(6,8): error TS2307: Cannot find module '@draft-js-plugins/mention' or its corresponding type declarations.
packages/docs/components/Examples/mention/RemoteMentionEditor/RemoteMentionEditor.tsx(12,33): error TS2307: Cannot find module '@draft-js-plugins/mention' or its corresponding type declarations.
packages/docs/components/Examples/mention/SimpleMentionEditor/Mentions.ts(1,29): error TS2307: Cannot find module '@draft-js-plugins/mention' or its corresponding type declarations.
packages/docs/components/Examples/mention/SimpleMentionEditor/SimpleMentionEditor.tsx(12,8): error TS2307: Cannot find module '@draft-js-plugins/mention' or its corresponding type declarations.
packages/docs/components/UnicornEditor/Mentions.ts(1,29): error TS2307: Cannot find module '@draft-js-plugins/mention' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(9,33): error TS2307: Cannot find module '@draft-js-plugins/hashtag' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(10,33): error TS2307: Cannot find module '@draft-js-plugins/sticker' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(11,33): error TS2307: Cannot find module '@draft-js-plugins/linkify' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(14,8): error TS2307: Cannot find module '@draft-js-plugins/mention' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(15,31): error TS2307: Cannot find module '@draft-js-plugins/emoji' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(16,30): error TS2307: Cannot find module '@draft-js-plugins/undo' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(17,39): error TS2307: Cannot find module '@draft-js-plugins/inline-toolbar' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(18,37): error TS2307: Cannot find module '@draft-js-plugins/side-toolbar' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(19,31): error TS2307: Cannot find module '@draft-js-plugins/image' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(20,31): error TS2307: Cannot find module '@draft-js-plugins/focus' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(21,35): error TS2307: Cannot find module '@draft-js-plugins/alignment' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(22,34): error TS2307: Cannot find module '@draft-js-plugins/drag-n-drop' or its corresponding type declarations.
packages/docs/components/UnicornEditor/UnicornEditor.tsx(120,26): error TS7006: Parameter 'val' implicitly has an 'any' type.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
husky - pre-commit hook exited with code 1 (error)
```

I undo exclude `packages/docs` and run `yarn` in root directory and in `packages/docs` before running this

fxOne:
You need to build the packages with `yarn build` before so that the `d.ts` files are generated.

The issue I have with the solution is that the mentions will also trigger if I enter an email. 
As mentioned in #1301 it may be better If we add an option to set a custom regex or change the behavior with a flag.

ikhattab:
I tried running `yarn build` but it fails in building `docs` will try to debug it if this has something to do with my setup may be a different node version or something, is `yarn build` working for you?

image

I believe the next step for me to update [CONTRIBUTING.md](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/CONTRIBUTING.md#development) file to add this step and update the `docs` directory in instructions to be `packages/docs` will do this once I can build docs on my machine.

for this PR I believe it may be a good idea to close it and start a new one using one of the approaches you suggested, do you prefer one of them. I'd like to go with the flag one maybe `disableLeadinSpace`

fxOne:
Are you running it on Windows?

`@draft-js-plugins/docs` has as build script only a `.` as there is nothing to build but the command must be there. I dint't find yet a solution which also work on windows.

I think a RegEx solution would be more flexible. Can you please describe your use case as I'm unsure if maybe multiple triggers may be better.

ikhattab:
I'm using macOS Big Sur, I'm ok with RegEx solution too

my use case is: we are building an editing experience like the one used in Notion, we've built it on top of the mention plugin. we used `/` as trigger and decorators to add custom components it works very well actually but we need users to be able to trigger command in any part of the text without leading space, unfortunately, multiple triggers won't do this for us

image

Issue #2328, opened 2021-09-22
Configure linkify-it in [DraftJS Linkify Plugin]


sohaibqasem:
## Describe the feature
I want to exclude emails from being recognized as a link in [DraftJS Linkify Plugin]

Ability to configure linkify-it object in **createLinkifyPlugin** Funcation & **extractLinks** util
## Motivation
it would be very useful to determine what can be detected as a link in the editor, for me, I would to exclude emails

## Possible implementations
[extractLinks.ts](https://github.com/draft-js-plugins/draft-js-plugins/blob/master/packages/linkify/src/utils/extractLinks.ts)
```js
import linkifyIt from 'linkify-it';
import tlds from 'tlds';

export const linkify = linkifyIt();
linkify.tlds(tlds);

export function extractLinks(text: string, linkifyIt : any): linkifyIt.Match[] | null {
  if(linkifyIt) {
   return linkifyIt.match(text);
}
  return linkify.match(text);
}
```

And we can do something like this for the **createLinkifyPlugin** as a config

fxOne:
Sounds great, do you want to provide a PR?

sohaibqasem:
@fxOne yes sure, I will create a PR. Thanks

PR #2329, opened 2021-09-26
Linkify it feature #2328

sohaibqasem:
## Checklist

## Use-case/Problem

Ability to configure linkify-it object in **createLinkifyPlugin** Funcation & **extractLinks** util. So we can control what is being recognized as a link in [DraftJS Linkify Plugin], you can read more [here](https://github.com/draft-js-plugins/draft-js-plugins/issues/2328)

## Implementation

pass linkify-it object as optional argument to **extractLinks** util & **createLinkifyPlugin** Funcation and if its not passed we use the default (old) one.

## Demo
linkifyIt

### DIFF
--- a/packages/docs/components/Examples/linkify/CustomComponentLinkifyEditor/index.js
+++ b/packages/docs/components/Examples/linkify/CustomComponentLinkifyEditor/index.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { EditorState } from 'draft-js';
 import Editor from '@draft-js-plugins/editor';
-import createLinkifyPlugin from '@draft-js-plugins/linkify';
+import createLinkifyPlugin, { linkifyIt } from '@draft-js-plugins/linkify';
 import editorStyles from './editorStyles.module.css';
 
 const linkifyPlugin = createLinkifyPlugin({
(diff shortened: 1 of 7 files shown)


sohaibqasem:
Still there is a potential improvements I would like to discuss it with you.
we can simplify the linkify-it configuration to object of Booleans so developer don't need to read about linkify-it library.
something like: 
```js
options: {
   emails: true,
   links: false,
   ...
}
```
But in the other hand this implementation give us a full customizable object

fxOne:
I was thinking about a solution where you can set a function to the plugin configuration which expects a string as argument and returns a list of start/end points like the `extractLinks` function. In this case the user can use `linkify-it` or any other solution.

```typescript
interface ExtractLinks {
  (text: string): Array<{ index: number; lastIndex: number }> | null;
}
```

fxOne:
Your example could even be more simple as there is no issue if you return more properties:
```typescript
const linkifyPlugin = createLinkifyPlugin({
  extractLinks: (text) => {
    return linkifyit().match(text);
  } 
});
```

I don't like the idea of removing `linkify-it` from the dependencies as the plugin wont work out of the box anymore.

sohaibqasem:
Hi @fxOne , I pushed all discussed changes in the last review
But there is something about using `url?: string` OR `url: string`
If you take a look in `line 59 + 60` : Link.tsx
```js
  const links = customExtractLinks(decoratedText);
  const href = links && links[0] ? links[0].url : '';
```
href const expect to get his value from `links` and if url not exist, then it will be empty.
So, if we want to use  `url?: string` we should use `decoratedText` value if url doesn't exist like:
```js
  const links = customExtractLinks(decoratedText);
  const href = links && links[0] ? links[0].url || `//${decoratedText}` : '';
```
Note: "//" used here so the link won't be bound to the base url.

I used `url: string` as I think its a better for the plugin because we don't know what the type of the link like email, url, ...  
And "//" will handle all links the same way
The user have the responsibility to provide the right url if he chose to use a custom function
Any suggestions here :)

PR #2414, opened 2021-11-03
Use current inline style for mention

nhippenmeyer:
When inserting a mention, the current inline styles should be applied to the text (for instance font size, font family and text decoration) rather than only the default styles of the mention.


fxOne:
If you can add an entry in the changelog I would love to merge this

fxOne (approved):

PR #2433, opened 2021-11-20
Feat: Emoji picker menu position Alignment  ( top/buttom )

rohanvachheta:

## Checklist

## Use-case/Problem
The Menu for the Emoji picker was situated at the End of the parent, where for some cases we do require that position to be on top so we've provided one prop based on the position for menu 

## Implementation

Here I've provided external prop for the Emoji picker which will be indecating the position for the menu 
where defualt position is bottom which is same as below, if we provide props as "top" it will be align on top

## Demo

we need to provide this props in the Emoji select if we want to show Emojij menu items on tha top

` <EmojiSelect menuPosition="top" closeOnEmojiSelect />`


fxOne:
I like the idea for the PR. 

Can you change your implementation to `popper.js` as this would give us more options and should simplify the code? You can see in this PR https://github.com/draft-js-plugins/draft-js-plugins/pull/2212 an implementation I have already done for an other plugin.

rohanvachheta:
@fxOne  here I've added just Emoji Header button position not the whole Emoji picker position where it can be shown from top or bottom , ex: 
image

fxOne:
Ah, I understand now :)

If you could add you changes to the changelog and to the documentation I would love to merge this PR!

rohanvachheta:
@fxOne  yes added change log here https://github.com/draft-js-plugins/draft-js-plugins/pull/2433/commits/fa9a7bbc71282bbe802b6109cb9fe2950da2fe07

fxOne (approved):

PR #2522, opened 2022-01-08
Custom image documentation - Missing styles

ben467834:
As is the code doesn't have the required files present. We can't add to the code as it's used as an example on the docs page and there's a conflict with global styles. Therefore documenting on the page rather than in code just like the getting started guide.

## Checklist

## Use-case/Problem

## Implementation

## Demo


fxOne:
Shouldn't something like this work?

```jsx
<Code code={`
import '@draft-js-plugins/image/lib/plugin.css';
import '@draft-js-plugins/alignment/lib/plugin.css';
import '@draft-js-plugins/focus/lib/plugin.css';
${customExampleCode}`} name="AddImageEditor.js" />
```

PR #2534, opened 2022-01-13
[2518] - Make DragNDrop Plugin handle custom upload functionality. 

### LINKED ISSUE #2518: DragNDrop Upload Plugin's custom upload function not working.
Hi everyone. 
First of all, thanks for the plugins, they are awesome 😄 . 

## Environment

Plugins used: 

  - @draft-js-plugins/alignment: "^5.0.2",
  - @draft-js-plugins/drag-n-drop: "^4.2.1",
  - @draft-js-plugins/drag-n-drop-upload: "^4.1.1",
  - @draft-js-plugins/editor: "^4.1.0",
  - @draft-js-plugins/focus: "^4.1.2",
  - @draft-js-plugins/image: "^4.1.2",
  - @draft-js-plugins/resizeable: "^5.0.2",
  -  draft-js: "0.11.7"

## Description

I’m using the drag-n-drop image uploader. Everything seems to work fine with the code posted on the Draft-Plugins page, except that custom upload behavior seems to be failing. 

My goal is to actually upload the images to the server, then show them in the editor instead of the current behavior (show them in the editor without uploading them to the server). 

According to the documentation (on the [Alignment + Resize + Focus + Drag'n'Drop + Drag'n'Drop Upload Example]( https://www.draft-js-plugins.com/plugin/image)) this should be possible.

This is the code that I’m currently using (there is a Codesandbox live example below) :

https://github.com/fdemian/draft-plugins-test/blob/master/src/Editor.js

However, the mock upload function does not seem to get called. 
I dug around a bit and it seems that the relevant code in the drag-n-drop upload component is commented out.

https://github.com/draft-js-plugins/draft-js-plugins/blob/master/packages/drag-n-drop-upload/src/handleDroppedFiles.ts#L56-L112

It seems like without that code the upload function will not work. The handleDroppedFiles function creates a new form data object and populates it with files, but does not seem to be calling the custom upload function.

## Reproducible Demo

Sandbox: 
https://codesandbox.io/s/damp-haze-2srcj

### Repro steps: 

1. Open the sandbox.
2. Drag and drop an image into the editor. 

**Expected result**: 
The custom upload function is called, it should print a few statements in the console and launch an alert popup. 

**Actual result**: 
The image is uploaded as a b64 file and the  custom upload function is not called.

fdemian:
## Checklist

## Use-case/Problem

Fixes #2518 ( https://github.com/draft-js-plugins/draft-js-plugins/issues/2518 )

Initial implementation of the custom upload functionality.

This change adds custom upload functionality to the editor making it possible to upload the pictures to a server instead of the default behaviour (to upload images as base 64 strings).

## Implementation

This change implements the commented out code on the DraftJS Upload plugin. I've tried to implement the intended use of the function though it is not entirely clear to me if there ever was a functioning implementation. The retainSrc parameter was ignored because I have no idea what it was supposed to do.

### Custom upload implementation

The upload function works as follows: 

1. Start of the upload process: when the user drops a file on the editor, the customUpload function (if specified) will be called. 
2. The function then inserts placeholder blocks inside the editor for each file. The placeholder text has the following format  ` Uploading {name}...{' '}`. Where `name` is the name of the uploaded file (including the extension) and `progress` is the progress percentage (starts at 0%).
3. (Optional) The user is using a function that supports progress markers and marks a progress percentage. This will update the placeholders which will now indicate the progress. 
4A) Upload was successful, the editor removes the placeholders and inserts the images on the editor on the given placeholders.
4B) The upload failed, the function removes the placeholders.

## How to use

Here is an example, taken from the customUpload function that is being used inside the DragNDropUpload story.

```
export const customUpload = (data, success, failed, progress) => {

 const mockResult = data.files.map(f => ({ name: f.name, src: FILE_SRC }));

 let intervalId = -1;
 let currentProgress = 0;

 intervalId = setInterval(() => {
   if (currentProgress < 100) {
      currentProgress += 10;
      progress(currentProgress, mockResult[0]);
    }

    if (currentProgress === 100) {
      clearInterval(intervalId);
      success(mockResult/*, { retainSrc: true }*/);
    }

 }, 1000);

};
```

The parameters are as follows: 

- data: a list of files and a formData object (`{[File], formData}`). 
- success: a function to call on success. Needs to be called to a list of uploaded files `[{name: string, src: string}]`. Removes the upload placeholders and adds an image located at the given src.
- failed: a function to call when a file fails to be uploaded. Needs to be called with an object like so `{ name: string, src: string}`. Removes the upload placeholders.
- progress: a function to call in order to update file progress. Must be called with two parameters percent and file  (`{name:string, src:string}`). Percent is a number from 0 to 100 representing the upload progress. Updates the upload placeholder text with the given percentage.

## Demo

dnd-upload-demo


fxOne:
I just tried to upload a `.jpg` by dragging it over the editor in the DragNDropUpload story but the image was opened in a new tab. So it looks like `handleDroppedFiles` of the plugin wasn't called. Do you have any hint for me if I need to setup anything else?

fdemian:
> I just tried to upload a `.jpg` by dragging it over the editor in the DragNDropUpload story but the image was opened in a new tab. So it looks like `handleDroppedFiles` of the plugin wasn't called. Do you have any hint for me if I need to setup anything else?

Hi, @fxOne . You don't have to set up anything else. This is standard browser behaviour (it happens to me with almost any textarea, at least with chrome). 

When you drag the image over the text area you should stay there for a few seconds or else the browser does not focus on the textarea and opens the image on a different window. It happened to me too a few times during development :)  (this also happens with the previous version of the uploader). 

Another thing, the version of the uploadFunction that I created on the library will always upload a copy of the 'canada landscape' image because I wanted to test that it worked with an actual source and not a base64 source (you can see that in the code that i provided, `FILE_SRC` is just a string that links to the image.), but I can change this if this is not ideal (I don't have an actual server to upload images to).

fdemian:
@fxOne , let me know if you need any help setting it up. Do you observe a different behavior on that story prior to my change?

fxOne (approved):

fxOne:
Thanks for your support!

Issue #1104, opened 2018-05-07
Text Alignment Feature


juliankrispel:
This has been requested several times, and also has been a source of confusion. Basically at the moment the alignment plugin works solely with atomic blocks. This isn't clear from the docs or examples and it absolutely makes sense to implement this feature. Therefore to fix this we need to enable text alignment and make it clear in the docs that the alignment plugin also works with atomic blocks and that they're treated differently

- [ ] Make clear in the docs that block alignment and text alignment is separate.
- [ ] Add Text alignment buttons to draft-js-buttons for `left`, `right`, `center` and `justify` - I expect different icons may be required. These buttons would toggle block data (something like `textAlignment: string`)
- [ ] Add a `blockStyleFn` to the alignment plugin which will add css classes depending on the `textAlignment` block data prop. 
- [ ] Implement `handleReturn` inside the alignment plugin to carry over text-alignment of the current block to the next block.

juliankrispel:
Previous discussion in #941

MattyBalaam:
I’m in the process of building an editor for a new product and apart from this one feature, draft-js-plugins handles all the requirements. Has any progress been started on this yet? If not I may see if I can get up to speed with the plugin system and give this a go myself over the next couple of weeks.

juliankrispel:
@MattyBalaam no progress as of yet, there's a doc here detailing how to write plugins https://github.com/draft-js-plugins/draft-js-plugins/blob/master/HOW_TO_CREATE_A_PLUGIN.md

MattyBalaam:
OK, great. I’ll hope to make a start this week. Would you mind if I reach out on the Slack channel if I hit any snags?

juliankrispel:
absolutely - please do

felipediogo:
Is there any progress on this? Does anybody need some help to get this finished?

MattyBalaam:
I was using Draft in my last job and had a rough POC I worked on in my own
time until the requirement was shelved. From memory it was lacking icons. I
can try and find it and share with you if you want to polish it.

syedammarzia:
Any progress on this one?

juliankrispel:
Nope.

sanishkr:
I read all this and then I saw [this](https://github.com/jpuri/react-draft-wysiwyg) where all types of text alignments are created using draft-js. Why is there no direct support from draft JS or plugin available officially? Tell me if I am missing something.

hengjiUSTC:
[...]

PR #2554, opened 2022-01-23
[1104] Text Alignment Feature

HemaSAli:
## Checklist

## Use-case/Problem

Add `Text Alignment Feature` based on #1104.

## Implementation

 - [x] Add Text alignment buttons to draft-js-buttons for left, right, center and justify - I expect different icons may be required. These buttons would toggle block data (something like textAlignment: string)
 - [x]  Add a `blockStyleFn` to the alignment plugin which will add css classes depending on the textAlignment block data prop.
 - [x]  Implement handleReturn inside the alignment plugin to carry over text-alignment of the current block to the next block.

#### Based on [Draftjs Documentation](https://draftjs.org/docs/advanced-topics-block-styling/#blockstylefn), the user need to add this style in his own CSS
```
.editor :global .draftBlockAlign-left .public-DraftStyleDefault-block {
  text-align: left;
}

.editor :global .draftBlockAlign-center .public-DraftStyleDefault-block {
  text-align: center;
}

.editor :global .draftBlockAlign-right .public-DraftStyleDefault-block {
  text-align: right;
}

```
If you have a better idea, we can Do it :)

## Demo
Text Alignment Feature - Static Toolbar
Text Alignment Feature - Inline Toolbar


fxOne:
Your changes looks great.

But you should create a new plugin instead of modifying the editor. The idea of this library is that users can compose the feature they want.

HemaSAli:
The reason why I did that, is that the alignment buttons are related to the toolbar, I mean he can customize the toolbar as he wants, What do you think about removing them from the default static toolbar?
so if he want to use them, he can import them .

HemaSAli:
#close 
Working on Text Alignment Plugin

PR #2576, opened 2022-02-01
Text alignment plugin

HemaSAli:
## Checklist

## Use-case/Problem
Add `Text Alignment Plugin`, which can be used with static/inline toolbar, based on #1104.

## Implementation
 - [x] Add Text alignment buttons to draft-js-buttons for left, right, center and justify - I expect different icons may be required. These buttons would toggle block data (something like textAlignment: string)
 - [x]  Add a `blockStyleFn` to the alignment plugin which will add css classes depending on the textAlignment block data prop.
 - [x]  Implement handleReturn inside the alignment plugin to carry over text-alignment of the current block to the next block.

## Demo
Text Alignment Feature - Static Toolbar

#### Note: In the doc I used it in ```static toolbar``` but we can use it in inline toolbar also.


fxOne (approved):

fxOne:
Thanks for the new plugin!

pwnedev:
"Add Text alignment buttons to draft-js-buttons for left, right, center and **justify**"

There isn't a justify button in this plugin, is it because it doesn't work well while inline styling?

Flip437:
> "Add Text alignment buttons to draft-js-buttons for left, right, center and **justify**"
> 
> There isn't a justify button in this plugin, is it because it doesn't work well while inline styling?

Same question here.

PR #2607, opened 2022-02-14
fix: type error on addImage function

sayjeyhi:

## Checklist

## Use-case/Problem

## Implementation

## Demo


sayjeyhi:
@fxOne please check this as well, there is an issue with types here

fxOne:
The types that you added does not look correct. Can you check them?
Can you also explain what error you want to fix?

sayjeyhi:
`addImage` function has a third parameter to pass additional data when we are adding images, that's why we need to have this feature here, I also added other changes if you want I can share them as well

fxOne:
I think it would be better to just add the `index` to the `addImage` callback as the `handleUpload` and `addImage` are both exposed function. With the `index` you should be able to get your `extraData` and I would guess the information could be usefull for other usecases.

What do you think about that?

PR #2687, opened 2022-04-03
feat: export Popover component for custom container usage without losing Popper functionally

igoldny:
## Checklist

## Use-case/Problem
When you want to overwrite the MentionSuggestions container with a custom style you lose Popper.js functionally which in some cases you want to overwrite the container only for styling without losing Popper.js functionally.
for that manner, you could wrap your container with a Popper component which will handle that for you.
Popper component currently privet, exporting that component will enable this feature.

## Implementation

Export Popover component for custom container usage without losing Popper functionally

```
<MentionSuggestions
        popoverContainer={({children, ...props}) => (
          <Popover {...props}>
            <div style={{ background: '#000', color: '#fff' }}>
              {children}
            </div>
          </Popover>
        )}
      />
```

## Demo

image

### DIFF
--- a/packages/mention/CHANGELOG.md
+++ b/packages/mention/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## To Be Released
 
+## 5.2.2
+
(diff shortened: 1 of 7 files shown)


fxOne (approved):

fxOne:
Thanks for the PR!

PR #2765, opened 2022-05-16
Inline toolbar plugin should account for ancestor scrollTop

trevor-davis-dover:
## Checklist

## Use-case/Problem

If an ancestor of DraftEditor-root has a max height and overflow: auto set, we don't position the inline toolbar correctly because we don't account for scrollTop on ancestor nodes. As a result, the toolbar can appear far down the screen and even off page in these scenarios.

## Implementation

Aggregate scrollTop on all ancestor nodes and include it in positioning calculation for the inline toolbar.

## Demo

### DIFF
--- a/packages/inline-toolbar/CHANGELOG.md
+++ b/packages/inline-toolbar/CHANGELOG.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 #4.1.4
 
-- Account for scroll offsets in parent elements of DraftEditor-root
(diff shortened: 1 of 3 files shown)


trevor-davis-dover:
@fxOne you seem to be a frequent contributor, can you look at this or point me to someone who can? Thank you!

fxOne:
Looks great, but I think it would be better to outsource all the calculation to `popper.js`. I have already implemented this in the side toolbar https://github.com/draft-js-plugins/draft-js-plugins/pull/2087. 

Can you do this for the inline toolbar as well? If not, I could merge this PR and address it later.

fxOne (approved):

PR #2786, opened 2022-05-23
Inline toolbar - account for situation

trevor-davis-dover:

## Checklist

## Use-case/Problem

A previous PR of mine added support for max-height parents by accounting for scrollTops on ancestor nodes of the draft editor root. However, we short-circuited erroneously when that scrollTop was 0. Instead, we should keep recursing up the DOM tree.

## Implementation

## Demo

### DIFF
--- a/packages/inline-toolbar/CHANGELOG.md
+++ b/packages/inline-toolbar/CHANGELOG.md
@@ -5,7 +5,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## To be released
 
-#4.1.4
+## 4.1.5
+
+- Account for scroll offsets of 0 in parent elements of DraftEditor-root
+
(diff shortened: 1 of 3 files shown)


trevor-davis-dover:
@fxOne sorry for the spam, it would be awesome if you could review / merge this when you get a chance. The issue was that I was checking for scrollTop being defined (due to a cast), but that logic also short-circuited on scrollTop of 0. This fixes that with a fallback to 0 when the scrollTop prop is null or undefined.

fxOne:
There are some conflicts. Would be nice if you could fix them.

trevor-davis-dover:
> There are some conflicts. Would be nice if you could fix them.

Weird, I can't understand why git would think these are conflicts. Working on it, thanks for the callout.

PR #2792, opened 2022-05-25
Trevor davis/inline toolbar ignore html scrolltop

trevor-davis-dover:

## Checklist

## Use-case/Problem

One more fix: we want to ignore the scrollTop on the root HTML node.

## Implementation

## Demo


trevor-davis-dover:
@fxOne @naveensrinivasan @hatsu38 @nikgraf please don't be annoyed, one more fix here :) Each node's offset is relative to the root HTML, so if we account for the root HTML node's scrollTop as well, we are "double counting" that value in the calculation. Sorry for so many PRs, I keep finding new edge cases. Hopefully this is the last of it.

trevor-davis-dover:
@fxOne just following up on this. Can you merge it when you get a chance? Thanks!

fxOne (approved):

PR #2884, opened 2022-07-27
Autoclose emoji popover for multiple instance

mistryshaan:

## 

## Use-case/Problem

When using multiple emoji picker plugins on one page (imagine a list of posts like fb timeline that each have its own comment input and emoji picker button) there is an issue that already open emoji picker popover want close if you click on another instances popover trigger button.
The issue can be seen on demo page of emoji plugin https://www.draft-js-plugins.com/plugin/emoji

80152072-676a6700-8591-11ea-9c60-c2348207c013

## Solution

As can be seen, in the change I have updated the event listener on the document to call a specific function `closeIfClickedOutside` which is simply checking if the click target element is something inside emoji popover component or no. And if no then it calls a function to close the popover and doing nothing otherwise.

``` 
closeIfClickedOutside = (e: React.MouseEvent<HTMLElement>) => {
    if (this.emojiSelectRef && !this.emojiSelectRef.contains(e.target)) {
      this.closePopover();
    }
}
```

## Demo

After fix the one popover instance will always close when opening another popover.
80152781-a9e07380-8592-11ea-8c62-7ca858e343b8

### DIFF
--- a/packages/emoji/CHANGELOG.md
+++ b/packages/emoji/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## To be released
 
+## 4.6.2
+
+- Autoclose emoji popover for multiple instance [#2875](https://github.com/draft-js-plugins/draft-js-plugins/issues/2875)
+
 ## 4.6.1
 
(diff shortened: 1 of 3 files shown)


fxOne:
@mistryshaan can you look at the typescript issues?

fxOne (approved):

PR #2885, opened 2022-07-30
Table plugin

dobesv:
## Use-case/Problem

This is a plugin to allow tables in draft-js.  This offers the ability to render tables and (later) some functions to insert and resize them.

Currently this PR is in an early stage of development and I wanted to get some early feedback from project maintainers about it before I go too far down the path of building this out as a plugin added to this repo.

Specifically:

- Is this something that  you would accept into the core project when it's "ready", or should I put it into some other project?
- Any early feedback on the changes so far?

## Implementation

The tables are structured as a series of content blocks, the first is the configuration for the table and the ones that follow are the cells.  When the table renders, it render all the cell blocks that follow it in the block list itself.  Draft's default rendering is blocked for the cells so they are only rendered once.

This work is based on an this example I found on stack overflow, although my approach is slightly different: https://github.com/EvanMorrison/draft-js-rte/blob/master/components/organisms/richEditor/index.js

## Demo

image


fxOne:
Is it correct that there are no controls to create a table, only to display one?

dobesv:
> Is it correct that there are no controls to create a table, only to display one?

Hi, it's early days on this one, I will add methods to create a table later if the way tables are represented is OK.

fxOne:
I think that it looks good for now. But without the edit mode it is hard to tell.

PR #2916, opened 2022-08-19
Speed up initial opening of select

BSeblu:
## Checklist

## Use-case/Problem

Opening the Emoji Select or the Emoji Suggestion for the first time is extremely slow, ca. 3 seconds.
Subsequent usage is fast though.

The reason for that is, that `shortnameToUnicode` (from the `emoji-toolkit` library) uses a really huge RegEx, which is slow for the first call.

## Implementation
Instead of using this library method, we create a mapping from shortname to unicode from the same data used to build the EmojiStrategy.

### DIFF
--- a/packages/emoji/src/utils/__test__/shortnameToUnicode.test.ts
+++ b/packages/emoji/src/utils/__test__/shortnameToUnicode.test.ts
@@ -1,0 +1,10 @@
+import shortnameToUnicode from '../shortnameToUnicode';
+
+describe('shortnameToUnicode', () => {
+  it('executes in less than 100ms', () => {
+    const start = performance.now();
+    shortnameToUnicode(':grinning:');
+    const executionMs = performance.now() - start;
+    expect(executionMs).toBeLessThan(100);
+  });
+});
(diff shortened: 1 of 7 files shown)


BSeblu:
> Wouldn't it make more sense to fix this in the `emoji-toolkit` library?

I don't think so. This plugin needs to map an emoji-name to a unicode character, which you can technically do with `emoji-toolkit`'s `shortnameToUnicode`. But that method is doing a different job. In an arbitrary string it replaces every instance of emoji-names with the corresponding  unicode character.

So is `emoji-toolkit`'s implementation of `shortnameToUnicode` inefficient? Maybe, maybe not. But I don't think it's the right tool for the job. The name `shortnameToUnicode` is in fact a bit misleading: If I haven't looked into the sources, I would have suggested that it just maps an emoji-name to a unicode character.

packages/emoji/src/utils/__test__/shortnameToUnicode.test.ts
```diff
@@ -0,0 +1,10 @@
+import shortnameToUnicode from '../shortnameToUnicode';
+
+describe('shortnameToUnicode', () => {
+  it('executes in less than 100ms', () => {
```
fxOne, line 4: I don't like the idea of testing the performance of a function in a unit test. On a very slow computer this test may fail, which is wrong.

It would make more sense to test only if the function works correctly.

BSeblu: added some tests...

fxOne (approved):

PR #3056, opened 2023-01-02
improve `shortnameToUnicode` performance not using `emoji-toolkit` function anymore

marcoburrometo:
Somehow, `shortnameToUnicode` was not finding all the shortcodes and was using `shortnameToUnicode` from emoji-toolkit, which is using a very slow regex.
Now it's very much faster 💯 

Some `toShort(item.unicode)` are UTF-16 encoded strings, while `str` is UTF-8. So I use `localeCompare` to compare this correctly.

Also, `mapShortnameToUnicode` was not including skins.

## Checklist

### DIFF
--- a/packages/emoji/src/utils/shortnameToUnicode.ts
+++ b/packages/emoji/src/utils/shortnameToUnicode.ts
@@ -3,24 +3,25 @@ import data from 'emojibase-data/en/compact.json';
 
 const mapShortnameToUnicode: { key: string; value: string }[] = [];
 
-for (const item of data) {
+data.forEach((item) => {
   mapShortnameToUnicode.push({
     key: toShort(item.unicode),
     value: item.unicode,
   });
   if (item.skins) {
-    for (const skin of item.skins) {
+    item.skins.forEach((skin) => {
       mapShortnameToUnicode.push({
         key: toShort(skin.unicode),
         value: skin.unicode,
       });
-    }
+    });
   }
-}
+});
 
 export default function shortnameToUnicode(str: string): string {
   return str.length
-    ? // Somehow encoding is different, checking with indexOf instead of === works
-      mapShortnameToUnicode.find((x) => x.key.indexOf(str) > -1)?.value || ''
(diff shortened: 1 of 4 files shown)


packages/emoji/src/utils/shortnameToUnicode.ts
```diff
@@ -1,13 +1,24 @@
...
+        key: toShort(skin.unicode),
+        value: skin.unicode,
+      });
+    });
+  }
+});
 export default function shortnameToUnicode(str: string): string {
-  return mapShortnameToUnicode[str] || emojiToolkitShortnameToUnicode(str);
+  return str.length
+    ? // Somehow encoding is different, checking with indexOf instead of === works
```
fxOne, line 21: Can you tell me in which case `===` does not work?
It would also be a better solution to use a map instead of iterating over the array.

marcoburrometo: for example with `:thumbsup` it's not working.

I will change it to map, thx

fxOne: You wrote that there might be an issue with encoding, that why I asked. As far as I know this function is always called with the full shortname.
By map I meant the native `Map`.

marcoburrometo: for example I got this:

`encodeURI(':thumbsup:️') --> ':thumbsup:%EF%B8%8F'`

marcoburrometo: image


that's the issue in it's happening

marcoburrometo:
@fxOne I dig deep in this. Basically some `toShort(item.unicode)` are UTF-16 encoded strings, while `str` is UTF-8. So I use `localeCompare` to compare this correctly.

fxOne:
can you add tests for it?

marcoburrometo:
> can you add tests for it?

Done.

fxOne (approved):

Issue #3233, opened 2023-07-05
Emoji plugin keyboard navigability issues


LBPeterson:
## Environment
- OS: MacOS 12.6
- Browser: Chrome 107
- @draft-js-plugins/emoji version: `^4.6.5`
- draft-js version: `^0.11.7`

## Description
When using the emoji plugin, keyboard navigation is not supported. This manifests itself in several ways:
- The EmojiSelect menu cannot be opened via keyboard when the EmojiSelect button has focus
- When the EmojiSelect menu is open, there is no visual indication of which emoji has focus when navigating via the tab key
- When the EmojiSelect menu is open and an emoji has focus, that emoji cannot be selected with the keyboard `Enter` or `space (' ')` keys

## Reproducible Demo
https://www.draft-js-plugins.com/plugin/emoji

LBPeterson:
I have created a PR to address this problem at https://github.com/draft-js-plugins/draft-js-plugins/pull/3234

fxOne:
Closed by #3234

PR #3234, opened 2023-07-05
Improve emoji keyboard navigability

LBPeterson:
## Checklist

## Use-case/Problem
Issue [#3233](https://github.com/draft-js-plugins/draft-js-plugins/issues/3233)
The emoji plugin lacks keyboard navigation ability

## Implementation

This creates the appropriate styles and keyboard handlers to allow the EmojiSelect menu to be navigable via keyboard.

## Demo
After:

https://github.com/draft-js-plugins/draft-js-plugins/assets/11545784/3eab1edd-a776-4c9b-a11a-4c6d645a8ed2

### DIFF
--- a/packages/emoji/CHANGELOG.md
+++ b/packages/emoji/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## To be released
 
+## 4.7.0
+- added basic keyboard navigation to fix [#3233](https://github.com/draft-js-plugins/draft-js-plugins/issues/3233)
+- Removed support for the `emojiSelectPopoverEntryFocused` theme prop in favor of using `:focus` on emojiSelectPopoverEntry
+
 ## 4.6.5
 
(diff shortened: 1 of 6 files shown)


fxOne:
I really like the solution, but do you see any way to not remove `emojiSelectPopoverEntryFocused` as this would be a breaking change?

LBPeterson:
> I really like the solution, but do you see any way to not remove `emojiSelectPopoverEntryFocused` as this would be a breaking change?

A simple solution to this may be to apply the `isFocused` tracking callbacks to the `onFocus` and `onBlur` event handles, in addition to the `onMouseEnter` and `onMouseLeave`.

This would incur a performance cost though, as keeping track of focus state in an elements internal state will cause a rerender with each `onMouseEnter`, `onMouseLeave`, `onFocus`, & `onBlur` event. 

Native css focus and hover states are much faster, but my linaria css knowledge is not enough to know if there is a way to do a similar thing completely within the theme file.

LBPeterson:
I have updated the PR to remove the breaking change and follow the existing strategy here, with a note for future refactor opportunity.

fxOne (approved):

Issue #3254, opened 2023-08-02
Emoji plugin suggestions popover overflows viewport


sblaz:
## Environment

- OS: MacOS 13.4.1
- Browser: Chrome 115, Firefox 116
- @draft-js-plugins/emoji version: `^4.6.5`
- draft-js version: `^0.11.7`

## Description

When the cursor in the editor is near the bottom of the viewport and one brings up the emoji suggestions popover, it overflows the viewport so that it isn't visible.

<img width="1016" alt="emoji-suggestions-overflow ong" src="https://github.com/draft-js-plugins/draft-js-plugins/assets/515609/bde0e28d-7246-4498-8a58-ec170bc24c4f">

## Reproducible Demo

[https://www.draft-js-plugins.com/plugin/emoji](https://www.draft-js-plugins.com/plugin/emoji)

sblaz:
I created a PR with an attempt at using popper.js for this like the mention plugin does at #3255

PR #3255, opened 2023-08-02
Improve emoji popover positioning

sblaz:

## Checklist

## Use-case/Problem

Issue #3254
The emoji plugins suggestions popover can overflow the viewport when it pops up near the bottom.

## Implementation

This follows the precedent set in the `mention` plugin, and follows that example by using [popper.js](https://popper.js.org/docs/v2/) to implement the popover. It also deprecates the `positionSuggestions` prop and introduces `popperOptions`, again trying to align with the example set in the mention plugin (PR #2029).

## Demo

After:
emoji_suggestions_popover

### DIFF
--- a/packages/docs/pages/plugin/emoji/index.js
+++ b/packages/docs/pages/plugin/emoji/index.js
@@ -331,13 +331,24 @@ export default class App extends Component {
           <div className={styles.paramBig}>
             <span className={styles.paramName}>positionSuggestions</span>
             <span>
-              The function can be used to manipulate the position of the popover
-              containing the suggestions. It receives one object as arguments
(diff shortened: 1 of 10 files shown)


fxOne (approved):

sblaz:
I think the EmojiSelect popover suffers from a similar issue. Would it make sense to try to bundle a fix for that up in the same release? At first glance it looks like to use the popover.js react hook would imply converting from a class component to functional component... 🤔  Assuming maybe we try to tackle that with a separate issue/PR?

fxOne:
Yep, the EmojiSelect popover has the same problem. 
I would say a separate issue would be best!

PR #3256, opened 2023-08-03
Speed up initial load for native Emojis

dugancathal:
## Checklist

## Use-case/Problem

When testing the Emoji plugin locally (and on the draft-js-plugins demo site),
we see that it takes upwards of 2 seconds to build mappings from shortname to
unicode and back again. These mappings are built *on initial import*. For us,
as we import the library early on, this slows down time to First Contentful
Paint to a level that's unnacceptable for our application. It's also notable
on the Emoji demo site.

In performance exploration, we found that the emoji plugin leverages the
`toShort` function (from `emoji-toolkit`) heavily to convert from unicode to
the shortname. The plugin previously also mapped from shortname to unicode
separately. This double-mapping appears to be the primary cause of the
slowdown (for the `useNativeEmoji` option, at least).

Note: timings in the screenshots from a 2022 MacOS M1 running latest Google Chrome,
but this appears to be replicable on Firefox as well.

## Implementation

To speed up that page load, we change the `createEmojisFromStrategy` function
to maintain the mapping of shortname to unicode for us in a new `EmojiShape`.
Downstream, this means that we don't have to repeatly do that translation. By
doing this, we can eliminate the need for the `shortnameToUnicode` function
altogether.

Further, we leverage the `shortnameLookup` object from the `emoji-toolkit`
library, and fall back to `toShort`. This fallback appears necessary as there
are some emoji in the `emojibase-data` set that aren't immediately translatable
(`:neutral_face:` was one).  Unfortunately, the `shortnameLookup` object is
initialized in `emoji-toolkit` as an array and Typescript can't deduce that it
is actually a key-value map - this necessitates a weird type-cast in the
`fasterUnicodeToShortname` shim.

In local testing through the demo site, this appears to drop the page-load time
by almost 1.5s on average.

## Demo


fxOne:
It looks very good so far. If you could resolve the conflicts, I would test it this weekend and merge it.

dugancathal:
Conflicts resolved. Thank you for the quick response! 💯

fxOne:
I noticed that there are now several similar emojis with the same short name:
image
Sometime there are 2, sometimes there are 4 it's a bit wired.

dugancathal:
Hello there! Thanks for noticing that and apologies for not catching it on my end. It looks like one _possible_ fix is to rename the secondary `heart` emoji unicode in the default `priorityList` (see newest commit). But this appears to break the `joypixels` emoji list 🤔 -- we could filter out the emoji names that don't exist for `joypixels`, but that feels weird.

I'm not exactly an expert at the emoji unicode, so I don't know if there's a better name for the two heart emojis here (that might work with joypixels), or if we can remove one of them ... what do you think?

image
image

fxOne:
I would expect that when using `joypixels` all supported emojis would be displayed correctly. So I think filtering them out would be the best solution.

dugancathal:
@fxOne That makes perfect sense. Since the two `heart` emoji are (basically) the same, and they were being collapsed for both JoyPixel and Native previously, I opted to remove the secondary unicode. That appears to fix the problem for both emoji libraries. Does that make sense?

fxOne (approved):

dugancathal:
Thank you @fxOne! We've pulled this into our code and see a *dramatic* improvement in page load time. Really appreciate your time and energy on helping us get this across the line 😀 🥳

